Issue #6, opened 2018-06-05
Any plans on supporting YAML?


fenollp:
BTW the readme says
```golang
openapi3.NewSwaggerLoader().LoadFromFile
```
while it should read 
```golang
openapi3.NewSwaggerLoader().LoadSwaggerFromFile
```

PR #8, opened 2018-06-12
lint

fenollp:
Fixes most warnings found by `golangci-lint run --enable-all`.
Note: also fixed the README typo mentioned in #6 

e2f282b2d5b100186585a3f72cc2b24073ff6c66 breaks some tests. Would you mind looking at it?

Thanks


fenollp:
This had changes that broke some tests. Any idea on how to fix it?

PR #13, opened 2018-06-30
connect: add the one missing HTTP method: CONNECT

fenollp:


fenollp:
Ah Travis needs to be re-enabled, after the rename ;)

PR #16, opened 2018-07-02
externalref: fragment prefix was just missing

rikkkky:


fenollp:
Hi! Could you add a quick test for this?
I’m on my phone right now but can review later today. 
Also I just merged a big PR that fixes many issues, maybe you won’t be happy about it so please fetch and report :)

My next step is to look for many correct and incorrect openapi3 specs and integrated them to the tests.

rikkkky:
I'll do it later. Thank you for the notice!

> My next step is to look for many correct and incorrect openapi3 specs and integrated them to the tests.

Strongly agreed.

fenollp:
Very nice!

PR #18, opened 2018-07-02
format: support format

rikkkky:
This PR supports formats strictly.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#data-types

### DIFF
--- a/openapi3/schema_test.go
+++ b/openapi3/schema_test.go
@@ -137,6 +137,34 @@ var schemaExamples = []schemaExample{
 	},
 
 	{
+		Title: "INTEGER",
+		Schema: openapi3.NewInt64Schema().
+			WithMin(2).
+			WithMax(5),
+		Serialization: map[string]interface{}{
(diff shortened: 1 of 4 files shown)


openapi3/schema_test.go
```diff
@@ -510,3 +538,79 @@ var schemaExamples = []schemaExample{
...
+		for _, typ := range example.AllInvalid {
+			err := validateType(t, schema, typ)
+			require.Error(t, err)
+		}
+	}
+}
+
+func validateType(t *testing.T, schema *openapi3.Schema, typ string) error {
+	schema.WithFormat(typ)
+	return schema.Validate(nil)
```
fenollp: Could you use `context.TODO` instead of `nil` here?

fenollp: Nah I'll do it

fenollp:
Thanks again!

PR #22, opened 2018-07-09
add missing `nil` check in `resolveSchemaRef()`

jokeyrhyme:
Howdie! Great project!

We're getting the following panic:

> panic: runtime error: invalid memory address or nil pointer dereference [recovered]
> 	panic: runtime error: invalid memory address or nil pointer dereference
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x158 pc=0x11e54af]
> 
> goroutine 21 [running]:
> testing.tRunner.func1(0xc4202801e0)
> 	/Users/.../go/src/testing/testing.go:742 +0x29d
> panic(0x17b9c00, 0x1d93d80)
> 	/Users/.../go/src/runtime/panic.go:502 +0x229
> .../vendor/github.com/getkin/kin-openapi/openapi3.(*SwaggerLoader).resolveSchemaRef(0xc420274840, 0xc420229e60, 0xc420316640, 0xc420390598, 0x0)
> 	.../vendor/github.com/getkin/kin-openapi/openapi3/swagger_loader.go:380 +0x16f

After reviewing swagger_loader.go, I noticed that all of the other `resolve...Ref()` methods have a `nil`-check that is missing from `resolveSchemaRef()`

This PR puts the missing check in, and I can confirm that this resolves our panic

Cheers!

### DIFF
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -375,6 +375,9 @@ func (swaggerLoader *SwaggerLoader) resolveSchemaRef(swagger *Swagger, component
 		component.Value = resolved.Value
 	}
 	value := component.Value
+	if value == nil {
+		return nil
+	}
 
 	// ResolveRefs referred schemas
 	if v := value.Items; v != nil {
--- a/openapi3/swagger_loader_test.go
+++ b/openapi3/swagger_loader_test.go
@@ -85,6 +85,15 @@ func TestResolveSchemaRef(t *testing.T) {
 	require.NotNil(t, refAVisited.Value)
 }
 
+func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
+	source := []byte(`{"info":{"description":"An API"},"paths":{"/foo":{"post":{"requestBody":{"content":{"application/json":{"schema":null}}}}}}}`)
+	loader := openapi3.NewSwaggerLoader()
+	doc, err := loader.LoadSwaggerFromData(source)
+	require.NoError(t, err)
+	err = doc.Validate(loader.Context)
+	require.EqualError(t, err, "Found unresolved ref: ''")
+}
+
 type sourceExample struct {
 	Location *url.URL
 	Spec     []byte


fenollp:
Hi there, thanks a lot for contributing!

Would you mind adding a minimal snippet that exhibits this bug to `openapi3/swagger_loader_test.go`?

Thanks & have a nice day

jokeyrhyme:
Hi @fenollp , I've added a test, but I've also made the possibly-controversial change of having the tests in the same package as the non-test code
The function under test was package-private, and I've been having issues isolating and producing a reduced test case that only involves exported methods

fenollp:
I see so
* there’s no need to directly call resolveSchemaRef. Calling LoadSwaggerFromData is enough. 
* The call to check that there is no error is require.NoError, not require.NotNil. Maybe they’re the same but you know, semantics. 
* If a $ref points to something that does not exist, surely an error should be returned, no?
* Are other ref-resolving functions free of this possible error?

jokeyrhyme:
> there’s no need to directly call resolveSchemaRef. Calling LoadSwaggerFromData is enough.
As I've perhaps poorly articulated earlier, I'm having trouble sticking to LoadSwaggerFromData, because it has been difficult for me to determine what part of our data structure is causing the issue. I'll spend more time today trying to isolate it further. At this stage I am unable to share our data structure, which means our conversation is going to unfortunately be limited.

> The call to check that there is no error is require.NoError, not require.NotNil. Maybe they’re the same but you know, semantics.
Good catch, I'll fix this.

> If a $ref points to something that does not exist, surely an error should be returned, no?
Yep, and this is definitely the case for all the minimal data structures that I've tested so far. However, it does seem that `resolveSchemaRef()` is called with a `nil` Component at some point.

> Are other ref-resolving functions free of this possible error?
Our data structure is large, but all of the references are schema references, so I cannot answer this question. However, all of the other `resolve...Ref()` functions are protected by the `nil`-check that is missing from `resolveSchemaRef()`

jokeyrhyme:
Okay, @fenollp , I figured it out :)
We had a `"schema": null` in our data structure, similar to the reduced test case I have in this PR now, which was causing the `nil`-panic
The reduced test case confirms that this loads and properly returns the expected error without panicking

PR #23, opened 2018-07-23
Resolve example references

danielgtaylor:
This adds support for resolving [example references in media types](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#mediaTypeObject) when using the `examples` list. Here is a sample:

```yaml
# ...
components:
  examples:
    foo:
      value:
        msg: Hello, world!
paths:
  /path:
    get:
      responses:
        200:
          application/json:
            examples:
              test:
                $ref: "#/components/examples/foo"
# ...
```

Without this change you get back the raw ref when trying to output the example:

```js
{
  "$ref": "#/components/examples/foo"
}
```

### DIFF
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -336,11 +336,11 @@ func (swaggerLoader *SwaggerLoader) resolveResponseRef(swagger *Swagger, compone
 		if contentType == nil {
 			continue
 		}
-		for i, example := range contentType.Examples {
+		for name, example := range contentType.Examples {
 			if err := swaggerLoader.resolveExampleRef(swagger, example); err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


fenollp:
Hi! Thanks for contributing! Just a couple of things:
* could you add a test maybe in `swagger_loader_test.go` that displays the behavior
* can you ensure that no other `Examples` fields need to be resolved also
* please rename that `i` identifier to something more explicit such as `exampleName`

danielgtaylor:
@fenollp thanks for the feedback! I have added a test, renamed the variable, and have double checked that a single example cannot be a reference via the spec. Should be good to merge!

fenollp:
Amazing! thanks!

PR #27, opened 2018-08-01
Fix of format: date-time

kwry:
I corrected the dateTime of format correctly because it is date-time.

ref. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types


fenollp:
Thanks!

Issue #26, opened 2018-08-01
securitySchema.flow is correctly flows


kwry:
The field name of the securitySchema `flow` is incorrect. Correctly is `flows`.

```diff
diff --git a/openapi3/security_scheme.go b/openapi3/security_scheme.go
index b427855..75615cf 100644
--- a/openapi3/security_scheme.go
+++ b/openapi3/security_scheme.go
@@ -17,7 +17,7 @@ type SecurityScheme struct {
        In           string      `json:"in,omitempty"`
        Scheme       string      `json:"scheme,omitempty"`
        BearerFormat string      `json:"bearerFormat,omitempty"`
-       Flow         *OAuthFlows `json:"flow,omitempty"`
+       Flow         *OAuthFlows `json:"flows,omitempty"`
 }
 
 func NewSecurityScheme() *SecurityScheme {
```

ref. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#securitySchemeObject

Together with the property name of the struct is also I think it is better to fix, this struct is not worried what to do because of the public to the outside.
Should I just modify the JSON tag without changing the property name?

PR #28, opened 2018-08-03
Fix securitySchema.flow is correctly flows and validation

kwry:
from. https://github.com/getkin/kin-openapi/issues/26

Fix above issue.
I also added a test using OpenAPI3 Sample. c646ac292a8058d4555e3d9096327c51236ab4c8

Also, I found that adding a test erroneously verified each Flow.
Required parameters differ depending on the grant type.
see. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#fixed-fields-25

Therefore, I changed the IF of Validate of OAuthFlow, but can not you see whether this policy is good?

### DIFF
--- a/openapi2conv/openapi2_conv.go
+++ b/openapi2conv/openapi2_conv.go
@@ -236,7 +236,7 @@ func ToV3SecurityScheme(securityScheme *openapi2.SecurityScheme) (*openapi3.Secu
 	case "oauth2":
 		result.Type = "oauth2"
 		flows := &openapi3.OAuthFlows{}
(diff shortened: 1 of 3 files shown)


kwry:
@fenollp 

I am sorry that the correspondence became late.

> I'd rather not export OAuthFlowTypeImplicit and others for now. Plus I'd prefer them to be const ints.

I changed it with f551e3e. Is it okay with such feeling?

fenollp:
Amazing! Thanks

PR #29, opened 2018-08-12
Added MediaType definition to be interpreted as JSON

kwry:
I also want to validate MediaType other than application/json as JSON.

- `application/json-patch+json`
    - [RFC 6902 - JavaScript Object Notation (JSON) Patch](https://tools.ietf.org/html/rfc6902)
- `application/problem+json`
    - [RFC 7807 - Problem Details for HTTP APIs](https://tools.ietf.org/html/rfc7807)

Therefore, MediaType to be treated as JSON can be designated externally.
I am glad if you get input if there is any sense of incongruity in this IF.

### DIFF
--- a/openapi3filter/media_type.go
+++ b/openapi3filter/media_type.go
@@ -2,7 +2,7 @@ package openapi3filter
 
 // Definition of MediaType to be interpreted as JSON
 var DefaultJSONMediaTypes = []string{
-	"aplication/json",
+	"application/json",
 }
 
 func isMediaTypeJSON(mediaType string) bool {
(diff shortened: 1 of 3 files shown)


openapi3filter/media_type.go
```diff
@@ -0,0 +1,15 @@
+package openapi3filter
+
+// Definition of MediaType to be interpreted as JSON
+var DefaultJSONMediaTypes = []string{
+	"aplication/json",
```
fenollp: Typo

kwry: opss...

kwry:
thx!

PR #31, opened 2018-08-22
Return original error if allOf

kwry:
When allOf becomes an error, it returns the original error.
For this implementation, I did not set the fast flag for allOf.
In order to be able to judge with other use cases, I have stored errors in Origin for use.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -694,7 +694,7 @@ func (schema *Schema) visitSetOperations(value interface{}, fast bool) (err erro
 		if v == nil {
 			return foundUnresolvedRef(item.Ref)
 		}
-		if err := v.visitJSON(value, true); err != nil {
+		if err := v.visitJSON(value, false); err != nil {
 			if fast {
 				return errSchema
 			}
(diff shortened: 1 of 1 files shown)


fenollp:
I see. How about this instead:
```patch
		if err := v.visitJSON(value, false); err != nil {
+			// Return full error message for allOf
-			if fast {
-				return errSchema
-			}
			return &SchemaError{
				Value:       value,
				Schema:      schema,
				SchemaField: "allOf",
-				Origin:      err,
			}
		}
```
This also removes the need for introducing `Origin`.

kwry:
Thank for comment.
Suggesting patch version output below error message.

```
Request body has an error: doesn't input the schema: Doesn't match schema "allOf"
```

I want to see information on the cause field.

```
panic: Request body has an error: doesn't input the schema: Error at "/category":Field must be set to string or not be present
```

fenollp:
Hm in that case could you add a test that compares the error message returned? A test that looks for the parts you are interested in in the error message of an invalid payload.

kwry:
Added nested schema tests by fe080e7 .
Please check the test you are expecting.

PR #32, opened 2018-08-26
Adding operation ID and summary fields to API V2 converter.

diogogmt:


fenollp:
Hi there, Thanks for taking the time of upstreaming!
This looks fine. Could you just add some testing code in `openapi2conv/openapi2_conv_test.go` that ensures the fields you want are actually set please?

Thanks & have a nice day.

PR #35, opened 2018-09-11
fix: openapi3 operation AddResponse nil pointer

magederek:
Fix a bug which will cause nil pointer panic. In the previous code in openapi3/operation:AddRespinse(), when operation.Responses == nil, any assignments to responses variable will panic as nil pointer since NewResponses() only assign to operation.Responses and remains responses to be nil.

### DIFF
--- a/openapi3/operation.go
+++ b/openapi3/operation.go
@@ -65,7 +65,8 @@ func (operation *Operation) AddParameter(p *Parameter) {
 func (operation *Operation) AddResponse(status int, response *Response) {
 	responses := operation.Responses
 	if responses == nil {
-		operation.Responses = NewResponses()
+		responses = NewResponses()
+		operation.Responses = responses
 	}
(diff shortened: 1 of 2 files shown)


magederek:
@fenollp please check the CI. When it tried to go get, it failed.

fenollp:
LGTM. I restarted the CI job as it failed just due to a network error.
Could you add a simple regression test?

magederek:
I added some tests for openapi3/operation. This test is passed after the bug fix. It would fail before fixing the nil pointer bug.

fenollp:
All good! Now if you could run `go fmt ./...` (`openapi3/operation_test.go` has a formatting issue) I'll be able to merge this.

magederek:
Forgot to run gofmt and gofmt and goimports. The format issue has been fixed now.

PR #36, opened 2018-09-26
Add support for path parameter refs

danielgtaylor:
This adds a missing feature where path parameter references were ignored and the `param.Value` was  never filled in. Added a test which failed before but passes now.


fenollp:
Hi there, thanks for upstreaming!
Could you edit your spec so that it is actually valid per https://editor.swagger.io/ please?

danielgtaylor:
@fenollp example updated :smile:

PR #50, opened 2018-11-12
Nested relative references handling

pruser:


fenollp:
This is great! Thanks

PR #52, opened 2018-11-13
Security scheme: adding in:cookie as an option

pruser:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#securitySchemeObject

```
in: The location of the API key. Valid values are "query", "header" or "cookie".
```


fenollp:
Do you mind editing the error message right below too?

fenollp:
Thanks!

Issue #25, opened 2018-07-27
JSON Schema discriminator fails to parse


danielgtaylor:
As noted in [apisprout#1](https://github.com/danielgtaylor/apisprout/issues/1#issuecomment-408490874), this [sample file](https://raw.githubusercontent.com/hskrasek/conference-api-docs/master/specs/openapi.yaml) fails to parse because the discriminator is an object instead of a string. It looks like [the spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md#discriminatorObject) says it should be an object like in the linked example file.

fenollp:
PRs welcomed! Sorry I don't have much time these days...

goller:
@danielgtaylor with my PR in this should now work!

PR #54, opened 2018-12-12
feat(schema): add discriminator struct to match OpenAPI

goller:
Here is a PR to handle the discriminator object rather than a string.

It would #25 

Love the library :)  I'm using it to do all sorts of swagger tests

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -96,7 +96,7 @@ type Schema struct {
 	MinProps             uint64                `json:"minProperties,omitempty"`
 	MaxProps             *uint64               `json:"maxProperties,omitempty"`
 	AdditionalProperties *SchemaRef            `json:"-" multijson:"additionalProperties,omitempty"`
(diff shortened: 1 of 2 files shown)


openapi3/schema.go
```diff
@@ -96,7 +96,7 @@ type Schema struct {
 	MinProps             uint64                `json:"minProperties,omitempty"`
 	MaxProps             *uint64               `json:"maxProperties,omitempty"`
 	AdditionalProperties *SchemaRef            `json:"-" multijson:"additionalProperties,omitempty"`
-	Discriminator        string                `json:"discriminator,omitempty"`
+	Discriminator        Discriminator         `json:"discriminator,omitempty"`
```
fenollp: Looks like this needs to be turned into a pointer so that tests can pass.

goller:
Hey @fenollp, updated!

fenollp:
Thanks!

PR #57, opened 2019-01-17
Add resolve Content in Query

zghnr1993:


openapi3/swagger_loader_relative_refs_test.go
```diff
@@ -113,6 +113,17 @@ var refTestDataEntries = []refTestDataEntry{
 			require.Equal(t, "id", swagger.Paths["/test/{id}"].Get.Parameters[0].Value.Name)
 		},
 	},
+
+	{
+		name:            "PathOperationParameterRefWithContentInQuery",
+		contentTemplate: externalPathOperationParameterWithContentInQueryTemplate,
+		testFunc: func(t *testing.T, swagger *openapi3.Swagger) {
+			fmt.Print(swagger.Paths["/test/{id}"].Get.Parameters[0].Value.Content["application/json"].Schema.Ref)
+			require.NotNil(t, swagger.Paths["/test/{id}"].Get.Parameters[0].Value.Content["application/json"].Schema.Value)
```
fenollp: Could you put this value in a variable & reuse it below?
Also remove the debug line above.

openapi3/swagger_loader_relative_refs_test.go
```diff
@@ -420,6 +431,40 @@ const externalPathOperationParameterSchemaRefTemplate = `
...
+                    }
+                ]
+            }
+        }
+    },
+    "components": {
+        "parameters": {
+            "CustomTestParameter": {
+                "content": {
+                    "application/json": {
```
fenollp: Per editor.swagger.io (which supports v3) parameters do not support `content > mediaType` so this spec is invalid. This is valid however:

```
openapi: 3.0.0
info:
  title: ''
  version: '1'
paths:
  '/test/{id}':
    get:
      responses: {default: {description: ''}}
      parameters:
      - $ref: '#/components/parameters/CustomTestParameter'
components:
  parameters:
    CustomTestParameter:
      in: path
      name: id
      required: true
      schema:
        $ref: '%s#/components/schemas/CustomTestSchema'
```

fenollp: Alright so editor.swagger.io is incomplete: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameter-object
So this is indeed valid! Sorry about that.

Could you add a validation pass in parameter.go that ensures 
>  the content property can define the media type and schema of the parameter. A parameter MUST contain either a schema property, or a content property, but not both

Along with a test that contains both. Thanks!

zghnr1993: Hi！ sorry , i can't understand you completely ! why a paramenter must contain either a schema property or a content property .but not both ? i think someone have such requirement that one or more single schema attribute and one or more content object in a parameter !

and i can't understand what validation pass needs me to add in parameter.go ??
i think you had implemented it

type Parameter struct {
	ExtensionProps
	Name            string                 `json:"name,omitempty"`
	In              string                 `json:"in,omitempty"`
	Description     string                 `json:"description,omitempty"`
	Style           string                 `json:"style,omitempty"`
	AllowEmptyValue bool                   `json:"allowEmptyValue,omitempty"`
	AllowReserved   bool                   `json:"allowReserved,omitempty"`
	Deprecated      bool                   `json:"deprecated,omitempty"`
	Required        bool                   `json:"required,omitempty"`
	Schema          *SchemaRef             `json:"schema,omitempty"`
	Example         interface{}            `json:"example,omitempty"`
	Examples        map[string]*ExampleRef `json:"examples,omitempty"`
	Content         Content                `json:"content,omitempty"`
}

func (parameter *Parameter) Validate(c context.Context) error {
	if parameter.Name == "" {
		return errors.New("Parameter name can't be blank")
	}
	in := parameter.In
	switch in {
	case
		ParameterInPath,
		ParameterInQuery,
		ParameterInHeader,
		ParameterInCookie:
	default:
		return fmt.Errorf("Parameter can't have 'in' value '%s'", parameter.In)
	}
	if schema := parameter.Schema; schema != nil {
		if err := schema.Validate(c); err != nil {
			return fmt.Errorf("Parameter '%v' schema is invalid: %v", parameter.Name, err)
		}
	}
	if content := parameter.Content; content != nil {
		if err := content.Validate(c); err != nil {
			return fmt.Errorf("Parameter content is invalid: %v", err)
		}
	}
	return nil
}

i haven't totally understand your code ! i just encountered the problem in my work !

openapi3/swagger_loader_relative_refs_test.go
zghnr1993: ok ,I will delete debug code . but i don't understand what value i can put in a variable and reuse it? i think i totally follow  style of your code ! could you give me a detailed remind ?

openapi3/swagger_loader_relative_refs_test.go
fenollp: > why can’t both be set?
That’s just what the OAS spec says ;)

> Validate un parameter.go
Really it’s just to add a check that ensures one cannot set both. So after the schema and content validation you copy pasted above, we should add a condition on parameter.Schema && parameter.Content

openapi3/swagger_loader_relative_refs_test.go
fenollp: Oh it’s just for DRY. Note that i didn’t write most of this code so if the style is to repeat many things then that is not something I agree with...

zghnr1993: well ! please review my new commit .

fenollp:
Amazing. Thanks for your work and patience!

PR #58, opened 2019-01-18
Fix typo in swagger v2 schemes

meier-christoph:
There was a typo in the json field name which prevented the parser from loading schemes properly.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields


fenollp:
Thanks!

PR #59, opened 2019-02-04
Support loading specifications from YAML files

wwerner:
This PR enables loading OAI specs from YAML files by making `LoadSwaggerFromDataWithPath` aware of the file extension.

If it is `yml` or `.yaml`, it tries unmarshalling YAML, otherwise it behaves like before.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
+# Macos file system
+.DS_Store
+
+# IntelliJ / GoLand
+.idea
--- a/README.md
+++ b/README.md
@@ -61,26 +61,60 @@ func GetOperation(httpRequest *http.Request) (*openapi3.Operation, error) {
 
 ## Validating HTTP requests/responses
 ```go
-import (
-  "net/http"
+package main
 
-  "github.com/getkin/kin-openapi/openapi3"
-  "github.com/getkin/kin-openapi/openapi3filter"
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"log"
+	"net/http"
+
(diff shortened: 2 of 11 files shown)


fenollp:
How about this instead?
```patch
diff --git a/openapi3/swagger_loader.go b/openapi3/swagger_loader.go
index 375d30b..b1ede49 100644
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -2,7 +2,6 @@ package openapi3

 import (
        "context"
-       "encoding/json"
        "errors"
        "fmt"
        "io/ioutil"
@@ -90,7 +89,7 @@ func (swaggerLoader *SwaggerLoader) LoadSwaggerFromFile(path string) (*Swagger,

 func (swaggerLoader *SwaggerLoader) LoadSwaggerFromData(data []byte) (*Swagger, error) {
        swagger := &Swagger{}
-       if err := json.Unmarshal(data, swagger); err != nil {
+       if err := yaml.Unmarshal(data, swagger); err != nil {
                return nil, err
        }
        return swagger, swaggerLoader.ResolveRefsIn(swagger, nil)
@@ -98,7 +97,7 @@ func (swaggerLoader *SwaggerLoader) LoadSwaggerFromData(data []byte) (*Swagger,

 func (swaggerLoader *SwaggerLoader) LoadSwaggerFromDataWithPath(data []byte, path *url.URL) (*Swagger, error) {
        swagger := &Swagger{}
-       if err := json.Unmarshal(data, swagger); err != nil {
+       if err := yaml.Unmarshal(data, swagger); err != nil {
                return nil, err
        }
        return swagger, swaggerLoader.ResolveRefsIn(swagger, path)
```

Does this pass your tests too?

wwerner:
It does indeed pass my tests and is much cleaner, thanks.
Since I messed up this PR, I'll open a new, clean one.

Issue #60, opened 2019-02-06
discriminator mapping should contain strings, not schema ref objects


rogpeppe:
The following program fails with the error:

```
Error while unmarshalling property 'components' (*openapi3.Components): Error while unmarshalling property 'schemas' (*map[string]*openapi3.SchemaRef): Error while unmarshalling property 'discriminator' (*openapi3.Discriminator): Error while unmarshalling property 'mapping' (*map[string]*openapi3.SchemaRef): Failed to unmarshal extension properties: json: cannot unmarshal string into Go value of type map[string]json.RawMessage
Input: "#/components/schemas/Cat"
```

The data is a schema simplified from the [the OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) - the mappings property should be `map[string]string`, or at least allow string values when unmarshaling.

```
package main

import (
	"fmt"

	"github.com/getkin/kin-openapi/openapi3"
)

var data = `
{
	"openapi": "3.0.0",
	"components": {
		"schemas": {
			"MyResponseType": {
				"discriminator": {
					"mapping": {
						"cat": "#/components/schemas/Cat",
						"dog": "#/components/schemas/Dog"
					},
					"propertyName": "pet_type"
				},
				"oneOf": [
					{
						"$ref": "#/components/schemas/Cat"
					},
					{
						"$ref": "#/components/schemas/Dog"
					}
				]
			}
		}
	}
}
`

func main() {
	loader := openapi3.NewSwaggerLoader()
	_, err := loader.LoadSwaggerFromData([]byte(data))
	if err != nil {
		fmt.Println("error: ", err)
	} else {
		fmt.Println("ok")
	}
}
```

fenollp:
Fixed by #62

PR #62, opened 2019-02-06
openapi3: use map[string]string for mapping

rogpeppe:
Technically this is a backwardly incompatible change, but
the Discriminator type has not been officially released yet,
so it should be OK.

Fixes issue #60.


fenollp:
Indeed I merged the previous implementation too hastily.

It seems this mapping should be of property names to schema references (the Ref per of SchemaRef). 
So this needs to be a string and needs some more handling. I’d love to see some validation and tests but let’s merge this right now to unblock some people.

PR #67, opened 2019-02-10
Support loading specifications from YAML files

wwerner:
Enables YAML format support for specifications. Since JSON is a subset of YAML, we can just use the YAML parser to read both (thanks @fenollp).

I also removed `LoadSwaggerFromYAMLData` from `swagger_loader` as it does the same as `LoadSwaggerFromData` now and adjusted the tests accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,10 @@
 
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
+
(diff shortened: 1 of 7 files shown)


fenollp:
cc https://github.com/getkin/kin-openapi/pull/59
LGTM. Im on my phone rn. Will merge once I’m home. Thanks :)

PR #73, opened 2019-03-15
Resolve Ref Inside Path Item

azharprabudi:
At [documentation OpenAPI Specification 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#pathItemObject) there is $ref property to get object from reference. But there is no explaination, how to use reference object either using relative path or network file or components. 

Inside this pr, the $ref object can be read but just working on relative path or network file. For component, i dont know what the key it will be use for that

### DIFF
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -2,6 +2,7 @@ package openapi3
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


openapi3/swagger_loader.go
```diff
@@ -145,17 +146,26 @@ func (swaggerLoader *SwaggerLoader) ResolveRefsIn(swagger *Swagger, path *url.UR
...
-	for _, pathItem := range swagger.Paths {
+	for endpoint, pathItem := range swagger.Paths {
 		if pathItem == nil {
 			continue
 		}
-		for _, parameter := range pathItem.Parameters {
+		// Refer on documentation at Open Api Specification 3.0 at
+		// https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#pathItemObject
+		// there is one property is $ref, to get the configuration at another reference object. For this time this issue
+		// will be fix, with the rules the value just come from relative path file or network file
```
fenollp: I think it would be better to move this out of this loop into its own loop, right before. This way paths get resolved first, then their contents get resolved.
Also can you remove your comments?

openapi3/swagger_loader.go
```diff
@@ -567,3 +577,85 @@ func (swaggerLoader *SwaggerLoader) resolveExampleRef(swagger *Swagger, componen
 	}
 	return nil
 }
+
+func (swaggerLoader *SwaggerLoader) resolvePathItemRef(swagger *Swagger, endpoint string, ref string) error {
+	if !swaggerLoader.IsExternalRefsAllowed {
```
fenollp: Are you not able to use `resolveComponent`? See for instance how `resolveRequestBodyRef` uses it.

azharprabudi: I am already check the resolveComponent function, i think it just resolve the network fille.
But doesnt resolve the relative path file, is it right ?

fenollp: Looking at the code, `resolveComponent` calls `LoadSwaggerFromURI` which [looks like it falls back to resolving `file://` URLs](https://github.com/getkin/kin-openapi/blob/d3180292eead101a815a9291fd8b6aca8b45927e/openapi3/swagger_loader.go#L67). If that is not possible then it should be made more explicit, but I believe constructing a `url.URL` with a the proper `.Path` set should arrive there. Let's try using `resolveComponent` :)

iron-s:
There is mention in spec about where $ref can point, here is an excerpt for the spec:

> Allows for an external definition of this path item

Note "external" here. So I assume it can't be local reference (and there is no component to hold it). This leaves us with just network path.

kapilratnani:
Is this PR still WIP or abandoned ?

fenollp:
Closing in favor of https://github.com/getkin/kin-openapi/pull/119

PR #74, opened 2019-03-19
Add support for RFC 3339 time-secfrac to the date-time regex

bkrebsbach:
Per the RFC 3339 spec, a date-time may contain an optional time-secfrac for fractional seconds: https://tools.ietf.org/html/rfc3339#section-5.6.

Add support for that case and update the tests.


fenollp:
Indeed! Thanks

Issue #77, opened 2019-03-24
"Required" validation does not work correctly with a request's body


nikolaas:
Hello!

There is an operation in my OpenAPI specification that contains a required body. I call `openapi3filter.ValidateRequest` with HTTP request that does not contains a body. I expect that the function will return an error but it returns `nil`.

## Steps to Reproduce the Problem (including complete and simple design if relevant)

```go
package main

import (
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
)

func main() {
	requestBody := openapi3.NewRequestBody().
		WithRequired(true).
		WithJSONSchema(openapi3.NewStringSchema())
	operation := openapi3.NewOperation()
	operation.OperationID = "test"
	operation.RequestBody = &openapi3.RequestBodyRef{Value: requestBody}
	spec.AddOperation("/test", http.MethodPost, operation)

	router := openapi3filter.NewRouter()
	if err := router.AddSwagger(spec); err != nil {
		return nil, err
	}
	
	req, err := http.NewRequest(http.Method, "/test", nil)
	if err != nil {
		return nil, err
	}
	
	route, pathParams, err := router.FindRoute(req.Method, req.URL)
	if err != nil {
		return
	}
	
	input := &openapi3filter.RequestValidationInput{Request: req, PathParams: pathParams, Route: route}
	if err := openapi3filter.ValidateRequest(req.Context(), input); err != nil {
		fmt.Printf("error %[1]T: %s\n", err)
		return
	}
	fmt.Println("no error")
}
```

## Actual Behavior

Function `openapi3filter.ValidateRequest` returns `nil`.

## Expected Behavior

Function `openapi3filter.ValidateRequest` returns `*openapi3filter.RequestError`.

## Specifications

kin-openapi version: commit [9acc5e9](https://github.com/getkin/kin-openapi/commit/9acc5e936becff2d656c4551e6c471d945809678).

fenollp:
Fixed by https://github.com/getkin/kin-openapi/pull/78

PR #78, opened 2019-03-24
Add "required" validation of a request's body

nikolaas:
Proposed fix for issue #77.


openapi3filter/validate_request.go
```diff
@@ -97,58 +98,85 @@ func ValidateParameter(c context.Context, input *RequestValidationInput, paramet
...
+			Err:         err,
+		}
+	}
+
+	// Validate JSON with the schema
+	if err := schemaRef.Value.VisitJSON(value); err != nil {
+		return &RequestError{
+			Input:       input,
+			RequestBody: requestBody,
+			Reason:      "doesn't input the schema",
```
fenollp: You mean `doesn't match the schema`?

openapi3filter/validate_request.go
```diff
@@ -97,58 +98,85 @@ func ValidateParameter(c context.Context, input *RequestValidationInput, paramet
...
-					RequestBody: requestBody,
-					Reason:      "doesn't input the schema",
-					Err:         err,
-				}
-			}
+	inputMIME := req.Header.Get("Content-Type")
+	mediaType := parseMediaType(inputMIME)
+	if !isMediaTypeJSON(mediaType) {
+		// We support validation of JSON data only.
+		return nil
```
fenollp: I'm really not sure about this. People might want to define a "JSON Schema" (openapi3 JSON schema) to validate non-JSON payloads.
Also the OpenAPIv3 spec is starting to support alternative formats for schemas other than the bastardized version of JSON Schema Dratf-04 that's in use today: protocol buffers, XML stuff, and anything really.

So I'm thinking this should be an externally pluggable thing. A public map of MediaType to schema validator func? What do you think?

fenollp: Also on that map idea: if there's a map fetch miss then we should be returning something else than `nil` here.

openapi3filter/validate_request.go
nikolaas: Yes, you are right. That's my bad. I will fix it.

openapi3filter/validate_request.go
fenollp: The minimum we can do about this for now is to return a non nil error here and later implement the public map mechanism or using context or whatever.

nikolaas: Do you mean different formats for schemas or different formats for body's data? I think that support of different data' format will be useful.

I suggest to support content type `application/json` by default, and add function `openapi3filter.RegisterBodyDecoder`. Using this function users can register content types which is needed for them. For example:

```go
func main() {
	// ...

	// Register a body's decoder for content type "application/xml".
	openapi3filter.RegisterBodyDecoder("application/xml", xmlBodyDecoder)

	// Now HTTP request can contain a body with content type "application/xml".
	input := &openapi3filter.RequestValidationInput{
		Request:    httpReq,
		PathParams: pathParams,
		Route:      route,
	}
	if err := openapi3filter.ValidateRequest(ctx, input); err != nil {
		panic(err)
	}
}

func xmlBodyDecoder(body []byte) (interface{}, error) {
	// Decode body to a primitive, []inteface{}, or map[string]interface{}.
}
```

Internally the library will use registered decoders to decode a body of HTTP request and response. After that the decoded value will be validated by JSON Schema.

fenollp: I like this RegisterBodyDecoder thing. I think we still need an additional RegisterBodyValidator too but that may be left as a todo for the future.

nikolaas: Done.

openapi3filter/req_resp_decoder.go
```diff
@@ -595,3 +596,47 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
...
+		return value, nil
+	},
+}
+
+// RegisterBodyDecoder registers a request body's decoder for a content types.
+//
+// If a decoder for the specified content type is already exists, the function replaces
+// it with the specified decoder.
+func RegisterBodyDecoder(contentType string, decoder BodyDecoder) {
+	bodyDecoders[contentType] = decoder
```
fenollp: How about nil-checking `decoder` and panic-ing here? It could be nicer than a later panic?
Your call.

openapi3filter/req_resp_decoder.go
```diff
@@ -595,3 +596,47 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
...
+		}
+		return value, nil
+	},
+}
+
+// RegisterBodyDecoder registers a request body's decoder for a content types.
+//
+// If a decoder for the specified content type is already exists, the function replaces
+// it with the specified decoder.
+func RegisterBodyDecoder(contentType string, decoder BodyDecoder) {
```
fenollp: Nit: we should also have
```golang
// UnregisterBodyDecoder dissociates a body decoder from a content type.
//
// Decoding this content type will result in an error.
func UnregisterBodyDecoder(contentType string) {
	delete(bodyDecoders, contentType)
}
```
Same thing here: if you feel like this should complain on error, add some checks ;)

openapi3filter/validate_request.go
```diff
@@ -139,9 +137,12 @@ func ValidateRequestBody(c context.Context, input *RequestValidationInput, reque
...
 	inputMIME := req.Header.Get("Content-Type")
 	mediaType := parseMediaType(inputMIME)
-	if !isMediaTypeJSON(mediaType) {
-		// We support validation of JSON data only.
-		return nil
+	if mediaType == "" {
+		return &RequestError{
+			Input:       input,
+			RequestBody: requestBody,
+			Reason:      "a content type of the request's body is missed",
```
fenollp: `"content type of request body is missing"`

openapi3filter/validate_request.go
```diff
@@ -159,13 +160,12 @@ func ValidateRequestBody(c context.Context, input *RequestValidationInput, reque
...
-	// Decode JSON
-	var value interface{}
-	if err := json.Unmarshal(data, &value); err != nil {
+	value, err := decodeBody(data, mediaType)
+	if err != nil {
 		return &RequestError{
 			Input:       input,
 			RequestBody: requestBody,
-			Reason:      "decoding JSON failed",
+			Reason:      "failed to decode the request's body",
```
fenollp: `"failed to decode request body"`

openapi3filter/validate_response.go
```diff
@@ -41,87 +40,99 @@ func ValidateResponse(c context.Context, input *ResponseValidationInput) error {
...
+		// An operation does not contains a validation schema for responses with this status code.
+		return nil
+	}
+
+	inputMIME := input.Header.Get("Content-Type")
+	mediaType := parseMediaType(inputMIME)
+	if mediaType == "" {
+		return &ResponseError{
+			Input:  input,
+			Reason: "a content type of the response's body is missed",
```
fenollp: ditto

openapi3filter/req_resp_decoder.go
```diff
@@ -595,3 +596,47 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
...
+	"application/json": func(body []byte) (interface{}, error) {
+		var value interface{}
+		if err := json.Unmarshal(body, &value); err != nil {
+			return nil, err
+		}
+		return value, nil
+	},
+}
+
+// RegisterBodyDecoder registers a request body's decoder for a content types.
```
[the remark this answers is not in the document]
nikolaas: Done. Sorry for my English.

openapi3filter/req_resp_decoder.go
```diff
@@ -595,3 +596,47 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
...
+//
+// If a decoder for the specified content type is already exists, the function replaces
```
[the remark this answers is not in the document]
nikolaas: Done.

openapi3filter/req_resp_decoder.go
nikolaas: Done.

openapi3filter/req_resp_decoder.go
nikolaas: Done.

openapi3filter/validate_request.go
nikolaas: Done.

openapi3filter/validate_request.go
nikolaas: Done.

openapi3filter/validate_response.go
nikolaas: Done.

PR #81, opened 2019-04-01
POC reference single file definitions

ifraixedes:
While trying to use this package with one an OpenAPI v3 spec that I have, I stumbled of not being able to use with it because in how I structured the spec across several files.

In such definition, I have elements which are referenced, but those references are to files where each file only contains the definition of one element and not an entirely OpenAPI definition.

If I understood correctly, according the [OpenAPI reference object section](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#reference-object) which follows the [JSON Reference RFC](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03), the [section 4](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-40) mentions the resolution and the _fragment_ is not mandatory, so it should be fine to define elements in a single file and reference. them.

I wanted to have a try and see if this package was not loading them because of a bug; after a few hours, I've realized that it isn't a bug, it's something not contemplated and I wanted to see if I could quickly implement it with only a few changes, however I wasn't able to do so.

I could have opened an issue regarding this feature, however I though that it was better to show the test which I used for identifying the problem because it clarifies what I explained above and also the dirty implementation, which I quickly wrote for passing the test (which only cover the response reference) for gathering feedback from you about if it could be a possible way in how to tackle the problem.

I may be able to find some time to work in this features in the following weeks, if it's something that you could consider to add, because I would love to have it, so I would appreciate some guidance in how to implemented despite of the way provided in this proof of concept is fine or not.

### DIFF
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -2,6 +2,7 @@ package openapi3
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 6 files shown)


ifraixedes:
Any feedback about this?

fenollp:
Hi! Indeed it is fine to split your OpenAPIv3 spec over more than one file and while this is not something I've personally had the need for (yet) I'd like to see this supported by kin-openapi.

I'm thinking $ref resolving should support local, remote and #/ references where the spec allows.

phamdt:
i'd like to have this supported as well. @ifraixedes, can you update your conflicting files so that perhaps @fenollp would further consider?

disposedtrolley:
This feature would be super valuable for us too. We've written a separate resolver for files, but having it as part of the core library would mean less code to maintain.

fenollp:
Superseeded by #131

PR #87, opened 2019-04-17
Fix crash in parameter validation

mromaszewicz:
Parameters specified using Content instead of Schema crashed the
openapi3 filter. Skip validation of content parameters to fix the
crash.

### DIFF
--- a/openapi3filter/validate_request.go
+++ b/openapi3filter/validate_request.go
@@ -74,6 +74,13 @@ func ValidateRequest(c context.Context, input *RequestValidationInput) error {
 // The function returns RequestError with ErrInvalidRequired cause when a value of a required parameter is not defined.
 // The function returns RequestError with a openapi3.SchemaError cause when a value is invalid by JSON schema.
 func ValidateParameter(c context.Context, input *RequestValidationInput, parameter *openapi3.Parameter) error {
+	// The code below validates parameters with schemas, and will crash on
+	// content-based parameters which have no direct schema. Skip validating
+	// content based ones for now.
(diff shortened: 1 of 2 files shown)


fenollp:
Why hide the issue? What's hard in validating content?

mromaszewicz:
> Why hide the issue? What's hard in validating content?

I started down that path, but it's a big change. I'll get it done eventually as I have time. First, you can't support arbitrary content types, since you don't know what the user is encoding in that parameter, so you need to have a way to register deserialization callbacks for each paramater/content-type combination. Once you have that, you're now validating much more complex objects than the filter code has validated before, so it starts to crash. I stopped at that point since I don't have time to dig much more right now, I'll revisit in a couple of weeks.

For now, you have a crash when there is a content-type parameter, and it's not even in that parameter. The mere existence of a content-type parameter in the spec causes some of the schema traversal code to crash, even if such a parameter isn't specified.

fenollp:
We merged in something along to deserialization callback registries recently. Did you get a look at that?

These crashes sound bad. I'd love to see these go away and the filters and validators get better :)

mromaszewicz:
My unit test modification crashes the filter with the latest commit. That parameter simply being present crashes one of the earlier test cases which doesn't even touch it.

mromaszewicz:
Ok, please check out my alternate PR for this. I polished up my code which adds proper content parameter deserialization. I added a callback to RequestValidationInput for decoding custom non-json parameters.

mromaszewicz:
Closing pull request since I've got another one which implements the missing functionality.

PR #90, opened 2019-04-19
Add content parameter validation

mromaszewicz:
Fix crash in validation of parameters which are specified with
content rather than schema. The problem was that the presence
of a parameter with a nil schema in the Swagger spec would
crash the loop which validated over all parameters.

Added a callback to the request validation input so the user
can decode custom content types, and also implemented an
alternate code path for validation of content arguments. Added
an internal default parameter validator which is capable of
validating JSON encoded parameters.


mromaszewicz:
Hi guys. Any feedback on this PR? I need it to fix some crashing API's and I'd like to avoid forking the repo to do it.

fenollp:
LGTM. The nit can be taken care of in another PR (and it's not the only instance anyway).

This is great work! Sorry for taking so long to review it. Hopefully you'll continue contributing :)

PR #93, opened 2019-05-07
Add content-type wildcard support to validation

### LINKED ISSUE #91: Request validation doesn't support content-type wildcards
The OpenAPI documentation (https://swagger.io/docs/specification/describing-request-body/) describes the content-type keys under `requestBody.content` as supporting wildcards:

> content allows wildcard media types. For example, image/* represents all image types; \*/\* represents all types and is functionally equivalent to application/octet-stream. Specific media types have preference over wildcard media types when interpreting the spec, for example, image/png \> image/* \> \*/\* 

However, the current request validation only supports a direct match of content types via a map lookup. Additionally, the request validation rejects any request without a content-type header which should otherwise be handled by a `*/*` if defined.

My team has a specialized case where this causes validation errors because we need to support a requestor that doesn't add the content-type header. I'm happy to put up a PR that helps with this but thought I might get some feedback on the approach first:

The current lookup is handled [here](https://github.com/getkin/kin-openapi/blob/master/openapi3filter/validate_request.go#L148) with a map[string]*MediaType lookup by key. However, the `requestBody.Content` type has a `Get` method for resolving the type that also incorporates the behavior of `parseMediaType` which is used [here](https://github.com/getkin/kin-openapi/blob/master/openapi3filter/validate_request.go#L139) before the map lookup. I'm guessing the `Get` method was added after the initial implementation of the request validation and that the parsing and map lookup in the validation is a hold-over.

I believe we can hide the complexity of the lookup behavior in the [`Content.Get`](https://github.com/getkin/kin-openapi/blob/master/openapi3/content.go#L26) method. If that approach sounds acceptable then I'll get a PR up that implements the hierarchical wild-card logic.

kconwayatlassian:
This implements the cascading wildcard behavior described in the OpenAPI
specification for request body and response body validation.

https://swagger.io/docs/specification/describing-request-body/

Closes #91

### DIFF
--- a/openapi3/content.go
+++ b/openapi3/content.go
@@ -24,14 +24,35 @@ func NewContentWithJSONSchemaRef(schema *SchemaRef) Content {
 }
 
 func (content Content) Get(mime string) *MediaType {
+	// Start by making the most specific match possible
+	// by using the mime type in full.
 	if v := content[mime]; v != nil {
 		return v
 	}
+	// If an exact match is not found then we strip all
+	// metadata from the mime type and only use the x/y
+	// portion.
 	i := strings.IndexByte(mime, ';')
 	if i < 0 {
-		return nil
+		return content["*/*"]
 	}
-	return content[mime[:i]]
+	mime = mime[:i]
+	if v := content[mime]; v != nil {
+		return v
+	}
+	// If the x/y pattern has no specific match then we
+	// try the x/* pattern.
+	i = strings.IndexByte(mime, '/')
+	if i < 0 {
+		return content["*/*"]
+	}
+	mime = mime[:i] + "/*"
+	if v := content[mime]; v != nil {
+		return v
+	}
+	// Finally, the most generic match of */* is returned
+	// as a catch-all.
+	return content["*/*"]
 }
 
 func (content Content) Validate(c context.Context) error {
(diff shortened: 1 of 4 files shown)


kconwayatlassian:
@fenollp I don't believe mime types are valid without the subtype. However, since they are arbitrary string input from a client I'd prefer if we degraded to the wildcard rather than panic. Alternatively, we can return `nil` for this case instead of the wildcard if that would be preferable. I'm not sure how the OpenAPI specification intends for this case to be handled other than this blerb from the docs:

> \*/\* represents all types and is functionally equivalent to application/octet-stream

This seems to suggest that `*/*` should be used in all cases where a better match is not found.

My concern is that if I'm using the request/response validation in production then a panic on bad input might be an avenue for denial of service if a client is sending invalid mime types. I'd prefer the response to be "400 Bad Request" wherever possible for faulty input.

fenollp:
Agreed wrt panics. 
I’d like ya to return `nil` as this would at least not go unnoticed. I’m sure there’s a nonzero amount of things out there that send an incorrect content type mime type but I very much prefer strictness and to discuss an optional behavior when someone has a use case. 
What do you think?

kconwayatlassian:
@fenollp I've updated the PR to return `nil` for cases of invalid mime types.

Adding the test for this actually exposed a bug in my initial patch where any mime type that did not have metadata such as `;encoding=utf-8` would return the double wildcard result rather than the single wildcard result if there was no direct match. I've increased the test coverage to account for this and patched the logic to behave as expected.

fenollp:
Amazing! Thanks!

PR #94, opened 2019-05-13
Support missing mime types in validation

kconwayatlassian:
When a mime type is missing, rather than present and invalid, we return
the */* match as the request is not invalid for refusing to send a mime
type.

The RFC defining content-type states that including is a `SHOULD` rather than a `MUST`.


kconwayatlassian:
@fenollp I found another edge case in the request/response validation. In this case, it is technically valid for a request or response to _not_ contain the content-type header. This is a case that my team has encountered where we'd like to still apply validation but the current version considers an empty string to be equal to an invalid mime type.

Issue #79, opened 2019-03-28
OAS3: Support for custom formats like `email`, `uuid`, etc for `string` properties?


dsolonenko:
Hello,

From OAS3 spec:
>  Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification.

I was wondering if the validation can be extended to support some reasonably common formats like `email`, `uuid`?

fenollp:
https://godoc.org/github.com/getkin/kin-openapi/openapi3#DefineStringFormat

steffakasid:
Is it possible to configure the validation that it ignores custom defined formats (instead of defining them)?

fenollp:
Same link, look for `SchemaFormatValidationDisabled`.

PR #96, opened 2019-05-29
Define UUID format for string schema

yaronius:
This fixes error `Validating Swagger failed: Unsupported 'format' value 'uuid'`


fenollp:
Yeah the spec https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md
says 
> Primitives have an optional modifier property: format. OAS uses several known formats to define in fine detail the data type being used. However, to support documentation needs, the format property is an open string-valued property, and can have any value. Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification. Types that are not accompanied by a format property follow the type definition in the JSON Schema. Tools that do not recognize a specific format MAY default back to the type alone, as if the format is not specified.

so as resolved on #79 albeit very expeditiously I don't think this lib must validate such formats by default (as opposed to e.g. `date-time`), however I'd love to have the formats defined in-lib and only activated at user's request.

So could you define an exported string constant in https://github.com/getkin/kin-openapi/blob/master/openapi3/schema_formats.go with the name `FormatOfStringForUUIDOfRFC4122`? Or something that has a similarly-easy-to-find-on-godoc-because-sorted prefix.

yaronius:
@fenollp thank you for replying and reasonable remarks, I made changes and now uuid format is optional.

fenollp:
Ah! Sorry I meant for `FormatOfStringForUUIDOfRFC4122` to be the value, not the uuid key :)

yaronius:
@fenollp I see, updated

yaronius:
@fenollp I tested my changes with various schema and I guess this change would not solve the issue with unsupported UUID format because all the supported formats are hardcoded [here](https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L500). I would suggest modifying the default case in `schema.go`:

```
default:
	// Try to check for custom defined formats
	if _, ok := SchemaStringFormats[format]; !ok {
		return unsupportedFormat(format)
	}
```

That works perfectly for cases when we define custom formats and looks flexible to me. What would you say about this?

fenollp:
I believe this is the most correct way indeed! Thanks for taking the time :) 
Add this in and I'll merge the whole thing.

yaronius:
Thanks, changes added

fenollp:
Thank you for your time Zorg! BTW I can easily quote you 90% of this movie (albeit in French) hehe

PR #102, opened 2019-06-14
Security requirements validation

ZakCodes:
When no [security requirements are defined for an operation](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-8), the [global security requirements](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields) need to be used according, which is what I implemented in the first commit.  

The second one modifies the way a security requirement is validated. Before, only the first scheme was validated, but according to the [spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-8), all schemes must be valid for the request to be security requirement to be met.

### DIFF
--- a/openapi3filter/validate_request.go
+++ b/openapi3filter/validate_request.go
@@ -65,9 +65,10 @@ func ValidateRequest(c context.Context, input *RequestValidationInput) error {
 	if security == nil {
 		if input.Route.Swagger == nil {
 			return errRouteMissingSwagger
+		} else {
+			// Use the global security requirements.
+			security = &input.Route.Swagger.Security
 		}
-		// Use the global security requirements.
-		security = &input.Route.Swagger.Security
 	}
 	if security != nil {
 		if err := ValidateSecurityRequirements(c, input, *security); err != nil {
--- a/openapi3filter/validate_request.go
+++ b/openapi3filter/validate_request.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"sort"
+	"sync"
 
 	"github.com/getkin/kin-openapi/openapi3"
 )
@@ -209,14 +210,18 @@ func ValidateSecurityRequirements(c context.Context, input *RequestValidationInp
 		return nil
 	}
 
-	doneChan := make(chan bool, len(srs))
+	var wg sync.WaitGroup
 	errs := make([]error, len(srs))
 
-	// For each alternative
+	// For each alternative security requirement
 	for i, securityRequirement := range srs {
 		// Capture index from iteration variable
 		currentIndex := i
 		currentSecurityRequirement := securityRequirement
+
+		// Add a work item
+		wg.Add(1)
+
 		go func() {
 			defer func() {
 				v := recover()
@@ -226,22 +231,25 @@ func ValidateSecurityRequirements(c context.Context, input *RequestValidationInp
 					} else {
 						errs[currentIndex] = errors.New("Panicked")
 					}
-					doneChan <- false
 				}
+
+				// Remove a work item
+				wg.Done()
 			}()
-			if err := validateSecurityRequirement(c, input, currentSecurityRequirement); err == nil {
-				doneChan <- true
-			} else {
+
+			if err := validateSecurityRequirement(c, input, currentSecurityRequirement); err != nil {
 				errs[currentIndex] = err
-				doneChan <- false
 			}
 		}()
 	}
 
 	// Wait for all
-	for i := 0; i < len(srs); i++ {
-		ok := <-doneChan
-		if ok {
+	wg.Wait()
+
+	// If any security requirement was met
+	for _, err := range errs {
+		if err == nil {
+			// Return no error
 			return nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


openapi3filter/validate_request.go
```diff
@@ -61,6 +61,14 @@ func ValidateRequest(c context.Context, input *RequestValidationInput) error {
 
 	// Security
 	security := operation.Security
+	// If there aren't any security requirements for the operation
+	if security == nil {
+		if input.Route.Swagger == nil {
```
fenollp: Is this case really a possibility by then?

openapi3filter/validate_request.go
```diff
@@ -61,6 +61,14 @@ func ValidateRequest(c context.Context, input *RequestValidationInput) error {
...
+			return errRouteMissingSwagger
+		}
+		// Use the global security requirements.
+		security = &input.Route.Swagger.Security
+	}
 	if security != nil {
```
fenollp: Let’s turn this into an else clause.

openapi3filter/validate_request.go
```diff
@@ -283,12 +291,15 @@ func validateSecurityRequirement(c context.Context, input *RequestValidationInpu
...
 		}
 		scopes := securityRequirement[name]
-		return f(c, &AuthenticationInput{
+		err := f(c, &AuthenticationInput{
 			RequestValidationInput: input,
 			SecuritySchemeName:     name,
 			SecurityScheme:         securityScheme,
 			Scopes:                 scopes,
 		})
+		if err != nil {
```
fenollp: Looks like you can inline this check with the line above.

openapi3filter/validate_request.go
ZakCodes: I am not sure about which case you're talking about since there are two if statements in this block. However, I believe that both are possible since neither of them have been checked. We don't have to look far to see whether the check has been made since this function is exported, we can just have to look whether the check is made inside of this function and it isn't.

openapi3filter/validate_request.go
ZakCodes: I've done it in d9d4d95

ZakCodes:
I'm sorry for not writing any tests, I don't know how to write them in Go. However, I'll figure it out and do it during this upcoming week.

ZakCodes:
Sorry for the long delay, I was a busy with my new job and some other projects.  

Ironically enough, the code that I had written made the test coverage go up instead of down, however my features weren't tested, so I've written 3 tests to make sure they work.  

The test `TestAnySecurityRequirementMet` revealed a bug with the `doneChan` where some threads were writing to it after it was closed. I've looked around and found this [stackoverflow thread](https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open) explaining that it isn't necessary to close a channel unless the receiver is waiting to receive the close signal (like for a `for range`) which isn't the case here. Therefore, I simply removed the line to close the channel and left the channel to be garbage collected, which solved the bug.  
You can revert the last commit that I made solving this bug if you want to see it.  

`validate_request.go` now has 77.4% coverage instead of 47.8%.

openapi3filter/validation_test.go
```diff
@@ -347,3 +354,369 @@ func toJSON(v interface{}) io.Reader {
...
+		require.NoError(t, err)
+		req := openapi3filter.RequestValidationInput{
+			Request: httptest.NewRequest(http.MethodGet, path.name, emptyBody),
+			Route:   route,
+			Options: &openapi3filter.Options{
+				AuthenticationFunc: func(c context.Context, input *openapi3filter.AuthenticationInput) error {
+					if schemesValidated != nil {
+						if validated, ok := (*schemesValidated)[input.SecurityScheme]; ok {
+							if validated {
+								t.Fatalf("The path \"%s\" had the schemes %v named \"%s\" validated more than once",
```
fenollp: `%q` does exactly the same as `\"%s\"` for strings ;)

ZakCodes: I haven't taken properly learned go yet, so I didn't know that. Thanks a lot!

ZakCodes:
I have done what you've asked. However, I was wondering if there was any problem when we were leaving other threads running after returning from the function that created them? Because I've tried adding some more security requirement to see if it caused any issue and it didn't seem to.  
However, maybe the calling function waits for all the threads it created to return before ending in case one of the created thread has a reference to one of its stack allocated variables. Because if that stack allocated variable was destroyed, it would make the created thread crash.

openapi3filter/validate_request.go
```diff
@@ -200,14 +210,18 @@ func ValidateSecurityRequirements(c context.Context, input *RequestValidationInp
...
 
-	// For each alternative
+	// For each alternative security requirement
 	for i, securityRequirement := range srs {
 		// Capture index from iteration variable
 		currentIndex := i
 		currentSecurityRequirement := securityRequirement
+
+		// Add a work item
+		wg.Add(1)
```
fenollp: This can be replaced by `wg.Add(len(srs))` before the loop. It supposedly saves a bit of time.

fenollp:
Thanks! I have added another nit (along with the `%q` thing) so if you find time to tackle that feel free to open another PR :)

PR #104, opened 2019-07-03
Added default option for Parameters

Andrei-Predoiu:
The default field is missing from parameter definition in V2


Andrei-Predoiu:
@jban332 Can you take a look

tschaub:
This looks like it breaks things if the parameter type is not a string.  From the [Swagger 2 docs](https://swagger.io/docs/specification/describing-parameters/#default) it looks like other types (e.g. `integer`) should be allowed here too.

tschaub:
See #137 for a proposed fix.

PR #114, opened 2019-09-06
Validate additional OpenAPI objects

disposedtrolley:
Hello! I'm incorporating this library into an OpenAPI validation and management service and have noticed that the included validation functions don't check the `openapi` version object (i.e. it can be omitted and validation succeeds).

I've made the following changes in an attempt to address this:

- Adds validation to the `openapi` object
- Removes redundant validation on the `paths` object
- Annotates validation errors in nested objects

@nogates

### DIFF
--- a/openapi2/openapi2.go
+++ b/openapi2/openapi2.go
@@ -129,7 +129,7 @@ type Operation struct {
 	Tags         []string               `json:"tags,omitempty"`
 	OperationID  string                 `json:"operationId,omitempty"`
 	Parameters   Parameters             `json:"parameters,omitempty"`
-	Responses    map[string]*Response   `json:"responses,omitempty"`
+	Responses    map[string]*Response   `json:"responses"`
 	Consumes     []string               `json:"consumes,omitempty"`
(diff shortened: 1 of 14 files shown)


fenollp:
Amazing!
Could you add tests for the error cases you're adding support for?
Also, why make `info` `omitempty`-able?

disposedtrolley:
> Could you add tests for the error cases you're adding support for?
Yup sure thing!

> Also, why make `info` `omitempty`-able?
Must've been an oversight on my part. I saw that all of the properties within the Info struct were optional, so I made the entire thing optional. I've just reviewed the [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#info-object) and I can see that the `title` and `version` properties of the `info` object are actually mandatory, so I'll make the changes to align to that.

disposedtrolley:
@fenollp I've added tests for the new errors we're returning and also have added validation to some more objects. I think it might be ready to be merged! Let me know if I should add anything else

fenollp:
I'm throwing it again: amazing work! Thanks

PR #119, opened 2019-09-30
$ref in paths block

aokht:
Hi! Thank you for the great library.

I'd like to use `$ref` in `paths` block.
I know we already have an opening pr for it https://github.com/getkin/kin-openapi/pull/73
But unfortunately it doesn't seem active with some change requests.
Then I tried to make another implementation.
I would be happy if this pr helps ones waiting for this feature :)

As you can see in the commit list, I made 3 changes;
1. Divided resolveComponent() into swagger part and component part
2. Supported `$ref` in `paths` block
3. Reset loader's internal state except from internal loading call

3rd change is needed, because in 2nd change, we need to have global state to avoid infinite recursion although we can use one SwaggerLoader for multiple loadings.

I also made a unit test case for this feature, and checked all tests passed.

Thank you for reviewing :)

### DIFF
--- a/openapi3/path_item.go
+++ b/openapi3/path_item.go
@@ -22,6 +22,7 @@ type PathItem struct {
 	Trace       *Operation `json:"trace,omitempty"`
 	Servers     Servers    `json:"servers,omitempty"`
 	Parameters  Parameters `json:"parameters,omitempty"`
+	Ref         string     `json:"$ref,omitempty"`
 }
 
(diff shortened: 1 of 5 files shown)


fenollp:
Indeed https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#fixed-fields-7
LGTM. Thanks!

PR #120, opened 2019-10-02
Add YAML tags to attributes of all structs

disposedtrolley:
This PR adds YAML tags to attributes of all structs to support marshaling with the [go-yaml](https://github.com/go-yaml/yaml) package.

**NB** the branch being merged in also contains the changes in the [validation PR](https://github.com/getkin/kin-openapi/pull/114). The only commit of interest for this PR is [bc7441a](https://github.com/getkin/kin-openapi/commit/bc7441a03cdcd243f1e53f05774181d249ab9e34)

PR #121, opened 2019-10-02
Add YAML tags to all exported struct attributes

disposedtrolley:
This PR adds `yaml` tags to all exported struct attributes. This provides compatibility with alternate YAML parsers like [go-yaml/yaml](https://github.com/go-yaml/yaml).

### DIFF
--- a/openapi3/components.go
+++ b/openapi3/components.go
@@ -11,16 +11,16 @@ import (
 // Components is specified by OpenAPI/Swagger standard version 3.0.
 type Components struct {
 	ExtensionProps
-	Schemas         map[string]*SchemaRef         `json:"schemas,omitempty"`
-	Parameters      map[string]*ParameterRef      `json:"parameters,omitempty"`
-	Headers         map[string]*HeaderRef         `json:"headers,omitempty"`
-	RequestBodies   map[string]*RequestBodyRef    `json:"requestBodies,omitempty"`
-	Responses       map[string]*ResponseRef       `json:"responses,omitempty"`
-	SecuritySchemes map[string]*SecuritySchemeRef `json:"securitySchemes,omitempty"`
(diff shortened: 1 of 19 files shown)


fenollp:
Re-asking https://github.com/getkin/kin-openapi/pull/120#issuecomment-537425581
Why not use https://github.com/ghodss/yaml?

disposedtrolley:
> Why not use https://github.com/ghodss/yaml?

This library doesn't retain the ordering of the struct's attributes when marshaling to YAML as it unmarshals into a map during the process. For example, the resulting YAML document from marshaling `docA` in `swagger_test.go` looks like this:

```
components:
  examples:
    otherExample:
      value:
        name: Some example
    someExample:
      $ref: '#/components/examples/otherExample'
  headers:
    otherHeader: {}
    someHeader:
      $ref: '#/components/headers/otherHeader'
  parameters:
    someParameter:
      description: Some parameter
      in: query
      name: example
      schema:
        $ref: '#/components/schemas/someSchema'
  requestBodies:
    someRequestBody:
      description: Some request body
  responses:
    someResponse:
      description: Some response
  schemas:
    someSchema:
      description: Some schema
  securitySchemes:
    otherSecurityScheme:
      description: Some security scheme
      in: query
      name: token
      type: apiKey
    someSecurityScheme:
      $ref: '#/components/securitySchemes/otherSecurityScheme'
info:
  title: MyAPI
  version: "0.1"
openapi: "3.0"
paths:
  /hello:
    parameters:
    - $ref: '#/components/parameters/someParameter'
    post:
      parameters:
      - $ref: '#/components/parameters/someParameter'
      requestBody:
        $ref: '#/components/requestBodies/someRequestBody'
      responses:
        "200":
          $ref: '#/components/responses/someResponse'
```

While I understand the YAML spec doesn't require ordering, it would be nice in the case of OpenAPI files.

PR #137, opened 2019-11-01
Allow any type of default and serialize v3 parameter schema

tschaub:
This makes it so the [v2 parameter default](https://swagger.io/docs/specification/2-0/describing-parameters/#default) is included as the [v3 parameter schema default](https://swagger.io/docs/specification/describing-parameters/#default).

Also includes a fix for serializing the v3 parameter schema max as a v2 parameter maximum.

Test change covers both.


fenollp:
Thanks!

PR #139, opened 2019-11-01
Use the full version specifier

tschaub:
The [OpenAPI spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md) requires that the full version identifier be used.


fenollp:
How about setting `3.0.2`?

tschaub:
Updated to use `3.0.2`.

PR #142, opened 2019-11-08
Add Vulcain to the list of projects using kin-openapi

dunglas:
Thanks for this library! Vulcain uses kin-openapi especially for this part: https://github.com/dunglas/vulcain/blob/master/docs/gateway/openapi.md


fenollp:
Thanks! Note that the two issues linked in that document could use your input ;)

Issue #143, opened 2019-11-18
Verify unique of array got panic when the array item is type of object


emnnqqs:
I have defined below array in openapi:
```
    Tests:
      type: array
      items:
        $ref: '#/components/schemas/Test'
      minItems: 0
      uniqueItems: true
    Test:
      type: object
      properties:
        attri1:
          type: integer
        attri2:
          type: integer
```
When i verify it by openapi3filter, I got below error:
```
panic: runtime error: hash of unhashable type map[string]interface {} [recovered]
	panic: runtime error: hash of unhashable type map[string]interface {}

goroutine 8 [running]:
testing.tRunner.func1(0xc0000d8200)
	C:/Go/src/testing/testing.go:874 +0x3aa
panic(0x7ea040, 0xc0001a6970)
	C:/Go/src/runtime/panic.go:679 +0x1c0
github.com/getkin/kin-openapi/openapi3.isSliceOfUniqueItems(0xc0001af380, 0x2, 0x2, 0x1)
	C:/Users/emnnqqs/git-repository/gopath/src/github.com/getkin/kin-openapi/openapi3/schema.go:1227 +0x187
github.com/getkin/kin-openapi/openapi3.(*Schema).visitJSONArray(0xc00019a000, 0xc0001af380, 0x2, 0x2, 0x0, 0x0, 0x2dd0008)
	C:/Users/emnnqqs/git-repository/gopath/src/github.com/getkin/kin-openapi/openapi3/schema.go:985 +0x8f2
github.com/getkin/kin-openapi/openapi3.(*Schema).visitJSON(0xc00019a000, 0x7bbca0, 0xc0001af3a0, 0xc00008d600, 0xb7d770, 0xc00008d5f0)
	C:/Users/emnnqqs/git-repository/gopath/src/github.com/getkin/kin-openapi/openapi3/schema.go:624 +0x5e2
github.com/getkin/kin-openapi/openapi3.(*Schema).VisitJSON(0xc00019a000, 0x7bbca0, 0xc0001af3a0, 0xa, 0xc00008d6b0)
	C:/Users/emnnqqs/git-repository/gopath/src/github.com/getkin/kin-openapi/openapi3/schema.go:593 +0x4b
github.com/getkin/kin-openapi/openapi3.(*Schema).visitJSONObject(0xc0000cc8c0, 0xc0001a8d20, 0xc0001a8d00, 0xc0001a6800, 0x0)
[...]

PR #145, opened 2019-11-20
Fix panic when verify array unique items(#143)

Arms-Li:
json.Unmarshal will store JSON object as map[string] interface{}
and store JSON array as []interface{}. Due to map and slice can
not be the key of map, so if the array items' type is JSON object
or JSON array, verify unique items of array will panic due map
and slice are unhashable.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"reflect"
 	"regexp"
 	"strconv"
 	"unicode/utf16"
@@ -1222,6 +1223,19 @@ func (err *SchemaError) Error() string {
 func isSliceOfUniqueItems(xs []interface{}) bool {
 	s := len(xs)
 	m := make(map[interface{}]struct{}, s)
+	xsKind := reflect.TypeOf(xs[0]).Kind()
+	if xsKind == reflect.Map ||
+		xsKind == reflect.Slice {
+		for i := 0; i < s; i++ {
+			m[fmt.Sprintf("%v", xs[i])] = struct{}{}
+		}
+		return s == len(m)
+	}
+
+	//  - reflect.Float64
+	//  - reflect.Bool
+	//  - reflect.String
+
 	for _, x := range xs {
 		m[x] = struct{}{}
 	}
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"math"
 	"math/big"
-	"reflect"
 	"regexp"
 	"strconv"
 	"unicode/utf16"
@@ -983,7 +982,7 @@ func (schema *Schema) visitJSONArray(value []interface{}, fast bool) (err error)
 	}
 
 	// "uniqueItems"
-	if v := schema.UniqueItems; v && !isSliceOfUniqueItems(value) {
+	if v := schema.UniqueItems; v && !sliceUniqueItemsChecker(value) {
 		if fast {
 			return errSchema
 		}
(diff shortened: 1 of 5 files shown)


fenollp:
Fixes https://github.com/getkin/kin-openapi/issues/143

openapi3/schema.go
```diff
@@ -1222,6 +1223,19 @@ func (err *SchemaError) Error() string {
 func isSliceOfUniqueItems(xs []interface{}) bool {
 	s := len(xs)
 	m := make(map[interface{}]struct{}, s)
+	xsKind := reflect.TypeOf(xs[0]).Kind()
+	if xsKind == reflect.Map ||
+		xsKind == reflect.Slice {
+		for i := 0; i < s; i++ {
+			m[fmt.Sprintf("%v", xs[i])] = struct{}{}
```
fenollp: This can collide with strings. How about making a `map[string]struct{}` and creating keys with `json.Marshal`? This allows dropping `reflect` too.

Arms-Li: The reason why I use reflect to distinguish Map and slice and also choose fmt.Sprintf is due to I do some benchmark test, result as below:
When input array item is type string or number, the array legnth is 10:
```
BenchmarkIsSliceOfUniqueItemsOnlyMapSliceUseJSONMarshal-8   	 2503779	       475 ns/op	     291 B/op	       1 allocs/op
BenchmarkIsSliceOfUniqueItemsOnlyMapSliceUseSprintf-8       	 2524917	       484 ns/op	     291 B/op	       1 allocs/op
BenchmarkIsSliceOfUniqueItemsAllJSONMarshal-8               	  442516	      2719 ns/op	     387 B/op	      12 allocs/op
BenchmarkIsSliceOfUniqueItemsAllSprintf-8                   	 1000000	      1104 ns/op	     291 B/op	       1 allocs/op
```
When input array item is type array,outter array length is 10,  inner array is 2:
```
BenchmarkIsSliceOfUniqueItemsOnlyMapSliceJSONMarshal-8   	  225642	      5381 ns/op	    1091 B/op	      31 allocs/op
BenchmarkIsSliceOfUniqueItemsOnlyMapSliceSprintf-8       	  254638	      4682 ns/op	     771 B/op	      21 allocs/op
BenchmarkIsSliceOfUniqueItemsAllJSONMarshal-8            	  249345	      4922 ns/op	     947 B/op	      22 allocs/op
BenchmarkIsSliceOfUniqueItemsAllSprintf-8                	  298896	      4233 ns/op	     611 B/op	      11 allocs/op
```

Acctualy, I would like suggest we can let user register their own function used to check unique of array, which might have better performance when that function is designed for specific input data.

fenollp: Agreed & I think this is a very good idea. Can you provide a default registered function that uses json.Marshal? Thanks

Arms-Li: New codes provided.

openapi3filter/validate.go
```diff
@@ -0,0 +1,11 @@
+package openapi3filter
+
+import (
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+// RegisterArrayUniqueItemsChecker is used to register a customized function
+// used to check if JSON array have unique items.
+func RegisterArrayUniqueItemsChecker(fn func(items []interface{}) bool) {
```
fenollp: Why repeat this here? Isn't the definition in package `openapi3` enough?

openapi3/schema.go
```diff
@@ -1221,13 +1221,31 @@ func (err *SchemaError) Error() string {
...
+type SliceUniqueItemsChecker func(items []interface{}) bool
+
+// By default using predefined func isSliceOfUniqueItems which make use of
+// json.Marshal to generate a key for map used to check if a given slice
+// have unique items.
+var sliceUniqueItemsChecker SliceUniqueItemsChecker = isSliceOfUniqueItems
+
+// RegisterArrayUniqueItemsChecker is used to register a customized function
+// used to check if JSON array have unique items.
+func RegisterArrayUniqueItemsChecker(fn func(items []interface{}) bool) {
```
fenollp: Can you use the `SliceUniqueItemsChecker` type you defined above?

Arms-Li: ok，no problem！

openapi3filter/validate.go
Arms-Li: From my point of view this function is used for validation， here is openapi3filter，namespace，so I think it’s better register function from openapi3filter. what you think?

fenollp: I prefer to not duplicate names + all schema validation code is already within `openapi3`. I'd rather keep it there & not defining this here.

You've documented some of this in the main README but maybe you should also mention this in other places that make sense to you? Maybe `openapi3filter` docs & `openapi3` validation docs?

Arms-Li: Do you mean add new README file under openapi3 folder?

fenollp: I was thinking of possibly adding documentation to this line: https://github.com/getkin/kin-openapi/blob/faf7556f33b5c39d08d7a54efbad5f13f319fa49/openapi3filter/validate_response.go#L14
Such as:
> ValidateResponse returns a non-nil error if the ResponseValidationInput does not match the OpenAPIv3 spec.
Note: one can tune the behavior of `uniqueItems: true` verification by registering a custom function with openapi3.RegisterArrayUniqueItemsChecker ...

fenollp:
Amazing! Thanks

Issue #147, opened 2019-11-28
Extension contains all fields


sayboras:
I am trying to load the simple yaml file and print all extension `x-***` field using below method

```golang
	swagger, e := openapi3.NewSwaggerLoader().LoadSwaggerFromFile("swagger.yaml")
	if e != nil {
		fmt.Println(e)
	}

	for k, v := range swagger.Paths.Find("/pets").Get.Extensions {
		fmt.Printf("key : %s, value %s\n", k, v)
	}
```

Surprisingly, the above map contains both `x-` fields and normal fields. Please find below the sample output. Kindly let me know if it's expected or is there any other way that we could do it.

```
key : summary, value "List all pets"
key : tags, value ["pets"]
key : x-api-limit, value 20
...
```

Swagger file 
```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    get:
      x-api-limit: 20
      summary: List all pets
      operationId: listPets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            format: int32
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pets"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Pets:
      type: array
      items:
        $ref: "#/components/schemas/Pet"
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

fenollp:
Fixed by https://github.com/getkin/kin-openapi/pull/148 Thanks!

sayboras:
Closed this issue :)

PR #148, opened 2019-11-28
Fix issue with x-* fields Unmarshal

sayboras:
Addressed https://github.com/getkin/kin-openapi/issues/147

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,5 +8,3 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
-
-go 1.13
--- a/go.mod
+++ b/go.mod
@@ -8,3 +8,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+go 1.13
--- a/openapi3/extension.go
+++ b/openapi3/extension.go
@@ -30,14 +30,10 @@ func (props *ExtensionProps) DecodeWith(decoder *jsoninfo.ObjectDecoder, value i
 		return err
 	}
 	source := decoder.DecodeExtensionMap()
-	if len(source) > 0 {
-		result := make(map[string]interface{}, len(source))
-		for k, v := range source {
-			result[k] = v
-		}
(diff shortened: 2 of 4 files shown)


fenollp:
Could you add a test for this? Could you also return on non-nil `err` before mutating `props`?

sayboras:
> Could you add a test for this? Could you also return on non-nil `err` before mutating `props`?

Thanks for your feedback. Let me make an update :)

sayboras:
@fenollp I have added unit test cases, which I think is required.

openapi3/extension_test.go
```diff
@@ -0,0 +1,95 @@
...
+				"field2": "value2",
+			},
+		}
+
+		var value = &struct {
+			Field1 string `json:"field1"`
+			Field2 string `json:"field2"`
+		}{}
+
+		err = extensionProps.DecodeWith(decoder, value)
```
fenollp: Please verify `value` in this and the following tests

openapi3/extension.go
```diff
@@ -25,13 +25,15 @@ func (props *ExtensionProps) EncodeWith(encoder *jsoninfo.ObjectEncoder, value i
 
 // DecodeWith will be invoked by package "jsoninfo"
 func (props *ExtensionProps) DecodeWith(decoder *jsoninfo.ObjectDecoder, value interface{}) error {
+	err := decoder.DecodeStructFieldsAndExtensions(value)
+	if err != nil {
```
fenollp: You can write this in one line:
```golang
	if err := decoder.DecodeStructFieldsAndExtensions(value); err != nil {
```

go.mod
```diff
@@ -8,3 +8,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+go 1.13
```
fenollp: This does not need to be in this PR. Please revert.

sayboras: yeah, my IDE added this, and i committed the change by mistake :(

openapi3/extension_test.go
sayboras: Thanks, I have added assert statement, another look if you don't mind 💯

openapi3/extension.go
sayboras: Updated.

PR #150, opened 2019-12-01
add anyof, allof, oneof in parameter

### LINKED ISSUE #106: query param with anyOf failed to parse
query param with anyOf failed to parse

Openapi yaml with:

paths:
/resource-store:
get:
operationId: SearchResources
tags:
- Resource Store)
parameters:
- name: resource-type
in: query
description: Type of the resource to get
required: true
schema:
$ref: '#/components/schemas/RESOURCEType'
responses:
xxxx

components:
schemas:
RESOURCEType:
anyOf:
- type: string
enum:
- MEM
- CPU
- DISK
- type: string

====
when I succeeded to build and run the server, it panics at requests with this message:
echo: http: panic serving [::1]:56426: schema has non primitive type ""
goroutine 12 [running]:
net/http.(*conn).serve.func1(0xc000088140)
/usr/local/go/src/net/http/server.go:1769 +0x139
panic(0x86c960, 0xc0004740a0)
/usr/local/go/src/runtime/panic.go:522 +0x1b5
github.com/getkin/kin-openapi/openapi3filter.parsePrimitive(0xc00002c6e0, 0x3, 0xc00022cd00, 0x23, 0xc00005b7b8, 0x3, 0x3)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/req_resp_decoder.go:705 +0xa1e
github.com/getkin/kin-openapi/openapi3filter.(*urlValuesDecoder).DecodePrimitive(0xc000010040, 0xc000161fe0, 0xd, 0xc000466080, 0xc00022cd00, 0x8, 0xc000010040, 0x5, 0x0)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/req_resp_decoder.go:412 +0x329
github.com/getkin/kin-openapi/openapi3filter.decodeValue(0x9c6ba0, 0xc000010040, 0xc000161fe0, 0xd, 0xc000466080, 0xc00022cd00, 0x0, 0x0, 0x0, 0x0)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/req_resp_decoder.go:243 +0xcf
github.com/getkin/kin-openapi/openapi3filter.decodeStyledParameter(0xc0000e0500, 0xc00045c450, 0x0, 0x0, 0x0, 0x0)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/req_resp_decoder.go:225 +0x134
github.com/getkin/kin-openapi/openapi3filter.ValidateParameter(0x9c7f20, 0xc00045c4b0, 0xc00045c450, 0xc0000e0500, 0x0, 0x0)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/validate_request.go:95 +0x350
github.com/getkin/kin-openapi/openapi3filter.ValidateRequest(0x9c7f20, 0xc00045c4b0, 0xc00045c450, 0x9b4cf0, 0x9117e0)
/home/admin/go/src/github.com/getkin/kin-openapi/openapi3filter/validate_request.go:49 +0x243
github.com/deepmap/oapi-codegen/pkg/middleware.ValidateRequestFromContext(0x9d4600, 0xc0000de5a0, 0xc000142b60, 0x0, 0x5d259d4c, 0xb0915dc)
/home/admin/go/src/github.com/deepmap/oapi-codegen/pkg/middleware/oapi_validate.go:115 +0x537
github.com/deepmap/oapi-codegen/pkg/middleware.OapiRequestValidatorWithOptions.func1.1(0x9d4600, 0xc0000de5a0, 0xd09620, 0xc00005bc28)
/home/admin/go/src/github.com/deepmap/oapi-codegen/pkg/middleware/oapi_validate.go:69 +0x54
github.com/labstack/echo/middleware.LoggerWithConfig.func2.1(0x9d4600, 0xc0000de5a0, 0x2, 0x2)
/home/admin/go/src/github.com/labstack/echo/middleware/logger.go:119 +0x1c0
github.com/labstack/echo.(*Echo).ServeHTTP(0xc000224380, 0x9c6fa0, 0xc0000c6000, 0xc0000e8000)
/home/admin/go/src/github.com/labstack/echo/echo.go:616 +0x22d
net/http.serverHandler.ServeHTTP(0xc00014a9c0, 0x9c6fa0, 0xc0000c6000, 0xc0000e8000)
/usr/local/go/src/net/http/server.go:2774 +0xa8
net/http.(*conn).serve(0xc000088140, 0x9c7e60, 0xc000458840)

The request is:
curl "localhost:8080/resource-store?resource-type=CPU"

I solved the issue by simply declaring:
RESOURCEType:
type: string

Still, I'd like to understand the cause behind the former syntax and the possibility to get the former syntax work.

Thanks,
Charlie

Luffy110:
Fix #106 query param with anyOf failed to parse . Now I have added AnyOf , OneOf and AllOf but haven't Not. because for Not ,i have no idea currently.

### DIFF
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -208,8 +208,14 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
 	var dec valueDecoder
 	switch param.In {
 	case openapi3.ParameterInPath:
+		if input.PathParams == nil {
+			return nil, nil
+		}
 		dec = &pathParamDecoder{pathParams: input.PathParams}
 	case openapi3.ParameterInQuery:
+		if len(input.GetQueryParams()) == 0 {
+			return nil, nil
+		}
 		dec = &urlValuesDecoder{values: input.GetQueryParams()}
 	case openapi3.ParameterInHeader:
 		dec = &headerParamDecoder{header: input.Request.Header}
@@ -219,51 +225,58 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
 		return nil, fmt.Errorf("unsupported parameter's 'in': %s", param.In)
 	}
 
-	return decodeValue(dec, param.Name, sm, param.Schema)
+	return decodeValue(dec, param.Name, sm, param.Schema, param.Required)
 }
 
-func decodeValue(dec valueDecoder, param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error) {
+func decodeValue(dec valueDecoder, param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef, required bool) (interface{}, error) {
 	var decodeFn func(param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error)
-	if len(schema.Value.AnyOf) > 0 {
-		for _, sr := range schema.Value.AnyOf {
-			value, err := decodeValue(dec, param, sm, sr)
-			if value == nil && err == nil {
-				continue
-			}
-			return value, err
-		}
-		return nil, nil
-	}
+
 	if len(schema.Value.AllOf) > 0 {
 		var value interface{}
 		var err error
 		for _, sr := range schema.Value.AllOf {
-			value, err = decodeValue(dec, param, sm, sr)
+			value, err = decodeValue(dec, param, sm, sr, required)
 			if value == nil || err != nil {
 				break
 			}
 		}
 		return value, err
 	}
+
+	if len(schema.Value.AnyOf) > 0 {
+		for _, sr := range schema.Value.AnyOf {
+			value, _ := decodeValue(dec, param, sm, sr, required)
+			if value != nil {
+				return value, nil
+			}
+		}
+		if required == true {
+			return nil, fmt.Errorf("decode AnyOf failed " + param)
+		} else {
+			return nil, nil
+		}
+
+	}
+
 	if len(schema.Value.OneOf) > 0 {
 		isMatched := 0
 		var value interface{}
 		for _, sr := range schema.Value.OneOf {
-			value, err := decodeValue(dec, param, sm, sr)
-			if err != nil {
-				return nil, err
-			}
-			if value != nil {
+			v, _ := decodeValue(dec, param, sm, sr, required)
+			if v != nil {
+				value = v
 				isMatched++
 			}
 		}
 		if isMatched == 1 {
 			return value, nil
+		} else if isMatched > 1 {
+			return nil, fmt.Errorf("decode Oneof failed, matched times:%d", isMatched)
 		}
-		if len(schema.Value.Required) == 0 {
-			return nil, nil
+		if required == true {
+			return nil, fmt.Errorf("decode Oneof failed, " + param + "is required")
 		} else {
-			return nil, errors.New("decode Oneof failed")
+			return nil, nil
 		}
 	}
 	// TODO: if schema.Value.Not isn't nil , how to deal with it?
@@ -866,7 +879,7 @@ func urlencodedBodyDecoder(body io.Reader, header http.Header, schema *openapi3.
 		}
 		sm := enc.SerializationMethod()
 
-		if value, err = decodeValue(dec, name, sm, prop); err != nil {
+		if value, err = decodeValue(dec, name, sm, prop, false); err != nil {
 			return nil, err
 		}
 		obj[name] = value
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -224,20 +224,62 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
 
 func decodeValue(dec valueDecoder, param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error) {
 	var decodeFn func(param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error)
-	switch schema.Value.Type {
-	case "array":
-		decodeFn = func(param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error) {
-			return dec.DecodeArray(param, sm, schema)
+	if len(schema.Value.AnyOf) > 0 {
+		for _, sr := range schema.Value.AnyOf {
+			value, err := decodeValue(dec, param, sm, sr)
+			if value == nil && err == nil {
+				continue
+			}
+			return value, err
 		}
-	case "object":
-		decodeFn = func(param string, sm *openapi3.SerializationMethod, schema *openapi3.SchemaRef) (interface{}, error) {
-			return dec.DecodeObject(param, sm, schema)
-		}
-	default:
-		decodeFn = dec.DecodePrimitive
+		return nil, nil
 	}
+	if len(schema.Value.AllOf) > 0 {
+		var value interface{}
+		var err error
+		for _, sr := range schema.Value.AllOf {
+			value, err = decodeValue(dec, param, sm, sr)
+			if value == nil || err != nil {
+				break
+			}
+		}
+		return value, err
+	}
+	if len(schema.Value.OneOf) > 0 {
+		isMatched := 0
+		var value interface{}
+		for _, sr := range schema.Value.OneOf {
+			value, err := decodeValue(dec, param, sm, sr)
+			if err != nil {
+				return nil, err
+			}
+			if value != nil {
+				isMatched++
+			}
+		}
+		if isMatched == 1 {
+			return value, nil
+		}
+		return nil, nil
+	}
+	// TODO: if schema.Value.Not isn't nil , how to deal with it?
 
-	return decodeFn(param, sm, schema)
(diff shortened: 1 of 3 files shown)


openapi3filter/req_resp_decoder.go
```diff
@@ -224,20 +224,65 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
...
+		}
+		return nil, nil
+	}
+	if len(schema.Value.AllOf) > 0 {
+		isMatched := true
+		var value interface{}
+		var err error
+		for _, sr := range schema.Value.AllOf {
+			value, err = decodeValue(dec, param, sm, sr)
+			if value == nil {
```
fenollp: Shouldn't this be checking `err` instead, as `value` can be `nil` for `type: null` values?

openapi3filter/req_resp_decoder.go
```diff
@@ -224,20 +224,65 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
...
-			return dec.DecodeArray(param, sm, schema)
+	if len(schema.Value.AnyOf) > 0 {
+		for _, sr := range schema.Value.AnyOf {
+			value, err := decodeValue(dec, param, sm, sr)
+			if value == nil && err == nil {
+				continue
+			}
+			return value, err
+		}
+		return nil, nil
```
fenollp: Shouldn't decoding continue & verify the other possible fields sibling to `anyOf`?

openapi3filter/req_resp_decoder.go
```diff
@@ -224,20 +224,65 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
...
+		isMatched := 0
+		var value interface{}
+		var err error
+		for _, sr := range schema.Value.OneOf {
+			value, err := decodeValue(dec, param, sm, sr)
+			if value != nil || err != nil {
+				isMatched++
+			}
+		}
+		if isMatched > 1 {
```
fenollp: > oneOf – validates the value against exactly one of the subschemas

This should be `matched == 1`.

openapi3filter/req_resp_decoder.go
Luffy110: decodeValue function will return nil,nil if type is null. here judge "value==nil" means the value doesn't match this schema rule,and no longer to match other schema rules. Yes ,maybe here have a bug. when err isn't nil.

openapi3filter/req_resp_decoder.go
Luffy110: Do you mean here will check all of anyof schema rules? even if matched one schema ?

openapi3filter/req_resp_decoder.go
Luffy110: No , i think if matched ==1 , it means this check is ok. it will return value, err  ,maybe code is 
`if isMatched == 1{
  return value, err
}
return nil,nil`

openapi3filter/req_resp_decoder.go
fenollp: I mean given the legal schema:
```yaml
anyOf:
- format: ipv4
- format: ipv4
- pattern: '^[a-zA-Z_]+$'
minLength: 3
```
its `minLength: 3` won't be used for decoding. I can't think of a better example right now but maybe you get the idea?

openapi3filter/req_resp_decoder.go
fenollp: Could checking `err` before `value` help?

openapi3filter/req_resp_decoder.go
fenollp: Oh I see. Actually this `> 1` check should be returning an error, no?

openapi3filter/req_resp_decoder.go
Luffy110: according to current code ,i think this function only decodes value according different types. it can't check this detailed schema. maybe it checks these format when invoke "VisitJSON" this function. i don't sure this . this is just my guess !

openapi3filter/req_resp_decoder.go
Luffy110: I have updated this.

openapi3filter/req_resp_decoder.go
Luffy110: Here . i think we shouldn't return error if doesn't match one of . because if API defined optional query part within oneof . but the request URI doesn't have any query part. so in this scenario, we can't deal with it as an error.

fenollp: How does one defines an optional query part within oneOf?

Luffy110: Sorry, I mean if we defined one optional(required is false) query part .and this query parameter has an object schema which included "oneof" , such i have added UT in the validation_test.go,  
`{
							Value: &openapi3.Parameter{
								In:   "query",
								Name: "queryArgOneOf",
								Schema: openapi3.NewOneOfSchema(
									openapi3.NewStringSchema().WithMaxLength(2),
									openapi3.NewDateTimeSchema(),
								).NewRef(),
							},
						},`

if i return error when oneof doesn't matech .such following codes: `if isMatched == 1 {
			return value, nil
		}
		return nil, errors.New("decode Oneof failed")`

then will appear following error , but i think we shouldn't report an error in this case. do you agree ? so i return nil instead of error.
--- FAIL: TestFilter (0.00s)
	/home/shenjie/go/src/github.com/getkin/kin-openapi/openapi3filter/validation_test.go:133: Request: POST http://example.com/api/prefix/v/suffix
	/home/shenjie/go/src/github.com/getkin/kin-openapi/openapi3filter/validation_test.go:188: 
			Error Trace:	validation_test.go:188
			Error:      	Received unexpected error:
			            	Parameter 'queryArgOneOf' in query has an error: decode Oneof failed
			Test:       	TestFilter
FAIL
FAIL	github.com/getkin/kin-openapi/openapi3filter	0.004s

fenollp: How about returning an error when `value == nil` and `param.Required == true`? I agree that if param isn't required that shouldn't be an error.

openapi3filter/req_resp_decoder.go
```diff
@@ -224,20 +224,62 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
...
+			if value != nil {
+				isMatched++
+			}
+		}
+		if isMatched == 1 {
+			return value, nil
+		}
+		return nil, nil
 	}
+	// TODO: if schema.Value.Not isn't nil , how to deal with it?
```
fenollp: As a user I'd prefer the decoder to fail on unhandled `not` rather than continue silently. Let's return an error when `schema.Value.Not` isn't `nil`.

Same for empty `schema.Value.Type`.

openapi3filter/req_resp_decoder.go
Luffy110: OK，good idea ! i will update it .

openapi3filter/req_resp_decoder.go
Luffy110: I think if Not isn't nil, here we still can't return an error. because if Not isn't nil, it still doesn't mean an error i think . so i don't know how to deal with the same as an empty Value.Type?
Not means: according to json schema defination https://tools.ietf.org/html/draft-wright-json-schema-validation-00#section-5.25 

This keyword's value MUST be an object.  This object MUST be a valid
   JSON Schema.

   An instance is valid against this keyword if it fails to validate
   successfully against the schema defined by this keyword.

fenollp: Oh I mean we should error on non-nil Not because if it is set and we don't handle decoding Not then we should return an error, instead of the current behavior: silently continuing.

openapi3filter/req_resp_decoder.go
```diff
@@ -224,20 +224,66 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
...
+				return nil, err
+			}
+			if value != nil {
+				isMatched++
+			}
+		}
+		if isMatched == 1 {
+			return value, nil
+		}
+		if len(schema.Value.Required) == 0 {
```
fenollp: Ah you've mistaken JSON schema's `required` field with openapiv3's params `required` field.
Around https://github.com/getkin/kin-openapi/blob/02f5fafb329ca3c8f21b1f7cf1aeb80f8dd4c656/openapi3filter/req_resp_decoder.go#L222 an additional argument should be passed: `param.Required`.

Luffy110: Yes  you are right, i have  mistaken about this.  and i have added this argu in decodeValue function.

openapi3filter/req_resp_decoder.go
```diff
@@ -208,8 +208,14 @@ func decodeStyledParameter(param *openapi3.Parameter, input *RequestValidationIn
 	var dec valueDecoder
 	switch param.In {
 	case openapi3.ParameterInPath:
+		if input.PathParams == nil {
```
fenollp: Could you check `len` instead of `nil`-ness here?

Luffy110: the PathParams is a map type. so i think we check nil is better.

openapi3filter/req_resp_decoder.go
Luffy110: Yes ,you are right. i have added this logic.

fenollp:
Thanks a lot :)

PR #154, opened 2019-12-13
Fixed missed case for parameter validation.

tuzkov:
From https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#parameterObject

- A parameter MUST contain either a schema property, or a content property, but not both


fenollp:
Thanks!

PR #155, opened 2019-12-16
Resolve references using the relative path of the referenced document

davidlopezre:
References should be resolved using the relative path of the referenced document.

According to the [Reference Object section](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#reference-object) in the OpenAPI 3.0.2 spec, reference resolution is accomplished as defined by the [JSON Reference specification](https://tools.ietf.org/id/draft-pbryan-zyp-json-ref-03.html#RFC3986) which states:

> If the URI contained in the JSON Reference value is a relative URI, then the base URI resolution MUST be calculated according to [RFC3986], section 5.2. Resolution is performed relative to the referring document.

Instead of always using the root document’s directory when resolving refs, this PR correctly tracks the directory of each component as they are resolved.

### DIFF
--- a/openapi3/swagger_loader.go
+++ b/openapi3/swagger_loader.go
@@ -385,7 +385,10 @@ func (swaggerLoader *SwaggerLoader) resolveParameterRef(swagger *Swagger, compon
 		return nil
 	}
 
-	refDocumentPath := referencedDocumentPath(documentPath, ref)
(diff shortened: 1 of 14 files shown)


fenollp:
LGTM. Thanks for this!

PR #157, opened 2019-12-19
Fix #156, various parameter issues

### LINKED ISSUE #156: Various issues with converting v2 -> v3 with parameters
Hi,

I've found a few issues with converting parameters across from v2 to v3:
* $refs don't convert at all
* path and global references don't convert across

(PR incoming)

anz-gordonj7:
* Params with references are converted correctly
* Global and Path params are converted across

### DIFF
--- a/openapi2conv/openapi2_conv.go
+++ b/openapi2conv/openapi2_conv.go
@@ -156,6 +156,11 @@ func ToV3Parameter(parameter *openapi2.Parameter) (*openapi3.ParameterRef, *open
 	if parameter == nil {
 		return nil, nil, nil
 	}
+	if ref := parameter.Ref; len(ref) > 0 {
+		return &openapi3.ParameterRef{
+			Ref: ToV3Ref(ref),
+		}, nil, nil
+	}
 	in := parameter.In
 	if in == "body" {
 		result := &openapi3.RequestBody{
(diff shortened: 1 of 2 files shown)


openapi2conv/openapi2_conv_test.go
```diff
@@ -1,11 +1,10 @@
-package openapi2conv_test
+package openapi2conv
```
fenollp: Why the test package rename?

openapi2conv/openapi2_conv.go
```diff
@@ -364,6 +370,7 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
 		if pathItem == nil {
 			continue
 		}
+		result.AddOperation(path, "GET", nil)
```
fenollp: What does this solve?

openapi2conv/openapi2_conv_test.go
anz-gordonj7: avoids needing to import the real package, no need to put the tests in a different package name

openapi2conv/openapi2_conv.go
anz-gordonj7: Some of the swagger files I was converting have paths defined with params but no operations, im not sure if that is even sensible swagger, but adding this call means that the call to `result.Paths[path].Parameters = params` is guaranteed safe, without having to check that the maps are initialised. The nil "GET" op becomes a no-op anyway

fenollp: This project is meant to support only the spec. Please fix your swagger files so they are valid and revert this.

openapi2conv/openapi2_conv_test.go
fenollp: This makes sure no test code lingers in the actual package. Please revert.

Issue #158, opened 2019-12-23
Receiving incorrect error, when endpoint is path with unsupported method


ErrorBoi:
Hi, I've found that when I send request to a path with unsupported method, error message in response is "Path was not found" instead of ["Path doesn't support the HTTP method"](https://github.com/getkin/kin-openapi/blob/master/openapi3filter/router.go#L199)

Steps to reproduce:
1. Run a server, generated from OpenAPI specification. For example, [Petstore](https://github.com/deepmap/oapi-codegen/blob/master/examples/petstore-expanded/chi/petstore.go#L16) from OpenAPI examples
2. Send request to a path with unsupported method, for example PUT /users

Actual result: 
```
{
    "message": "Path was not found"
}
```
Expected result:
```
{
    "message": "Path doesn't support the HTTP method"
}
```

fenollp:
`pathpattern` takes the method and the path pattern as one whole argument for some reason that I've yet to uncover, and that is the reason why it can't differentiate between "path not found" and "path doesnt support method".

I have started reworking this but feel free to chime in if you feel like it.

ErrorBoi:
Method and path pattern are concatenated [here](https://github.com/getkin/kin-openapi/blob/master/openapi3filter/router.go#L176), maybe it's the reason.

Maybe "Path doesn't support the HTTP method" error should be thrown, if node doesn't match any suffix, but path pattern is listed in swagger spec?

ErrorBoi:
I made a simple fix, please take a look if it's ok @fenollp 
https://github.com/getkin/kin-openapi/compare/master...ErrorBoi:master

ErrorBoi:
After current fix "Path was not found" error is never thrown. "Path doesn't support the HTTP method" error is always thrown instead.
I created Pull Request with fix (tested it on 2 server instances and it works correctly)
https://github.com/getkin/kin-openapi/pull/161

PR #161, opened 2019-12-26
Throw 404 if no pathItem, 405 if pathItem has no operation

ErrorBoi:
Throw 404 error if no pathItem found, throw 405 if pathItem has no operation


ErrorBoi:
This PR is a fix of [this issue](https://github.com/getkin/kin-openapi/issues/158), have no rights to reopen it.

fenollp:
Ah! Sorry about that. Could you modify at least one test in router_test.go to explicitly check for the error message?

PR looks good. I’ll merge once regression test is in.

ErrorBoi:
As far as I see, we might get "Path was not found", "Method not allowed" and "Doesn't match any server", so I added error handling for them in router_test.go.

PR #166, opened 2020-01-23
Also resolve refs in additionalProperties

chirino:


openapi2conv/openapi2_conv.go
```diff
@@ -242,6 +242,9 @@ func ToV3SchemaRef(schema *openapi3.SchemaRef) *openapi3.SchemaRef {
 	for k, v := range schema.Value.Properties {
 		schema.Value.Properties[k] = ToV3SchemaRef(v)
 	}
+	if schema.Value.AdditionalProperties != nil {
```
fenollp: `AdditionalProperties` can also be a boolean.

chirino: What do you mean? It's defined as `AdditionalProperties *SchemaRef`
Do you mean AdditionalPropertiesAllowed??

chirino: What change are you requesting?

fenollp: Oh you're right I forgot that this is handled with multijson tags.

PR #170, opened 2020-02-03
Add `uuid` case so schema validation doesn't fail

disposedtrolley:
This PR adds an empty case for `uuid` so schema validation doesn't fail when processing OAS files containing a `format: uuid` attribute.


fenollp:
Working as intended.
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#data-types
> Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification.

Please use https://github.com/getkin/kin-openapi/pull/170/files#diff-741cd2de3f96fb896e215807cd8cca12R522
See https://godoc.org/github.com/getkin/kin-openapi/openapi3#DefineStringFormat

PR #173, opened 2020-02-12
Change type of scopes in openapi2 to map[string]string

mbilski:
According to the spec https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#scopesObject the scopes are map, not a list.

```
{
  "write:pets": "modify pets in your account",
  "read:pets": "read your pets"
}
```


fenollp:
Thanks for the fix. Could you add a test that for example reads an openapiv2 doc and ensures the scopes are parsed properly?

mbilski:
Sure, I added a test for reading example petstore 2.0 swagger

fenollp:
Thanks!

Issue #174, opened 2020-02-12
Validation does not allow endpoints with two forward slashes


FrancisLennon17:
I'm using this library to validate openapi definitions and I'm getting errors relating to the endpoints having two forward slashes in the endpoints. 

For example `/test/api//asdf/` as a path would fail due to the validation [here](https://github.com/getkin/kin-openapi/blob/75b6118a7234f5751e51831a23718abaaf059a7e/openapi3/paths.go#L22-L24)

Could this check be removed? 

Although two forward slashes are not recommended, [it is valid](https://stackoverflow.com/questions/20523318/is-a-url-with-in-the-path-section-valid) to use, and therefore should not fail. 

Various other online tools that validate openapi definitions allow for endpoints with double forward slashes in endpoints
- https://editor.swagger.io/
- https://mermade.github.io/openapi-gui/
- https://apitools.dev/swagger-parser/online/

fenollp:
Fixed by https://github.com/getkin/kin-openapi/pull/175

PR #175, opened 2020-02-14
removing validation for double slashes

FrancisLennon17:
https://github.com/getkin/kin-openapi/issues/174


fenollp:
Indeed. Thanks!

PR #177, opened 2020-02-17
Use `net/http` method constants

zchee:
Use the `net/http` method constants to the operation methods.

Currently, hardcoded by a naive string type. But each method already provided from the `net/http` package.
Also, `Operations` are certainly http(s) access operations. I think using them is not wrong.


fenollp:
Thanks :)

PR #184, opened 2020-03-04
Encode security requirements correctly

Jleagle:


fenollp:
Thanks for the catch! Could you add a simple test that can serve as a regressor?

Issue #188, opened 2020-03-15
`linkRef.Value.Headers` has type `Schema`


pseudo-su:
I have recently written some code that walks the `openapi3.Swagger` object, traversing all objects that can contain nested refs.

I have noticed that for some reason linkRef.Value.Headers is a `map[string]*Schema` instead of a `map[string]*SchemaRef` or `map[string]*HeaderRef`.

I think this is a bug.

The original PR where I ran into this: https://github.com/deepmap/oapi-codegen/pull/154

pseudo-su:
Actually after investigating this a little bit I'm even more confused, I can't find any reference to `links` having `headers` in the OpenAPI spec https://swagger.io/specification/#linkObject

fenollp:
Indeed this struct seems way out of spec https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#linkObject
Missing fields:
* operationRef
* requestBody
* server
Out-of-spec fields:
* href
* headers

If you can turn your PR around in order add the missing fields & remove the out of spec ones that'd be great :)

pseudo-su:
awesome, I'll have a look tomorrow

PR #189, opened 2020-03-15
fix: make link.Headers use HeaderRef

pseudo-su:
Change the type of `Link.Value.Headers` from `map[string]*Schema` to `map[string]*HeaderRef`

Fixes: https://github.com/getkin/kin-openapi/issues/188

### DIFF
--- a/openapi3/link.go
+++ b/openapi3/link.go
@@ -2,6 +2,7 @@ package openapi3
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/getkin/kin-openapi/jsoninfo"
 )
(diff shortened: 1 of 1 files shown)


pseudo-su:
@fenollp I've updated my PR to at least remove Href, Headers and add Server.

As far as what needs to be done in order to properly support `operationRef` and `requestBody` I'm quite uncertain.

- `operationRef` doesn't seem to be like a normal ref, and I get the impression that creating a `*OperationRef` wouldn't be appropriate because I assume those are for locations where you can either have inline objects or `{ "$ref": "#/components/schemas/thething"}`. Not exactly sure if I should just make it a `OperationRef string`.
- `requestBody` isn't/shouldn't be a responseBodyRef because it's actually an `expression` according to the OpenAPI spec but I don't really know what that means. `parameters` uses an empty interface for it's `Any | {expression}` component so should I also make requestBody `interface{}`?

fenollp:
This fixes your issue and brings this lib closer to supporting the spec. Merging as is.
Thanks :)

Issue #190, opened 2020-03-16
patternProperties validation is not aligned with JSON-schema


peter-plochan:
Even the OpenAPI 3 doesn't support pattern properties in schemas yet, the validator includes validation of `patternProperties` field within the schema. I would expect that these `patternProperties` should be the same as defined in [JSON-schema](https://json-schema.org/understanding-json-schema/reference/object.html#pattern-properties).
The problem is that the validator accepts something like this (see [implementation in schema.go](https://github.com/getkin/kin-openapi/blob/9ae9769e5f4e33fe78ee828bc9587f670847d6de/openapi3/schema.go#L1100-L1117)):
```
Pet:
  type: object
  patternProperties: '^somePattern$'
  additionalProperties:
    $ref: '#/components/schemas/PatternPropertySchema'
```

fenollp:
Indeed patternProperties field support should be deleted. I'll review any PR you send my way :)

peter-plochan:
@fenollp https://github.com/getkin/kin-openapi/pull/192 - not sure if it's enough for complete removal, but hopefully yes :)

PR #192, opened 2020-03-17
Remove patternProperties field support

peter-plochan:
Fixes https://github.com/getkin/kin-openapi/issues/190

The patternProperties field is not supported by OpenAPI 3 Specification as well as it is not compliant with the newest JSON-schema spec.


fenollp:
Thanks!

Issue #195, opened 2020-03-27
Would you accept an `http.Handler` implementation?


codyaray:
This is really what I was looking for in openapi-validation. This makes for an extremely easy integration path with existing applications. This is already how some other services like 	OpenCensus integrate as well (see `ochttp` in the sample `main` below).

This would invoke the request and response validators automatically. It would require a `ValidationErrorEncoder` and I think we can provide a [good default implementation](https://github.com/getkin/kin-openapi/issues/194) for that as well.

Here's an example of how an integration may look:
```
h, err := app.NewHTTPHandler(...)
check("unable to create handler", err)

// wrap with openapi validation handler
h = &openapi3filter.ValidationHandler{
  Handler: h,
  SwaggerFile: "docs/openapi.yaml",
  ErrorEncoder: openapi3filter.ValidationsErrorEncoder{Encoder: myAppEncoder}
}

// wrap by ochttp handler
h = &ochttp.Handler{Handler: h}

// start server
s := &gohttp.Server{Addr: *httpAddr, Handler: h}
ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
defer s.Shutdown(ctx)
defer cancel()

level.Info(l).Log("msg", fmt.Sprintf("Starting http server at addr: %s", *httpAddr))
errc <- s.ListenAndServe()
```

fenollp:
An `http.Handler` implementation makes total sense.
I'd also accept an implementation of https://golang.org/pkg/net/http/#RoundTripper

codyaray:
Addressed in #197

Issue #194, opened 2020-03-27
Would you accept a custom error encoder representation?


codyaray:
There's a non-trivial amount of effort to transform the returned errors into something we can encode and send to the user. I'm writing a custom encoder right now for our own purposes, but wanted to know if this is something you would consider accepting upstream to share with other users in the community?

Since the exact encoding is application dependent, we write it into an intermediate "error object" that's easier to just encode into whatever media type and structure the app uses. This is an easier jumping off point for other developers.

Below I've included samples of the tests from my work. This intermediate error object can be seen in the `wantErrResponse` below. They also include some actual TODOs in my code where I think we could improve the upstream to return better errors generally.

```
//
// Basics
//

{
	name: "error - unknown path",
	args: validationArgs{
		r: badPath,
	},
	wantErrReason:   "Path was not found",
	wantErrResponse: &corev1.Error{Code: http.StatusNotFound, Message: "Path was not found"},
},

{
	name: "error - unknown method",
	args: validationArgs{
		r: badMethod,
	},
	wantErrReason:   "Path doesn't support the HTTP method",
	// TODO: By HTTP spec, this should have an Allow header with what is allowed
	// but kin-openapi doesn't provide us the requested method or path, so impossible to provide details
	wantErrResponse: &corev1.Error{Code: http.StatusMethodNotAllowed, Message: "Path doesn't support the HTTP method"},
},

<snip>

//
// Content-Type
//

{
	name: "error - unknown content-type on POST",
	args: validationArgs{
		r: unknownContentType,
	},
	wantErrReason:      "failed to decode request body",
	wantErrParseKind:   openapi3filter.KindUnsupportedFormat,
	wantErrParseReason: "unsupported content type \"application/xml\"",
	wantErrResponse:    &corev1.Error{Code: http.StatusUnsupportedMediaType, Message: "unsupported content type \"application/xml\""},
},
{
	name: "error - unsupported content-type on POST",
	args: validationArgs{
		r: unsupportedContentType,
	},
	wantErrReason:   "header 'Content-Type' has unexpected value: \"text/plain\"",
	wantErrResponse: &corev1.Error{Code: http.StatusUnsupportedMediaType, Message: "unsupported content type \"text/plain\""},
},

<snip>

//
// Query strings
//

{
	name: "error - invalid enum value for query string parameter",
	args: validationArgs{
		r: newPetstoreRequest(t, http.MethodGet, "/pet/findByStatus?status=sold&status=watdis", nil),
	},
[...]

PR #197, opened 2020-03-28
Add validation handler and error encoder

codyaray:
Implement https://github.com/getkin/kin-openapi/issues/194 and https://github.com/getkin/kin-openapi/issues/195

This is what I had to add on top of kin-openapi in order to provide openapi validation errors in my API service. More discussion at https://github.com/getkin/kin-openapi/issues/194#issuecomment-605488953

### DIFF
--- a/openapi3filter/validation_handler.go
+++ b/openapi3filter/validation_handler.go
@@ -1,0 +1,78 @@
+package openapi3filter
+
+import (
+	"context"
+	"net/http"
+)
+
+type AuthenticationFunc func(context.Context, *AuthenticationInput) error
+
+func NoopAuthenticationFunc(context.Context, *AuthenticationInput) error { return nil }
+
+var _ AuthenticationFunc = NoopAuthenticationFunc
+
+type ValidationHandler struct {
+	Handler            http.Handler
+	AuthenticationFunc AuthenticationFunc
+	SwaggerFile        string
+	ErrorEncoder       ErrorEncoder
+	router             *Router
+}
+
+func (h *ValidationHandler) Load() error {
+	h.router = NewRouter()
+
+	err := h.router.AddSwaggerFromFile(h.SwaggerFile)
+	if err != nil {
+		return err
+	}
+
+	// set defaults
+	if h.Handler == nil {
+		h.Handler = http.DefaultServeMux
+	}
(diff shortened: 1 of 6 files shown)


openapi3filter/validation_handler.go
```diff
@@ -0,0 +1,78 @@
...
+	return nil
+}
+
+func (h *ValidationHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	err := h.validateRequest(r)
+	if err != nil {
+		h.ErrorEncoder(r.Context(), err, w)
+		return
+	}
+	// TODO: validateResponse
```
fenollp, line 50: Indeed

openapi3filter/validation_kit.go
```diff
@@ -0,0 +1,85 @@
+package openapi3filter
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+///////////////////////////////////////////////////////////////////////////////////
+// We didn't want to tie kin-openapi too tightly with go-kit.
```
fenollp, line 10: Let's not copy https://github.com/go-kit/kit/blob/266ff8dc37c693d0649707e519c93c1f85868bdc/transport/http/server.go here but rather import this package.

fenollp: On second thought this brings a pretty large dependency tree so disregard the above comment.

openapi3filter/validation_handler.go
fenollp: This can come later as another PR.

fenollp:
Could you look at the WIP new router at https://github.com/getkin/kin-openapi/pull/210?
I believe I could add a `Handle` or `HandleFunc` that could help you there.

wolfeidau:
This looks pretty close, anything i can do to help?

codyaray:
@fenollp thanks for merging. I agree that I really don't like matching on error strings. That was the only option given what we have at the moment though.

Sorry I missed your follow-up questions. What did you mean by "producing correct JSON pointers" ? I assume you mean this kinda stuff? https://github.com/getkin/kin-openapi/pull/197/files#diff-ee2d579818e5124734fb3ce2cd7e87acR155

PR #200, opened 2020-04-02
fix panic when multipart form contains extra parts not present in schema

Ventrosky:
Validating a request body with content 'multipart/form-data' that contained parts not defined in the schema caused Panic, 
fixed by ignoring the undefined properties

### DIFF
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -923,13 +923,21 @@ func multipartBodyDecoder(body io.Reader, header http.Header, schema *openapi3.S
 		subEncFn := func(string) *openapi3.Encoding { return enc }
 		// If the property's schema has type "array" it is means that the form contains a few parts with the same name.
 		// Every such part has a type that is defined by an items schema in the property's schema.
-		valueSchema, exists := schema.Value.Properties[name]
+		var valueSchema *openapi3.SchemaRef
+		var exists bool
+		valueSchema, exists = schema.Value.Properties[name]
 		if !exists {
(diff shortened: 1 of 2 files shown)


fenollp:
Nice catch. Why should these be silently ignored? Shouldn't an error be returned here?

Ventrosky:
@fenollp well if somebody is migrating API version and have some app that still has a form which  uses extra form fields we might not want to break everything and let them handle extra values, that's what I thought at least, returning an error is a good option too

fenollp:
Is it not possible to also define the old form schema and wrap the two in a `oneOf`?
Or maybe un-schema'd data should be ignored as long as `additionalProperties: false` is not in the schema? otherwise an error should be returned. 

I'd like kin-openapi to be as strict as can be :)
I think there should be a check on additionalProperties. Could that work for your use case?

Ventrosky:
Yes, you have a point. Sure works for me :)

Ventrosky:
I have added an extra commit trying to implement the proposed solution, thanks for the quick response btw :)

Ventrosky:
Additional properties that are defined in the schema, for example
```
additionalProperties:
  type: object
  properties:
    code:
      type: integer
```
 those might have to be included in the decoded body form and only ignore only those undefined
```
additionalProperties: true
```

fenollp:
LGTM. Could you add yet another (thanks!) that explicitly checks for error on `additionalProperties: false`?
Thank you

Ventrosky:
Sure thing, done, thank you so much for your work and responsiveness!

fenollp:
Likewise!

Issue #146, opened 2019-11-26
DataType Validation


FrancisLennon17:
Using the openapi3 library to validate a definition example found under the OpenAPI-Specification GitHub found [here](https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v2.0/json/uber.json).

(Im using this library to convert from swagger 2 to openapi 3, then validating it)

This library returns an `unsupportedFormat` error for the below parameter as uuid is not a supported format for the string type.

![image](https://user-images.githubusercontent.com/28549164/69651305-c9b54f00-1067-11ea-81de-78ea37a5bcfa.png)

[Here](https://github.com/getkin/kin-openapi/blob/1116f3ecaa880095aa0e8f4eb64a6dc04c8a288e/openapi3/schema.go#L488-L525) is where the error is being thrown. A comment says that those specified are the formats listed there are the formats supported by OpenAPIv3, but after reading through the OpenAPI-Specification [docs](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md#format) it says:
> Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification

I have a number of undefined formats in some openapi3 and swagger2 definitions I'm working with, and the validate function will not pass due to this. Could this part of the validate function be removed as it seems these formats can be used?

Arms-Li:
you can register your own format and pattern by call

```go
openapi3.DefineStringFormat(name string, pattern string)
```

benfdking:
We are encountering the same issue. Whereas the validator against an enitity works:

```go
case "string":
		if format := schema.Format; len(format) > 0 {
			switch format {
			// Supported by OpenAPIv3.0.1:
			case "byte", "binary", "date", "date-time", "password":
				// In JSON Draft-07 (not validated yet though):
			case "regex":
			case "time", "email", "idn-email":
			case "hostname", "idn-hostname", "ipv4", "ipv6":
			case "uri", "uri-reference", "iri", "iri-reference", "uri-template":
			case "json-pointer", "relative-json-pointer":
			default:
				return unsupportedFormat(format)
			}
		}
```

The validation of the schema itself does not:

```go 
case "string":
		if format := schema.Format; len(format) > 0 {
			switch format {
			// Supported by OpenAPIv3.0.1:
			case "byte", "binary", "date", "date-time", "password":
				// In JSON Draft-07 (not validated yet though):
			case "regex":
			case "time", "email", "idn-email":
[...]

PR #202, opened 2020-04-17
Disable Format Validation Bool

FrancisLennon17:
https://github.com/getkin/kin-openapi/issues/146#issuecomment-614729892

Adding a variable that allows the disabling of schema type format validation

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -19,6 +19,9 @@ var (
 	// SchemaErrorDetailsDisabled disables printing of details about schema errors.
 	SchemaErrorDetailsDisabled = false
 
+	//SchemaFormatValidationDisabled disables validation of schema type formats.
+	SchemaFormatValidationDisabled = false
+
 	errSchema = errors.New("Input does not match the schema")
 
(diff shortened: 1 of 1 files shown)


FrancisLennon17:
@fenollp Is this ok?

fenollp:
Shouldn't this be allowing any formats not specified by the openapi spec?
The ones specified should not be disabled by this flag no?

FrancisLennon17:
@fenollp I'm not sure I understand what the difference between allowing formats not specified by the openapi spec, and disabling the format check altogether?
There is no set list of "unsupported formats", people can have custom formats, which is why I set the variable to disable the format check. Which is what I thought you had suggested in your comment

> @FrancisLennon17 I'm open to a (probably package wide var) setting that disables checking format but it'd have to be off by default.

fenollp:
I'm asking you to move that `	if !SchemaFormatValidationDisabled {` check inside the `default:` clause. Can you do that?

FrancisLennon17:
@fenollp like so?

fenollp:
Precisely. Thanks :)

FrancisLennon17:
Nice, than you!

Issue #185, opened 2020-03-05
omitempty in structs producing invalid definitions


FrancisLennon17:
There are omitempty tags that end up removing a required field when sending the struct. The `description` field is required as per the Open API 3 specification, and therefore is not a valid definition. 

Could either 
- The omit empty tags be removed from any fields required by the Open API 3 specification
- Or validation added not to allow any required fields, like description, to have an empty string (similar to what has been done for info.Title/info.Version)

This behaviour worked as expected in previous versions up until [this](https://github.com/getkin/kin-openapi/pull/148/files#diff-4da5065edd98a754c28386877c1f2598R28) change was made

```
info:
  title: 'asdf'
  version: '1'
openapi: 3.0.0
paths:
  /test:
    post:
      responses:
        '200':
          description: ''
  /test2:
    post:
      parameters: 
        - name: ''
          in: 'query'
          schema: {}
          required: false
        - name: '1'
          in: 'query'
          content: 
            application/json: {}
          required: false
        - name: '2'
          in: 'query'
          content: 
            application/json: 
              schema: {}
          required: false
      requestBody:
        content: {}
      responses:
        '200':
          description: ''
```

fenollp:
It is unclear whether `description: ''` is to be considered different than not mentioning the field at all IMO.
While I am sorry some of your specs were broken by me merging #148 I believe that second option (forbidding empty required fields) should be the way to go.

FrancisLennon17:
A blank string is valid against the Open API specification from what I've seen in various online validation tools. Enforcing non-empty strings for fields seems counterproductive. Its impractical for developers to ensure that each and every response/parameter/reference description string is not empty in a definition. And I can't imagine that it is common for each and every one to be filled in.

fenollp:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#response-object
> REQUIRED. A short description of the response. CommonMark syntax MAY be used for rich text representation.

IMO various validation tools are disregarding the part about `A short description of the response.` for convenience when a spec is produced by a human. This lib however strives to implement the spec.

bryanbecker:
My two cents (some team as Francis):

We've looked at a huge number of OpenAPI docs, and something closing in on 100% of them have at least on operation with a missing description.  This could be attributed to the fact that generally API docs are forever a work-in-progress.

[...]

PR #204, opened 2020-04-21
Remove omitempty from response description

FrancisLennon17:
https://github.com/getkin/kin-openapi/issues/185

If the response description field was set empty string and passed into the LoadSwaggerFromData function, the field would be removed when sent due to the omitempty tag. This results in a definition that is invalid. 

Removing the omitempty tag on response description so the field is not dropped with a definition like below

```
{
	"openapi": "3.0.0",
	"info": {
		"version": "1.0.0",
		"title": "Sample API",
		"description": "A sample API to illustrate OpenAPI concepts"
	},

	"paths": {
		"/path1": {
			"get": {
        "responses": {
          "200": {
             "description": ""
          }
       }
    }
  }
 }
}
```

image

I looked for other fields that might have similar issues but I could not find any

### DIFF
--- a/openapi3/response.go
+++ b/openapi3/response.go
@@ -34,7 +34,7 @@ func (responses Responses) Validate(c context.Context) error {
 // Response is specified by OpenAPI/Swagger 3.0 standard.
 type Response struct {
 	ExtensionProps
-	Description string                `json:"description,omitempty" yaml:"description,omitempty"`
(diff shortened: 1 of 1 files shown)


fenollp:
There's still the issue of `""` being the empty value for the library itself.
I'll merge this and work on a PR that I'll run by you.

PR #205, opened 2020-04-21
Amend #204 to fix a edge case

### LINKED ISSUE #185: omitempty in structs producing invalid definitions

fenollp:
Fixes #185
Amends #204 such that `decode(encode(spec))` does not set empty response descriptions.


fenollp:
@FrancisLennon17 @bryanbecker does this suit you?
Sorry for the noise, the actual changes are at https://github.com/getkin/kin-openapi/pull/205/commits/ec984fa1d561c3cc81965d1ad00362d668784695#diff-fbcdbb7d909690df2b35770fcbc3ed19L37

FrancisLennon17:
Yes this works for us also. Thanks @fenollp

Issue #118, opened 2019-09-27
Request validation fails on localhost


bjmc:
Hello, I'm not sure if this is a bug strictly-speaking, but we found it to be unexpected behavior.

Say you have a basic spec like this:

```yaml
openapi: "3.0.0"
info:
  title: Simple API Example
  version: 0.1.0
servers:
- url: "https://example.com/api/"
  description: Deployed absolute URL
- url: "http://localhost:8080/"
  description: Running locally
paths:
  /:
    get:
      summary: Application health-check.
      responses:
        200:
          description: Application name and version.
          content:
            application/json:
              schema:
                properties:
                  data:
                    type: object
                    properties:
                      name:
                        type: string
                        example: "Sample App"
                      version:
                        type: string
                        example: "0.1.0"

```

And then a simple web app that includes validation using `kin-openapi/openapi3filter`:

```go
package main

import (
	"context"
	"flag"
	"fmt"
	"io"
	"log"
	"net/http"
	"path/filepath"

	"github.com/getkin/kin-openapi/openapi3filter"
)

func main() {
	// OpenAPIv3 request validation:
	specPath, _ := filepath.Abs("./spec.yaml")
	oa3router := openapi3filter.NewRouter().WithSwaggerFromFile(specPath)

	validationHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		ctx := context.TODO()
		log.Printf("request.URL is %s", r.URL)
		route, pathParams, _ := oa3router.FindRoute(r.Method, r.URL)
		// Validate request
		requestValidationInput := &openapi3filter.RequestValidationInput{
			Request:    r,
			PathParams: pathParams,
			Route:      route,
		}
		if err := openapi3filter.ValidateRequest(ctx, requestValidationInput); err != nil {
			w.WriteHeader(400)
			io.WriteString(w, err.Error()+"\n")
		} else {
			fmt.Fprint(w, "{\"version\": \"0.1.0\", \"name\": \"Sample App\"}\n")
		}
	})

	httpAddr := flag.String("http.addr", ":8080", "HTTP listen address")
	log.Printf("HTTP server starting on %v\n", *httpAddr)
	log.Fatal(http.ListenAndServe(*httpAddr, validationHandler))
}
```

If you run `$ go run main.go` and make a request using `$ curl localhost:8080` you'll find you get an `invalid route` response.

The logging output is:

```
2019/09/27 14:24:17 HTTP server starting on :8080
[...]

PR #210, opened 2020-04-25
Rework router

### LINKED ISSUE #129: Paths ending with an extension aren't matched
When using the router, the following definition is matched properly (the corresponding route is found):

```
paths:
  '/books/{id}':
    parameters:
      - schema:
          type: integer
        name: id
        in: path
        required: true
    get:
      # ...
```

But this doesn't:

```
paths:
    parameters:
      - schema:
          type: integer
        name: id
        in: path
        required: true
  '/books/{id}.json':
    get:
      # ...
```

fenollp:
Fixes #129 

TODO:
* [x] use gorilla/mux
* [x] support Servers (note: complex Server URLs & variables)
* [x] cleanup previous router (remove pathpattern too)
* [ ] maybe add `(*Router) Handle(handler http.Handler)`
* [ ] maybe add `(*Router) HandleFunc(func(http.ResponseWriter, *http.Request))`
* [x] look into https://github.com/getkin/kin-openapi/issues/118
* [x] look into https://github.com/getkin/kin-openapi/pull/149


fenollp:
Note to self: look into https://github.com/julienschmidt/httprouter over gorillamux. Maybe provide routers for both (and gin?) in dedicated packages (to not add on dependencies of users that do not use this router)

PR #218, opened 2020-06-05
Add IgnoreServers option to validation handler

codyaray:
This is useful for bundling an OpenAPI spec with your service and promoting through your pipelne. For example, your OpenAPI doc probably just wants to list your prod servers; but you want to compile it in to your code/docker image and promote between environments. So this docker image needs to run in devel, stag, and prod... each of which have different URLs. The easiest way is to just not validate the http request host matches the OpenAPI server definition at all. That's what this does... a long with a bit of path rewriting.


fenollp:
I appreciate you sharing your work :)
Is it possible to replace the Servers during container startup?
* read the dev/staging/prod host from env
* set `doc.Server` accordingly
* continue with initialization
Then no need to disable host validation.

codyaray:
@fenollp yeah that's possible, just requires more infra to set the `env` correctly in the deployment.

does it need to be a blocker for this PR?

fenollp:
Yeah I am cautious of changes that disable some validation. Also since there's a way to achieve this that doesn't require disabling Hosts validation while being compliant with the 12FactorsApp principles I'll be closing this PR.

PR #225, opened 2020-06-12
Swagger2 Extensions

FrancisLennon17:
Added extensions, marshalling/unmarshalling and conversion to and from V3 for:
- Root swagger
- Path 
- Operation
- Parameter (& v3 Requestbody)
- Responses
- SecurityDefinitions

### DIFF
--- a/openapi2/openapi2.go
+++ b/openapi2/openapi2.go
@@ -11,10 +11,12 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/getkin/kin-openapi/jsoninfo"
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
 type Swagger struct {
+	openapi3.ExtensionProps
 	Info                openapi3.Info                  `json:"info"`
 	ExternalDocs        *openapi3.ExternalDocs         `json:"externalDocs,omitempty"`
 	Schemes             []string                       `json:"schemes,omitempty"`
@@ -27,6 +29,14 @@ type Swagger struct {
 	SecurityDefinitions map[string]*SecurityScheme     `json:"securityDefinitions,omitempty"`
 	Security            SecurityRequirements           `json:"security,omitempty"`
 	Tags                openapi3.Tags                  `json:"tags,omitempty"`
+}
+
+func (swagger *Swagger) MarshalJSON() ([]byte, error) {
+	return jsoninfo.MarshalStrictStruct(swagger)
+}
+
+func (swagger *Swagger) UnmarshalJSON(data []byte) error {
+	return jsoninfo.UnmarshalStrictStruct(data, swagger)
 }
 
 func (swagger *Swagger) AddOperation(path string, method string, operation *Operation) {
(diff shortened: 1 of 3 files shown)


openapi2/openapi2.go
```diff
@@ -179,14 +291,58 @@ type Header struct {
...
+	StripNonCustomExtensions(&s.Extensions)
+	*securityScheme = SecurityScheme(s)
+	return nil
+}
+
+func (s SecurityScheme) MarshalJSON() ([]byte, error) {
+	return marshallFromMap(s.Extensions, reflect.ValueOf(s), reflect.TypeOf(s))
+}
+
+func StripNonCustomExtensions(extensions *map[string]interface{}) {
```
fenollp, line 325: I don't think you need to use a pointer to a map as map are already references, no?

openapi2/openapi2.go
```diff
@@ -43,16 +65,49 @@ func (swagger *Swagger) AddOperation(path string, method string, operation *Oper
...
+	if err := json.Unmarshal([]byte(b), &p.Extensions); err != nil {
+		return err
+	}
+	StripNonCustomExtensions(&p.Extensions)
+	*pathItem = PathItem(p)
+	return nil
+}
+
+func (p PathItem) MarshalJSON() ([]byte, error) {
+	return marshallFromMap(p.Extensions, reflect.ValueOf(p), reflect.TypeOf(p))
```
fenollp, line 110: Why not use the jsoninfo package that's widely used in openapi3 package?

FrancisLennon17: Honestly I was a little intimidated by it at first, but I think I've a better grasp of it now. I'll make that change today 👍

FrancisLennon17:
@fenollp Could you take another look. jsoninfo package made that a lot easier 🤦

openapi2/openapi2.go
```diff
@@ -11,10 +11,13 @@ import (
 	"fmt"
 	"net/http"
 
+	"github.com/getkin/kin-openapi/jsoninfo"
+
```
fenollp, line 15: Can you remove the newline here?

openapi2conv/openapi2_conv_test.go
```diff
@@ -22,7 +22,7 @@ func TestConvOpenAPIV3ToV2(t *testing.T) {
 }
 
 func TestConvOpenAPIV2ToV3(t *testing.T) {
-	var swagger2 openapi2.Swagger
+	swagger2 := openapi2.Swagger{}
```
fenollp, line 25: Could you revert this? I'm not seeing what it enables.

openapi2/openapi2.go
```diff
@@ -43,7 +54,22 @@ func (swagger *Swagger) AddOperation(path string, method string, operation *Oper
 	pathItem.SetOperation(method, operation)
 }
 
+func (swagger *Swagger) AddPathExtensions(path string, extensionProps openapi3.ExtensionProps) {
```
fenollp, line 57: It looks like this is only called from the conversion package and could very well be implemented there.
How do you see this function being useful to other callers? Can we make this a private func within `openapi2_conv`?

FrancisLennon17:
@fenollp Made the changes you requested there.

FrancisLennon17:
Thank you @fenollp 🙏

fenollp:
Thank you :)

PR #233, opened 2020-07-10
conversion: conversion for file upload from v2 to v3 and vice versa

gitforbit:
This change enables converting spec for file upload with POST from v2 to v3 and vice versa
according to https://swagger.io/docs/specification/2-0/file-upload/ specification.
Added support for `multipart/form-data`.
Added `in: formData` case for file upload.
Updated unit tests related sections.

### DIFF
--- a/openapi2conv/openapi2_conv.go
+++ b/openapi2conv/openapi2_conv.go
@@ -121,6 +121,51 @@ func ToV3PathItem(swagger *openapi2.Swagger, pathItem *openapi2.PathItem) (*open
 	return result, nil
 }
 
+func ToV3RequestBodyFormData(parameters []*openapi2.Parameter) *openapi3.RequestBodyRef {
+	if len(parameters) == 0 || parameters[0].In != "formData" {
+		return nil
+	}
+	schema := &openapi3.Schema{
+		Type:       "object",
+		Properties: make(map[string]*openapi3.SchemaRef, len(parameters)),
+	}
+	for _, parameter := range parameters {
+		if parameter.In != "formData" || parameter.Name == "" {
+			continue
+		}
+		format := parameter.Format
+		typ := parameter.Type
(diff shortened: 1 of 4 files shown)


gitforbit:
Hi @fenollp,

Can you review this please?

fenollp:
Great work! Thanks!

gitforbit:
Thank you for quick review :)

openapi2conv/openapi2_conv.go
```diff
@@ -121,6 +121,51 @@ func ToV3PathItem(swagger *openapi2.Swagger, pathItem *openapi2.PathItem) (*open
 	return result, nil
 }
 
+func ToV3RequestBodyFormData(parameters []*openapi2.Parameter) *openapi3.RequestBodyRef {
+	if len(parameters) == 0 || parameters[0].In != "formData" {
```
philLITERALLY, line 125: Just a quick question, this logic implies that any `formData` parameters will only be converted if the very first parameter is a `formData`. This means that if the `formData` parameter is second in the list it'll not be converted?

Also, the logic below implies that if there was any `formData` parameters that any other parameters will be ignored? (https://github.com/getkin/kin-openapi/pull/233/files#diff-29b4895fc45bb3c5db76a7a9635a745eR186-R200)

fenollp: Indeed it looks like the spec does not enforce this logic. We need a PR that processes the other parameters as well.

gitforbit: I am working on that. I totally omited the fact that other parameters are possible too, since https://swagger.io/docs/specification/2-0/file-upload/ examples not use any and my use-case didn't need any of them.

fenollp: Addressed in https://github.com/getkin/kin-openapi/pull/239

PR #237, opened 2020-07-16
Fix wrong `Found unresolved ref` error when converting from Swagger/OpenAPIv2 spec

### LINKED ISSUE #187: Getting `Found unresolved ref` errors during parse swagger 2.0 contract
I just tried to test this project with our swagger 2.0 json file. During implementation, i couldn't find a life-saver v2 to v3 converter func or auto converter something like: 

```
For 3.0: _ = openapi3filter.NewRouter().WithSwaggerV3FromFile("swagger.json")
For 2.0: _ = openapi3filter.NewRouter().WithSwaggerV2FromFile("swagger.json")
```

I found [this](https://github.com/getkin/kin-openapi/issues/113#issuecomment-549157194) simple implementation:

```
func V2ToV3(v2Data []byte) (*openapi3.Swagger, error) {
	v2Spec := &openapi2.Swagger{}
	if err := json.Unmarshal(v2Data, v2Spec); err != nil {
		return nil, err
	}
	v3Spec, err := openapi2conv.ToV3Swagger(v2Spec)
	if err != nil {
		return nil, err
	}
	return v3Spec, nil
}
```

Anyway, the main question is that actually why i am getting `Found unresolved ref errors`. `openapi3filter` couldn't recognize our `$ref` model i think. Simplified `json` is here:

```
{
  "swagger": "2.0",
  "info": {
    "description": "Test Golang Application",
    "version": "1.0",
    "title": "Test",
    "contact": {
      "name": "Test",
      "email": "test@test.com"
    }
  },
  "host": "",
  "basePath": "/test",
  "definitions": {
    "model.ProductSearchAttributeRequest": {
      "type": "object",
      "properties": {
        "filterField": {
          "type": "string"
        },
        "filterKey": {
          "type": "string"
        },
        "type": {
          "type": "string"
        },
        "values": {
          "$ref": "#/definitions/model.ProductSearchAttributeValueRequest"
        }
      },
      "title": "model.ProductSearchAttributeRequest"
    },
    "model.ProductSearchAttributeValueRequest": {
      "type": "object",
      "properties": {
        "imageUrl": {
          "type": "string"
        },
        "text": {
          "type": "string"
        }
      },
      "title": "model.ProductSearchAttributeValueRequest"
    }
  }
}
```

`main` func:

```
func main() {
	input, err := ioutil.ReadFile("min.json")
	if err != nil {
		panic(err)
	}
	oapi3, err := V2ToV3(input)
	if err != nil {
		panic(err)
	}
	_ = openapi3filter.NewRouter().WithSwagger(oapi3)
}
```

Output:

`panic: Validating Swagger failed: Found unresolved ref: '#/definitions/model.ProductSearchAttributeValueRequest'`

So, `#/definitions/model.ProductSearchAttributeValueRequest` is defined under `definitions` as we see. That simplified json is **auto-generated** file from [Swag](https://github.com/swaggo/swag). I don't know why it throws an error, i think it shouldn't. Should it?

fenollp:
Fixes #187

### DIFF
--- a/openapi2conv/issue187_test.go
+++ b/openapi2conv/issue187_test.go
@@ -1,26 +1,26 @@
 package openapi2conv
 
 import (
-  "context"
-  "encoding/json"
-  "testing"
+	"context"
+	"encoding/json"
+	"testing"
 
-  "github.com/getkin/kin-openapi/openapi2"
-  "github.com/getkin/kin-openapi/openapi3"
-  "github.com/stretchr/testify/require"
+	"github.com/getkin/kin-openapi/openapi2"
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/stretchr/testify/require"
 )
 
 func v2v3(spec2 []byte) (doc3 *openapi3.Swagger, err error) {
-  var doc2 openapi2.Swagger
-  if err = json.Unmarshal(spec2, &doc2); err != nil {
-    return
-  }
-  doc3, err = ToV3Swagger(&doc2)
-  return
+	var doc2 openapi2.Swagger
+	if err = json.Unmarshal(spec2, &doc2); err != nil {
+		return
+	}
+	doc3, err = ToV3Swagger(&doc2)
+	return
 }
 
 func TestIssue187(t *testing.T) {
-  spec := `
+	spec := `
 {
   "swagger": "2.0",
   "info": {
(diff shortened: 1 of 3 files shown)


philLITERALLY:
Hey @fenollp, I've been testing this locally and noticed an issue, could you verify what should occur in the definition below? It seems like this update has broken the `allOf` ref value but maybe that's expected behaviour?

```
swagger: '2.0'
info:
  version: 1.0.0
  title: title
paths:
  '/test':
    get:
      parameters:
        - $ref: '#/definitions/TestRef'
      responses:
        '200':
          description: description
definitions:
  TestRef:
    type: object
    allOf:
      - $ref: '#/definitions/TestRef2'
  TestRef2:
    type: object
```

PR #238, opened 2020-07-17
conversion: fileUpload: handle parameters other than formData

gitforbit:
This change enables using all possible parameters for file upload.
Previously, only `in: formData` was handled. Now, other parameters
are suported e.g.: `in: path`.
If at least one parameter has `in: formData`, it is considered to
be handled for file upload.


fenollp:
Handled in https://github.com/getkin/kin-openapi/commit/8459893af4ad6c5efe9a42f13d685dfe52ceb91f

fenollp:
Please open an issue if this broke some things for you.

gitforbit:
It is OK, thanks.

PR #248, opened 2020-09-02
Fix openapi3.referencedDocumentPath

kshlm:
`openapi3.referencedDocument` tried to `path.Join` two URLs, and then
generate a new `url.URL` by parsing the obtained path. The URL
obtained at the end would be invalid, when the base path is HTTP URL.

Given base path as 'http://example.com/schemas' and ref as '/schema1',
the obtained end URL would be 'http:/example.com/schemas'. This happens
because `path.Join` cleans the path it generates and replaces repeated
'//' with a single '/'.

This is fixed by doing a `path.Join` on the base URL.Path and the ref,
and setting this result as the URL.Path of the new URL.

### DIFF
--- a/openapi3/swagger_loader_referenced_document_path_test.go
+++ b/openapi3/swagger_loader_referenced_document_path_test.go
@@ -9,13 +9,11 @@ import (
 
 func TestReferencedDocumentPath(t *testing.T) {
 	httpURL, err := url.Parse("http://example.com/path/to/schemas/test1.yaml")
-	if err != nil {
-		panic(err)
-	}
+	require.NoError(t, err)
+
 	fileURL, err := url.Parse("path/to/schemas/test1.yaml")
-	if err != nil {
-		panic(err)
-	}
+	require.NoError(t, err)
+
 	refEmpty := ""
 	refNoComponent := "moreschemas/test2.yaml"
 	refWithComponent := "moreschemas/test2.yaml#/components/schemas/someobject"
@@ -56,8 +54,7 @@ func TestReferencedDocumentPath(t *testing.T) {
 		},
 	} {
 		result, err := referencedDocumentPath(test.path, test.ref)
-		require.NotNil(t, result)
-		require.Nil(t, err)
+		require.NoError(t, err)
 		require.Equal(t, test.expected, result.String())
 	}
 }
--- a/openapi3/swagger_loader_referenced_document_path_test.go
+++ b/openapi3/swagger_loader_referenced_document_path_test.go
@@ -1,0 +1,63 @@
+package openapi3
+
+import (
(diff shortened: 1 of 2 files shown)


fenollp:
Hi there, thanks for the patch!
Could you add a few tests on this function? Especially one that would not pass before your fix.
Thanks

kshlm:
I've added the tests.

I'm not a 100% sure on the about the expected result of `referencedDocumentPath`. I would have expected it to be `path.Dir(basePath)+refPath`. But other tests seem to expect it to be `path.Dir(basePath)+path.Dir(refPath)+"/"`. Which is what I've implemented the changes and the tests for.

openapi3/swagger_loader_referenced_document_path_test.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestReferencedDocumentPath(t *testing.T) {
+	httpURL, err := url.Parse("http://example.com/path/to/schemas/test1.yaml")
+	if err != nil {
```
fenollp, line 12: Please call `require.NoError(t, err)` instead.

openapi3/swagger_loader_referenced_document_path_test.go
```diff
@@ -0,0 +1,63 @@
...
+		panic(err)
+	}
+	fileURL, err := url.Parse("path/to/schemas/test1.yaml")
```
fenollp, line 15: Please call `require.NoError(t, err)` instead.

openapi3/swagger_loader_referenced_document_path_test.go
```diff
@@ -0,0 +1,63 @@
...
+			expected: "path/to/schemas/moreschemas/",
+		},
+		{
+			path:     fileURL,
+			ref:      refWithComponent,
+			expected: "path/to/schemas/moreschemas/",
+		},
+	} {
+		result, err := referencedDocumentPath(test.path, test.ref)
+		require.NotNil(t, result)
```
fenollp, line 59: Remove this line.

openapi3/swagger_loader_referenced_document_path_test.go
```diff
@@ -0,0 +1,63 @@
...
+		require.Nil(t, err)
```
fenollp, line 60: Please call `require.NoError(t, err)` instead.

fenollp (approved):
Thanks for the tests :) I have a few nitpicks but after that it'll be merged.

kshlm:
Not sure why tests are failing. They're unrelated to my change in the `openapi3` package.

Issue #171, opened 2020-02-04
Incorect validation of nullable in allOf


pali:
When `allOf` contains two structures and both of them have defined `"nullable" : true` then validator should accept `null` value (as `null` value is valid for both defined structures).

But kin-openapi validator currently reject `null` value. Below is schema and test case.

```json
{
    "openapi": "3.0.0",
    "info": {
        "version": "1.0.0",
        "title": "test"
    },
    "paths": {
        "/": {
            "get": {
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "object",
                                    "properties": {
                                        "value": {
                                            "$ref": "#/components/schemas/NULL"
                                        }
                                    }
                                }
                            }
                        }
                    }
                }
            }
        }
    },
    "components" : {
        "schemas" : {
            "NULL" : {
                "allOf" : [
                    {
                        "nullable" : true,
                        "type" : "boolean",
                        "enum" : [ false ]
                    },
                    {
                        "nullable" : true,
                        "type" : "boolean",
                        "enum" : [ true ]
                    }
                ]
            }
        }
    }
}
```

Test input: `{ "value" : null }`. Validator should accept this input, but currently it rejects it.

Looks like that PR https://github.com/getkin/kin-openapi/pull/169 did not fixed this case.

arnaubennassar:
Same problem here:

I have this object that is used as a response:

```yaml
    MyObject:
      type: object
      description: MyObject description
      properties:
        nullableRef:
          allOf:
            - $ref: '#/components/schemas/NotNullableObject'
            - nullable: true
            - description: nullableRef description
            - example: 5432
          required:
            - nullableRef
          additionalProperties: false
```

And when validating I'm geting the error: `response body doesn't match the schema: Error at "nullableRef":Value is not nullable`

m-mohr:
The example by @pali conflicts with https://github.com/OAI/OpenAPI-Specification/blob/master/proposals/003_Clarify-Nullable.md

nullable: true doesn't add "null" to the values of the enum and thus is not allowed. The allOf needs to look like this:
```
                "allOf" : [
                    {
                        "type" : "boolean",
                        "enum" : [ false ]
                    },
                    {
                        "type" : "boolean",
                        "enum" : [ true ]
                    },
                    {
                        "enum" : [ null ]
                    }
                ]
```

or simply:
```
                    {
                        "type" : "boolean",
                        "nullable": true
                    }
```

[...]

PR #253, opened 2020-09-06
reproduce incorrect allOf + nullable behaviour

fenollp:
cc https://github.com/getkin/kin-openapi/issues/171

### DIFF
--- a/openapi3/schema_test.go
+++ b/openapi3/schema_test.go
@@ -205,6 +205,32 @@ var schemaExamples = []schemaExample{
 	},
 
 	{
+		Title: "NULLABLE ALLOF",
+		Schema: NewAllOfSchema(
+			NewBoolSchema().WithNullable(),
+			NewBoolSchema().WithNullable(),
+		),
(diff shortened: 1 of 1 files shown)


openapi3/schema_test.go
```diff
@@ -161,6 +161,43 @@ var schemaExamples = []schemaExample{
...
+			false,
+		},
+		AllInvalid: []interface{}{
+			2,
+			4.2,
+			[]interface{}{42},
+			"bla",
+			map[string]interface{}{},
+		},
+	},
```
ori-shalom, line 199: This test will always fail due to the schema being an object and expected valid values are not.

The following schema:

```json
{
	"type": "object",
	"properties": {
		"value": {
			"allOf": [
				  { "type": "bool", "nullable": true },
				  { "type": "bool", "nullable": true }
			]
		}
	}
}
```

Is valid with the following JSON:

```
{ "value": null }
{ "value": true }
{ "value": false }
```

But not with the following:

```
null
true
false
```

PR #262, opened 2020-10-26
Prevent a panic in the error encoder

richard-rance:
The Parameter can be nil when the enum is set of one of the body fields.


fenollp:
Thanks for catching this & going all the way to a patch.

Could you provide a test that would fail if it weren't for your PR, please?

richard-rance:
I've added couple of tests and fixed another error I noticed.

openapi3filter/validation_error_encoder.go
```diff
@@ -148,7 +148,7 @@ func convertSchemaError(e *RequestError, innerErr *openapi3.SchemaError) *Valida
 	}
 
 	// Add error source
-	if e.Parameter != nil && e.Parameter.In == "query" {
```
fenollp, line 151: Why is this needed?

richard-rance: This was the other issue I found. There were no tests for parameters in other locations. When I added a test for one in the header the source was not getting set.

PR #263, opened 2020-10-28
Consumes request bodies

FrancisLennon17:
v2Tov3 - Takes the root level consumes, or the operation level consumes array, and generates a requestbody for each entry in the array

v3Tov2 - Adds each requestbody contentType from the list to the consumes array, and adds either formData Parameters or a `in:body` parameter (V3toV2 is not backward compatible as v3 can have multiple request bodies for an operation)


openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+func consumesToArray(consumes map[string]bool) []string {
+	var consumesArr []string
+	for key, _ := range consumes {
+		consumesArr = append(consumesArr, key)
+	}
+	sort.Strings(consumesArr)
+	return consumesArr
+}
+
+func FromV3RequestBodies(name string, requestBodyRef *openapi3.RequestBodyRef, components *openapi3.Components) (*openapi2.Parameters, *openapi2.Parameters, map[string]bool, error) {
```
fenollp, line 653:
```suggestion
func FromV3RequestBodies(name string, requestBodyRef *openapi3.RequestBodyRef, components *openapi3.Components) (*openapi2.Parameters, *openapi2.Parameters, map[string]struct{}, error) {
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	formParameters := openapi2.Parameters{}
+	bodyOrRefParameters := openapi2.Parameters{}
+	if ref := requestBodyRef.Ref; ref != "" {
+		bodyOrRefParameters = append(bodyOrRefParameters, &openapi2.Parameter{Ref: FromV3Ref(ref)})
+		return &bodyOrRefParameters, &formParameters, nil, nil
+	}
+
+	//Only select one formData or request body for an individual requesstBody as swagger 2 does not support multiples
```
fenollp, line 661: How about returning an error "both form data and request body were provided"

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	consumes := make(map[string]bool)
```
fenollp, line 662:
```suggestion
	var consumes map[string]struct{}
```
Let's not allocate if requestBodyRef.Value == nil.

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	if requestBodyRef.Value != nil {
+		for contentType, mediaType := range requestBodyRef.Value.Content {
+			//add each type to the consumes array
```
fenollp, line 665:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+			consumes[contentType] = true
```
fenollp, line 666:
```suggestion
			consumes[contentType] = struct{}{}
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+			formParams := FromV3RequestBodyFormData(mediaType)
+			if len(formParams) != 0 {
```
fenollp, line 668:
```suggestion
			if formParams := FromV3RequestBodyFormData(mediaType); len(formParams) != 0 {
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+		return &bodyOrRefParameters, &formParameters, nil, nil
+	}
+
+	//Only select one formData or request body for an individual requesstBody as swagger 2 does not support multiples
+	consumes := make(map[string]bool)
+	if requestBodyRef.Value != nil {
+		for contentType, mediaType := range requestBodyRef.Value.Content {
+			//add each type to the consumes array
+			consumes[contentType] = true
+			formParams := FromV3RequestBodyFormData(mediaType)
```
fenollp, line 667:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+				formParameters = formParams
+			} else {
+				paramName := name
+				if originalName, ok := requestBodyRef.Value.Extensions["x-originalParamName"]; ok {
+					json.Unmarshal(originalName.(json.RawMessage), &paramName)
+				}
+
+				r, err := FromV3RequestBody(paramName, requestBodyRef, mediaType, components)
+				if err != nil {
+					return &bodyOrRefParameters, &formParameters, consumes, err
```
fenollp, line 678: Again `formParameters` may be empty here. How about returning named values? e.g.:
```go
func myfunc(name string, requestBodyRef *openapi3.RequestBodyRef, components *openapi3.Components) (
	bodyOrRefParameters *openapi2.Parameters,
	formParameters *openapi2.Parameters,
	consumes map[string]struct{},
	err error,
) {
```
this way you can just 
```go
if err != nil {
	return
}
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,49 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+				}
+				bodyOrRefParameters = append(bodyOrRefParameters, r)
+			}
+		}
+	}
+	return &bodyOrRefParameters, &formParameters, consumes, nil
```
fenollp, line 684: I'm not able to see a need for returning pointers here. These types represent slices so this seems backwards to me.

Feel free to edit https://github.com/getkin/kin-openapi/blob/fc3f90d8183b71442d5730b8e6cff877c9b93e45/openapi2/openapi2.go#L168 if it helps (please try), into:
```go
type Parameters = []*Parameter
```
NOTE: I'm not sure why it's `[]*Parameter` instead of `[]Parameter` but there are other types like this and it's surely a matter for another PR anyway.

openapi2conv/openapi2_conv.go
```diff
@@ -711,8 +757,13 @@ func FromV3SchemaRef(schema *openapi3.SchemaRef, components *openapi3.Components
 	if v := schema.Value.Items; v != nil {
 		schema.Value.Items, _ = FromV3SchemaRef(v, components)
 	}
-	for k, v := range schema.Value.Properties {
-		schema.Value.Properties[k], _ = FromV3SchemaRef(v, components)
+	keys := make([]string, 0, len(schema.Value.Properties))
+	for k, _ := range schema.Value.Properties {
```
fenollp, line 761:
```suggestion
	for k := range schema.Value.Properties {
```

openapi2conv/openapi2_conv.go
```diff
@@ -848,27 +895,30 @@ func FromV3Operation(swagger *openapi3.Swagger, operation *openapi3.Operation) (
...
+		bodyOrRefParameters, formDataParameters, consumes, err := FromV3RequestBodies(name, v, &swagger.Components)
+		if err != nil {
+			return nil, err
+		}
+		if len(*formDataParameters) != 0 {
+			result.Parameters = append(result.Parameters, *formDataParameters...)
+		} else if len(*bodyOrRefParameters) != 0 {
+			//add a single request body
+			for i, param := range *bodyOrRefParameters {
+				if i == 0 {
```
fenollp, line 913:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -848,27 +895,30 @@ func FromV3Operation(swagger *openapi3.Swagger, operation *openapi3.Operation) (
...
+					result.Parameters = append(result.Parameters, param)
```
fenollp, line 914:
```suggestion
				result.Parameters = append(result.Parameters, param)
```

openapi2conv/openapi2_conv.go
```diff
@@ -848,27 +895,30 @@ func FromV3Operation(swagger *openapi3.Swagger, operation *openapi3.Operation) (
...
+				}
```
fenollp, line 915:
```suggestion
				break
```

openapi2conv/openapi2_conv.go
```diff
@@ -848,27 +895,30 @@ func FromV3Operation(swagger *openapi3.Swagger, operation *openapi3.Operation) (
...
+
+		bodyOrRefParameters, formDataParameters, consumes, err := FromV3RequestBodies(name, v, &swagger.Components)
+		if err != nil {
+			return nil, err
+		}
+		if len(*formDataParameters) != 0 {
+			result.Parameters = append(result.Parameters, *formDataParameters...)
+		} else if len(*bodyOrRefParameters) != 0 {
+			//add a single request body
+			for i, param := range *bodyOrRefParameters {
```
fenollp, line 912:
```suggestion
			for _, param := range *bodyOrRefParameters {
```

openapi2conv/openapi2_conv.go
```diff
@@ -848,27 +895,30 @@ func FromV3Operation(swagger *openapi3.Swagger, operation *openapi3.Operation) (
...
+			return nil, errors.New("could not find a name for request body")
+		}
+
+		bodyOrRefParameters, formDataParameters, consumes, err := FromV3RequestBodies(name, v, &swagger.Components)
+		if err != nil {
+			return nil, err
+		}
+		if len(*formDataParameters) != 0 {
+			result.Parameters = append(result.Parameters, *formDataParameters...)
+		} else if len(*bodyOrRefParameters) != 0 {
```
fenollp, line 910: Again, I feel there should be an error when there are both formDataParameters and bodyOrRefParameters. Thoughts?

Maybe a system a registerable callbacks (so the user can tune behavior in this case and when picking which body to use)  can be added in later, when someone requires it.

openapi3/content.go
```diff
@@ -12,6 +12,32 @@ func NewContent() Content {
 	return make(map[string]*MediaType, 4)
 }
 
+func NewContentWithSchema(schema *Schema, consumes []string) Content {
```
fenollp, line 15:
```suggestion
func NewContentWithSchema(schema *Schema, consumes ...string) Content {
```

openapi3/content.go
```diff
@@ -12,6 +12,32 @@ func NewContent() Content {
...
+		}
+	}
+	content := make(map[string]*MediaType, len(consumes))
+	for _, mediaType := range consumes {
+		content[mediaType] = NewMediaType().WithSchema(schema)
+	}
+	return content
+}
+
+func NewContentWithSchemaRef(schema *SchemaRef, consumes []string) Content {
```
fenollp, line 28:
```suggestion
func NewContentWithSchemaRef(schema *SchemaRef, consumes ...string) Content {
```

fenollp (changes requested):

openapi2conv/openapi2_conv.go
```diff
@@ -601,25 +606,23 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
-				}
+		if len(*formDataParameters) != 0 {
+			for _, param := range *formDataParameters {
+				result.Parameters[param.Name] = param
 			}
-
-			r, err := FromV3RequestBody(swagger, paramName, requestBody)
-			if err != nil {
-				return nil, err
+		} else if len(*bodyOrRefParameters) != 0 {
```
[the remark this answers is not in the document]
FrancisLennon17: It is valid in V3. Do we want to do our best to convert it, or only convert if it is perfectly backward compatible?

openapi2conv/openapi2_conv.go
FrancisLennon17: again, is the priority only to convert when it is perfectly backward compatible?

openapi2conv/openapi2_conv.go
FrancisLennon17: I'm not sure I follow, it never allocates if  ref.Value == nil.  It would have returned an empty map. I use a map[string]bool just for the ease of accessing the key in the KVP

openapi2conv/openapi2_conv.go
FrancisLennon17: sure

openapi2conv/openapi2_conv.go
FrancisLennon17: Sure I'll change it. I don't need it for my own use case

openapi3/content.go
FrancisLennon17: Why is this needed, will this work?

fenollp: The idea is to not have to pass a `consumes` such that `len(consumes) == 0` as a second argument to trigger that empty consumes argument behaviour.

i.e. this way `NewContentWithSchema(schema, []string{}) == NewContentWithSchema(schema)`.

openapi2conv/openapi2_conv.go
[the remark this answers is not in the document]
fenollp: I'm down to do our best to convert it, definitely. I guess my question is more: is this conversion lossless? If not: should we return a non-nil error or should we silently discard some V3 spec items?

openapi2conv/openapi2_conv.go
fenollp: You're right: let's do best effort.

openapi2conv/openapi2_conv.go
fenollp: Doesn't `make(...)` perform allocation? It is called before the `if`, regardless.

openapi2conv/openapi2_conv.go
fenollp: I mean lets do best effort, as discussed a bit above. If one needs more then we can always plug in callbacks later.

fenollp (commented):
Alright, thanks for your time. There shouldn't be much changes left now :)

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+func consumesToArray(consumes map[string]bool) []string {
+	consumesArr := make([]string, 0, len(consumes))
+	for key := range consumes {
+		consumesArr = append(consumesArr, key)
+	}
+	sort.Strings(consumesArr)
+	return consumesArr
+}
+
+func fromV3RequestBodies(name string, requestBodyRef *openapi3.RequestBodyRef, components *openapi3.Components) (*openapi2.Parameters, *openapi2.Parameters, map[string]bool, error) {
```
fenollp, line 653:
```suggestion
func fromV3RequestBodies(name string, requestBodyRef *openapi3.RequestBodyRef, components *openapi3.Components) (
	bodyOrRefParameters openapi2.Parameters,
	formParameters openapi2.Parameters,
	consumes map[string]struct{},
	err error,
) {
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	var formParameters openapi2.Parameters
```
fenollp, line 654:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	var bodyOrRefParameters openapi2.Parameters
```
fenollp, line 655:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	if ref := requestBodyRef.Ref; ref != "" {
+		bodyOrRefParameters = append(bodyOrRefParameters, &openapi2.Parameter{Ref: FromV3Ref(ref)})
+		return &bodyOrRefParameters, &formParameters, nil, nil
```
fenollp, line 658:
```suggestion
		return
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+	}
+
+	//Only select one formData or request body for an individual requesstBody as swagger 2 does not support multiples
+	var consumes map[string]struct{}
```
fenollp, line 662:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+			consumes[contentType] = struct{}{}
+			if formParams := FromV3RequestBodyFormData(mediaType); len(formParams) != 0 {
+				formParameters = formParams
+			} else {
+				paramName := name
+				if originalName, ok := requestBodyRef.Value.Extensions["x-originalParamName"]; ok {
+					json.Unmarshal(originalName.(json.RawMessage), &paramName)
+				}
+
+				r, err := FromV3RequestBody(paramName, requestBodyRef, mediaType, components)
```
fenollp, line 674:
```suggestion
				var r *openapi2.Parameter
				if r, err = FromV3RequestBody(paramName, requestBodyRef, mediaType, components); err != nil {
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+				if err != nil {
```
fenollp, line 675:
```suggestion
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+					return &bodyOrRefParameters, &formParameters, consumes, err
```
fenollp, line 676:
```suggestion
					return
```

openapi2conv/openapi2_conv.go
```diff
@@ -638,6 +641,47 @@ func FromV3Swagger(swagger *openapi3.Swagger) (*openapi2.Swagger, error) {
...
+				}
+				bodyOrRefParameters = append(bodyOrRefParameters, r)
+			}
+		}
+	}
+	return &bodyOrRefParameters, &formParameters, consumes, nil
```
fenollp, line 682:
```suggestion
	return
```

fenollp (changes requested):

fenollp (approved):

FrancisLennon17:
@fenollp my apologies, I had started a new job and hadn't gotten around to making the changes you requested. Thanks so much for getting that merged!

Issue #264, opened 2020-11-08
path validation panic when pathItem is nil


schmilmo:
Hi,

If I try to validate the following schema the Paths.Validate function panics because it's accessing pathItem == nil
I used the following schema
```
openapi: 3.0.1
info:
  title: test schema
  description: test schema
  termsOfService: https://www.company.com
  contact:
    name: info
    email: info@company.com
  license:
    name: open-licence
    url: https://www.company.com/open-licence
  version: v3.1.6
servers:
- url: www.company.com
paths:
  /endpoint1/:
```

fenollp:
I doubt this is a valid OpenAPI spec but anyway I agree this call shouldn't panic. 

Please open a quick PR with a nil check and a test that relies on your example. Thanks.

fenollp:
Fixed by https://github.com/getkin/kin-openapi/pull/266

PR #266, opened 2020-11-12
fixed panic in path validation (issue #264)

schmilmo:


openapi3/paths.go
```diff
@@ -22,30 +22,32 @@ func (paths Paths) Validate(c context.Context) error {
...
-			if parameterRef != nil {
-				if parameter := parameterRef.Value; parameter != nil && parameter.In == ParameterInPath {
-					globalCount++
-				}
-			}
-		}
-		for method, operation := range pathItem.Operations() {
-			var count uint
-			for _, parameterRef := range operation.Parameters {
+		if pathItem != nil {
```
fenollp, line 25: More like 
```go
if pathItem == nil {
    paths[path] = &PathItem{}
    pathItem = paths[path]
}

fenollp:
Note that tests are failing due to non deterministic behavior of some other tests so don't worry about that.

schmilmo:
> Note that tests are failing due to non deterministic behavior of some other tests so don't worry about that.

I noticed

Issue #270, opened 2020-11-18
OpenAPI v3 validation fails over identifier regex


heyvister:
Hi,

I'm trying to convert existing customer's v2 to v3. The conversion has succeeded but when I validate it afterwards it fails over invalid regex:
_"identifier "DeltaMapper[Infant]" is not supported by OpenAPIv3 standard (regexp: "^[a-zA-Z0-9._-]+$")_"

Is there an option to expose _const identifierPattern = `^[a-zA-Z0-9._-]+$`_ @ components.go ?

Tnx

fenollp:
Indeed `ValidateIdentifier` doesn't seem to be backed by any explicit wording in https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md
Git blame returns a 4 years old commit with no description by the original author of this lib.

I'd say we can safely remove this whole function. I'd rather we have per-case validation instead:
* header names can't contain whitespace nor colons
* query parameters probably shouldn't contain `=?&[]#` chars (but then again, maybe they can if they get urlencoded?)
* path parameters must not contain slashes (ditto wrt encoding)

A PR replacing `ValidateIdentifier` usage with these constraints (thus relaxing the current constraints) would be great.

fenollp:
Alright I got my identifiers mixed up: this is about the names of components. It turns out [the spec does mention this](https://github.com/OAI/OpenAPI-Specification/blob/f8449d1a893cc6a811c6f3d87e88b05761dc5397/versions/3.1.0.md)
> All the fixed fields declared above are objects that MUST use keys that match the regular expression: ^[a-zA-Z0-9\.\-_]+$.

If handling not quite OpenAPI spec is alright with you we can work out a way for you to override this. Do you need this tweakable
* at the package variable level, or
* per specification?

The first one is the quickest to get working, in which case I'd suggest exporting a `ComponentsIdentifiersRegexp` that one can override through `regexp.MustCompile`.

heyvister:
Hi Pierre,

Thanks for the clarification regarding OpenAPI specification for components keys regular expression.
Indeed I would rather it to be tweakable, I would go with the quickest option of exporting _ComponentsIdentifiersRegexp_

PR #271, opened 2020-11-19
fixed allow square brackets in reference identifier (issue #270)

schmilmo:


openapi3/components.go
```diff
@@ -92,7 +92,7 @@ func (components *Components) Validate(c context.Context) (err error) {
 	return
 }
 
-const identifierPattern = `^[a-zA-Z0-9._-]+$`
+const identifierPattern = `^[\w\[\]._-]+$`
```
fenollp, line 95: Hi please see https://github.com/getkin/kin-openapi/issues/270#issuecomment-730451234

openapi3/components_test.go
```diff
@@ -0,0 +1,37 @@
...
+func TestValidateIdentifier(t *testing.T) {
+	tt := []struct {
+		name       string
+		identifier string
+		isValid    bool
+		errMsg     string
+	}{
+		{name: "valid", identifier: "User-12_3.4", isValid: true, errMsg: ""},
+		{name: "valid with []", identifier: "User[Admin]", isValid: true, errMsg: ""},
+		{name: "invalid", identifier: "User Admin", isValid: false, errMsg: ""},
```
fenollp, line 17: Looks like errMsg is never not empty.
Instead of all this, how about just:
```go
	for identifier, isValid := range map[string]bool{
		"User-12_3.4": true,
		"User[Admin]": true,
		"User Admin": false,
	}{

```

fenollp (changes requested):

PR #287, opened 2021-01-18
Add OIDC Schema format as per spec

krotscheck:
https://swagger.io/docs/specification/authentication/openid-connect-discovery/

### DIFF
--- a/openapi3/security_scheme.go
+++ b/openapi3/security_scheme.go
@@ -123,7 +123,7 @@ func (ss *SecurityScheme) Validate(c context.Context) error {
 		hasFlow = true
 	case "openIdConnect":
 		if ss.OpenIdConnectUrl != "" {
-			return fmt.Errorf("No OIDC URL found for openIdConnect security scheme %s", ss.Name)
+			return fmt.Errorf("No OIDC URL found for openIdConnect security scheme %q", ss.Name)
 		}
 	default:
 		return fmt.Errorf("Security scheme 'type' can't be '%v'", ss.Type)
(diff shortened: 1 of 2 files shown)


openapi3/security_scheme.go
```diff
@@ -114,7 +122,9 @@ func (ss *SecurityScheme) Validate(c context.Context) error {
 	case "oauth2":
 		hasFlow = true
 	case "openIdConnect":
-		return fmt.Errorf("Support for security schemes with type '%v' has not been implemented", ss.Type)
+		if len(ss.OpenIdConnectUrl) == 0 {
```
fenollp, line 125:
```suggestion
		if ss.OpenIdConnectUrl != "" {
```

openapi3/security_scheme.go
```diff
@@ -114,7 +122,9 @@ func (ss *SecurityScheme) Validate(c context.Context) error {
...
+			return fmt.Errorf("No OIDC URL found for openIdConnect security scheme %s", ss.Name)
```
fenollp, line 126:
```suggestion
			return fmt.Errorf("No OIDC URL found for openIdConnect security scheme %q", ss.Name)
```

openapi3/security_scheme.go
```diff
@@ -114,7 +122,9 @@ func (ss *SecurityScheme) Validate(c context.Context) error {
 	case "oauth2":
 		hasFlow = true
 	case "openIdConnect":
-		return fmt.Errorf("Support for security schemes with type '%v' has not been implemented", ss.Type)
+		if ss.OpenIdConnectUrl != "" {
```
fenollp, line 125:
```suggestion
		if ss.OpenIdConnectUrl == "" {
```

PR #292, opened 2021-01-29
Return a more specific error when more than oneOf schemas match

### LINKED ISSUE #289: oneof schema error improvment
**Context:** 
currently, the library when validating an input against a schema and when an error occurred it returns schema error contains scheme filed where the error happens or input is invalid, under some circumstance, this is still not a piece of useful information to exactly know what happened behind the scene. 

**Pain:** 
Inappropriate error message returned when input matches more than "oneof" scheme.   

**Goal:** 
return a different error to distinguish input match more than oneof schemas (conflict) from input does not match oneof schemas. 

**Got:** 
`Error at "/address": Doesn't match schema "oneOf"`

**Expected:**
`Error at "/address": Conflict input match all oneof schemas"`

**Reproduce:** 
```go
package main

import (
	"fmt"

)

func main() {
	data := map[string]interface{}{
		"name":      "kin-openapi",
		"address": "127.0.0.1",
	}
	s, err := openapi3.NewSwaggerLoader().LoadSwaggerFromData([]byte(api))
	if err != nil {
		panic(err)
	}
	err = s.Components.Schemas["Server"].Value.VisitJSON(data)
	fmt.Println(err)

	//
}

var api = 
`components: 
  schemas: 
    Server: 
      properties: 
        address: 
          oneOf: 
            - $ref: "#/components/schemas/ip-address"
            - $ref: "#/components/schemas/domain-name"
        name: 
          type: string
      type: object
    domain-name: 
      maxLength: 10
      minLength: 5
      pattern: "((([a-zA-Z0-9_]([a-zA-Z0-9\\-_]){0,61})?[a-zA-Z0-9]\\.)*([a-zA-Z0-9_]([a-zA-Z0-9\\-_]){0,61})?[a-zA-Z0-9]\\.?)|\\."
      type: string
    ip-address: 
      pattern: "^(([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])\\.){3}([0-9]|[1-9][0-9]|1[0-9][0-9]|2[0-4][0-9]|25[0-5])$"
      type: string
openapi: "3.0.1"
`
```

**Note:** 
* Releated to https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L850. 
* Oneof in the above example added for only to reproduce otherwise should be anyof.

**Suggested Solution:**

```go 
var ErrOneOfConflict = errors.new("input match all oneof schemas")  

//  https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L850.
		if ok != 1 {
			if settings.failfast {
				return errSchema
			}
			 err := &SchemaError{
				Value:       value,
				Schema:      schema,
				SchemaField: "oneOf",
			}
                        if ok > 1 {
                           err.Origin = ErrOneOfConflict
                         }
                        return err 
		}
```

fenollp:
Fixes #289

cc @shaj13

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -25,6 +25,9 @@ var (
 
 	errSchema = errors.New("Input does not match the schema")
 
+	// ErrOneOfConflict is the SchemaError Origin when data matches more than one oneOf schema
+	ErrOneOfConflict = errors.New("input matches more than one oneOf schemas")
+
 	// ErrSchemaInputNaN may be returned when validating a number
 	ErrSchemaInputNaN = errors.New("NaN is not allowed")
 	// ErrSchemaInputInf may be returned when validating a number
@@ -851,11 +854,15 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 			if settings.failfast {
 				return errSchema
 			}
-			return &SchemaError{
+			e := &SchemaError{
 				Value:       value,
 				Schema:      schema,
 				SchemaField: "oneOf",
 			}
+			if ok > 1 {
+				e.Origin = ErrOneOfConflict
+			}
+			return e
 		}
 	}
 
--- a/openapi3/schema_issue289_test.go
+++ b/openapi3/schema_issue289_test.go
@@ -1,13 +1,13 @@
 package openapi3
 
 import (
-  "testing"
+	"testing"
 
-  "github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


fenollp:
@shaj13 please take a look :)

shaj13:
@fenollp LGTM

fenollp:
:+1: Thanks!

PR #295, opened 2021-02-02
fix bug on indice to compare

chemidy:


fenollp:
Thank you very much :)

PR #298, opened 2021-02-04
fix: resolving paths for embedded schema fails if relative doc is not also an OpenAPI spec file

rjelierse:
The OpenAPI spec states that you can embed any schema from a relative document: https://swagger.io/specification/#reference-object, but this currently only works if the relative document is formatted as an OpenAPI spec file.


rjelierse:
Will revisit in the future

Issue #80, opened 2019-03-29
Request validation does not work correctly with discriminator 


evgenykireev:
Discriminator property is not taken into account when validating request body.
Given example schema  from https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/
request body
```json
{
  "pet_type": "Cat",
  "breed":    "Dingo",
  "bark":     true,
}
```
should not be valid, however it passes ValidateRequest.

Code to reproduce:
```go
package main

import (
	"bytes"
	"context"
	"encoding/json"
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
)

const testSchema = `openapi: 3.0.0
info:
  title: ''
  version: 0.0.1
paths:
  /:
    post:
      requestBody:
        required: true
        content:
          application/json:
            schema:
              anyOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Dog'
              discriminator:
                propertyName: pet_type

components:
  schemas:
    Pet:
      type: object
      required:
        - pet_type
      properties:
        pet_type:
          type: string
      discriminator:
        propertyName: pet_type

    Dog:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            breed:
              type: string
              enum: [Dingo, Husky, Retriever, Shepherd]
    Cat:
      allOf:
        - $ref: '#/components/schemas/Pet'
        - type: object
          properties:
            hunts:
              type: boolean
            age:
              type: integer
`

func main() {
	payload := map[string]interface{}{
		"pet_type": "Cat",
		"breed":    "Dingo",
		"bark":     true,
	}

	p, _ := json.Marshal(payload)

	swagger, _ := openapi3.NewSwaggerLoader().LoadSwaggerFromData([]byte(testSchema))

	router := openapi3filter.NewRouter().WithSwagger(swagger)
	req, _ := http.NewRequest(http.MethodPost, "/", bytes.NewReader(p))
	route, pathParams, _ := router.FindRoute(req.Method, req.URL)
	req.Header.Set("Content-Type", "application/json")

	requestValidationInput := &openapi3filter.RequestValidationInput{
		Request:    req,
		PathParams: pathParams,
		Route:      route,
	}

	err := openapi3filter.ValidateRequest(context.TODO(), requestValidationInput)
	if err == nil {
		fmt.Println("Valid")
	} else {
[...]

PR #321, opened 2021-03-05
Adds oneOf/discriminator/mapping management

RiccardoManfrin:
Ciao, the change managed to make me validate/parse this spec
```
openapi: 3.0.0
info:
  version: 0.2.0
  title: yaAPI

paths:

  /config:
    put:
      operationId: ConfigSet
      requestBody:
        required: true
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/config'
      responses:
        '200':
          description: Ok

components:
  schemas:
    config:
      oneOf:
        - $ref: '#/components/schemas/simplified'
        - $ref: '#/components/schemas/full'
      discriminator:
        propertyName: conf_type
        mapping:
          full: '#/components/schemas/full'
          simplified: '#/components/schemas/simplified'
    generic_config:
      type: object
      required:
        - conf_type
      properties:
        conf_type:
          type: string
          enum:
            - simplified
            - full
      discriminator:
        propertyName: conf_type
        mapping:
          full: '#/components/schemas/full'
          simplified: '#/components/schemas/simplified'
    full:
      allOf:
      - $ref: '#/components/schemas/generic_config'
      - type: object
        description: Full configuration
        properties:
          base64:
            type: string
      
    simplified:
      allOf:
      - $ref: '#/components/schemas/generic_config'
      - type: object
        description: Simplified (aided) minimal config
        properties:
          output:
            type: object
```
with this call

    curl -X PUT "http://127.0.0.1:8667/api/1/mgmt/config" -H  "accept: */*" -H  "Content-Type: application/json" -d "{\"conf_type\":\"full\",\"base64\":\"stuff\"}"

Before the change the validation failed cause the code hit "ok" two times here

https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L850

I'm positive this is breaking other cases but maybe you know how to rework it. Besides..maybe this has to do with

https://github.com/getkin/kin-openapi/issues/80

Let me know if it's rubbish or if it could be used with some polishing

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -847,7 +847,19 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 			err := v.visitJSON(settings, value)
 			settings.failfast = oldfailfast
 			if err == nil {
-				ok++
+				if schema.Discriminator != nil {
+					pn := schema.Discriminator.PropertyName
+					if valuemap, okcheck := value.(map[string]interface{}); okcheck {
+						if discriminatorVal, okcheck := valuemap[pn]; okcheck == true {
+							mapref, okcheck := schema.Discriminator.Mapping[discriminatorVal.(string)]
+							if okcheck && mapref == item.Ref {
+								ok++
+							}
+						}
+					}
+				} else {
+					ok++
+				}
 			}
 		}
 		if ok != 1 {
--- a/openapi3filter/validation_discriminator_test.go
+++ b/openapi3filter/validation_discriminator_test.go
@@ -1,0 +1,98 @@
+package openapi3filter
+
+import (
+	"bytes"
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


fenollp:
Hi! Thanks for taking time to do this.
Could you add your example code as a test? I can add #80's code as test as well later if you can't do it.
Some notes:
* do `okcheck` instead of `okcheck == true`
* casting to string needs a check as it panics if it isn't a string

In what way do you expect this to break other cases?

RiccardoManfrin:
> Hi! Thanks for taking time to do this.
> Could you add your example code as a test? 
will try
> I can add #80's code as test as well later if you can't do it.
> Some notes:
> 
>     * do `okcheck` instead of `okcheck == true`
will do.. really I didn't mean to make it prod grade
> 
>     * casting to string needs a check as it panics if it isn't a string
will do
> 
> In what way do you expect this to break other cases?
I've no idea but this seems to be so needed (at least by me) that I thought it could not be _so easy_ unless something else was breaking..

fenollp:
This patch looks pretty similar: https://github.com/xeipuuv/gojsonschema/pull/297/files so I guess you're on point :)

PR #323, opened 2021-03-10
reproduce incorrect discriminator handling with ValidateRequest

fenollp:
Reproduces #80 
cc #321 

@RiccardoManfrin Could you take a look at this as you've touched discriminator recently please?That'd help a lot :)

### DIFF
--- a/openapi3filter/validate_request_test.go
+++ b/openapi3filter/validate_request_test.go
@@ -26,7 +26,7 @@ paths:
         content:
           application/json:
             schema:
-              anyOf:
+              oneOf:
               - $ref: '#/components/schemas/Cat'
               - $ref: '#/components/schemas/Dog'
               discriminator:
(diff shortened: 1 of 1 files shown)


RiccardoManfrin:
From a first look you are missing the "mapping" key which the code I wrote expected all along with the discriminator.
Like so: https://github.com/getkin/kin-openapi/blob/master/openapi3filter/validation_discriminator_test.go#L42

fenollp:
@RiccardoManfrin Thanks! So it turns out `mapping` isn't required. It all boiled down to oneOf v. allOf.

ohasanliUW:
reviving this thread. this PR has not resolved the validation for handling discriminator, as far as I understand. The unit test does not cover negative test cases where validation should fail. I added one locally and it incorrectly passes validation.

fenollp:
@ohasanliUW Please open a PR with this test.

PR #328, opened 2021-03-21
new: add detetion for undescribed request parameters

Alexey19:
Added the ability, when checking a request, to find parameters that are not described in the schema.

### DIFF
--- a/openapi3filter/options.go
+++ b/openapi3filter/options.go
@@ -11,7 +11,8 @@ type Options struct {
 	ExcludeResponseBody   bool
 	IncludeResponseStatus bool
 
-	MultiError bool
+	MultiError             bool
+	FailOnUnknownParameter bool
 
 	// See NoopAuthenticationFunc
 	AuthenticationFunc AuthenticationFunc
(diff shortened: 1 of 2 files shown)


fenollp:
does openapi allow unspecified query params?

Alexey19:
> does openapi allow unspecified query params?

I think yes, some kind. But this functionality added for user, who has, for example, a lot of request/response and want to ensure, what specification consists full information about all parameters (later I want to add same check for request and response body).

Alexey19:
@fenollp , what about pull-request?

fenollp:
Hi! This is interesting. Are you still looking to merge this?

I'm thinking that the OpenAPIv3 spec does not treat query params a certain way. Schemas do allow "undescribed query params" in that a schema for a parameter of type "query" *may* sport the setting `additionalProperties: true`.

If you want a route not to accept query fields that are not listed in the spec, use `additionalProperties: false`:
```yaml
# ...
- name: MyQueryParam
  in: query
  schema:                                    # only allows "?myOnlyProperty=42"
    type: object
    additionalProperties: false
    properties:
      myOnlyProperty:
        type: number
# ...
```

fenollp:
@Alexey19 Feel free to re-open if needed.

PR #338, opened 2021-04-08
As per commit message

RiccardoManfrin:
Commit 5547859892303a538277b0e9411564507a2b19fa permits, when the discriminator and mapping are used to `visitJSON` for the specific `Schema` without losing time, and to return the specific error instead of the generic one returned when mapping is not specified.


fenollp:
Cool so can't the duplicate code be removed? and if not could you avoid the duplication by introducing a function please.

openapi3/schema.go
```diff
@@ -837,6 +838,37 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
+						}
+					} else {
+						/* Assume implicit mapping on objectType as stated in Mapping Type Names section:
+						``It is implied, that the property to which discriminator refers, contains the
+						name of the target schema. In the example above, the objectType property should
+						contain either simpleObject, or complexObject string.''*/
+						for _, oneof := range schema.OneOf {
+							/* TODO: ugly.. should this be a property of the SchemaRef? */
+							objectType := strings.ReplaceAll(oneof.Ref, "#/components/schemas/", "")
+							if objectType == discriminatorVal {
```
fenollp, line 863:
```suggestion
							if strings.HasSuffix(oneof.Ref, "/"+discriminatorVal) {
```

fenollp: Although I'm unsure why this whole `} else {` clause is needed?

RiccardoManfrin: Do you mean the busness logic _within_ the else? it supports the discriminator when the mapping is implicit the TestExample fails without such else

openapi3filter/validate_request_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"testing"
```
fenollp, line 8: What made you edit this file?

openapi3/schema.go
```diff
@@ -837,6 +838,35 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
 
 	if v := schema.OneOf; len(v) > 0 {
+
+		if schema.Discriminator != nil {
+			/* Find mapped object by ref */
+			if valuemap, okcheck := value.(map[string]interface{}); okcheck {
+				pn := schema.Discriminator.PropertyName
+				if discriminatorVal, okcheck := valuemap[pn]; okcheck {
+					if len(schema.Discriminator.Mapping) > 0 {
+						if mapref, okcheck := schema.Discriminator.Mapping[discriminatorVal.(string)]; okcheck {
```
fenollp, line 848: Can't casting to string go away? There's a cast to `map[string]interface{}` before.

openapi3/schema.go
```diff
@@ -837,6 +838,35 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
+								}
+							}
+						}
+					} else {
+						/* Assume implicit mapping on objectType as stated in Mapping Type Names section:
+						``It is implied, that the property to which discriminator refers, contains the
+						name of the target schema. In the example above, the objectType property should
+						contain either simpleObject, or complexObject string.''*/
+						for _, oneof := range schema.OneOf {
+							if strings.HasSuffix(oneof.Ref, discriminatorVal.(string)) {
```
fenollp, line 861: Same question about casting to string

openapi3/schema.go
```diff
@@ -837,6 +838,35 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
+									return item.Value.visitJSON(settings, value)
+								}
+							}
+						}
+					} else {
+						/* Assume implicit mapping on objectType as stated in Mapping Type Names section:
+						``It is implied, that the property to which discriminator refers, contains the
+						name of the target schema. In the example above, the objectType property should
+						contain either simpleObject, or complexObject string.''*/
+						for _, oneof := range schema.OneOf {
```
fenollp, line 860: Use `v`

openapi3/schema.go
```diff
@@ -837,6 +838,35 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
 	if v := schema.OneOf; len(v) > 0 {
+
+		if schema.Discriminator != nil {
+			/* Find mapped object by ref */
+			if valuemap, okcheck := value.(map[string]interface{}); okcheck {
+				pn := schema.Discriminator.PropertyName
+				if discriminatorVal, okcheck := valuemap[pn]; okcheck {
+					if len(schema.Discriminator.Mapping) > 0 {
+						if mapref, okcheck := schema.Discriminator.Mapping[discriminatorVal.(string)]; okcheck {
+							for _, item := range v {
```
fenollp, line 849: s/item/oneof/g

fenollp (changes requested):

fenollp:
Just a few nits but this is otherwise correct.

openapi3filter/validate_request_test.go
RiccardoManfrin: I realized I had not made it compliant with go test requirements (Test fnction prefix and args). The result was that when I run the tests I could not track where the specific problem was among the tests

openapi3/schema.go
RiccardoManfrin: Do you mean to to type assert to `map[string]string` (<- this doesn't work)

RiccardoManfrin: erlang would have been easier I guess :)

openapi3filter/validate_request_test.go
fenollp: Please revert changes to this file. 

See https://blog.golang.org/examples section "Larger examples".

openapi3/schema.go
fenollp: I meant the casting on this line i.e. `discriminatorVal.(string)`

RiccardoManfrin: If I don't force the cast I get this error
`schema.go:848:57: cannot use discriminatorVal (type interface {}) as type string in map index: need type assertion`

fenollp: Oh right! Sorry, my mind stopped working...

fenollp:
Looks like the test fails to print the expected error. You can use this to reproduce:
```
go test -v -run Example ./openapi3filter
```

RiccardoManfrin:
> Looks like the test fails to print the expected error. You can use this to reproduce:
> 
> ```
> go test -v -run Example ./openapi3filter
> ```

I didn't even know about this notation .. I searched for where Example was called, expecting the validation would come from outside or something ..didn't know it was in the comments..

So that is where the else would come into play (from your original question)

openapi3filter/validate_request_test.go
```diff
@@ -107,26 +107,4 @@ func Example() {
...
-	//       }
-	//     ]
-	//   }
-	//
-	// Value:
-	//   {
-	//     "bark": true,
-	//     "breed": "Dingo",
-	//     "pet_type": "Cat"
-	//   }
```
fenollp, line 131: I understand the request data should be invalid per the provided schema. What am I missing?

The test used to not pass because there was no output, because no error is returned. Are you sure this should not output anything?

fenollp (changes requested):

openapi3filter/validate_request_test.go
RiccardoManfrin: This relates to your original question about not being sure regarding the presence of the "`else`" statement:

Quoting from https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/

>"It is implied, that the property to which discriminator refers, contains the name of the target schema. In the example above, the objectType property should contain either simpleObject, or complexObject string. If the property values do not match the schema names, you can map the values to the names. To do this, use the discriminator/mapping keyword:"

I'm interpreting this as saying that if you don't specify the mapping the default mapping is based on the names of the objects included in the oneOf list: `simpleObject`, `complexObject` in the spec doc, `Dog` or `Cat` in the `Example()` from `validate_request_test.go`.

Let me know if you believe I'm getting it all wrong or it makes sense.

PR #346, opened 2021-04-23
Fix following refs to non-openapi3 root documents (but that are sub-documents)

### LINKED ISSUE #344: Local references not resolved correctly in external files
From https://swagger.io/docs/specification/using-ref/:

> Local Reference – $ref: '#/definitions/myElement' # means go to the root of the current document and then find elements definitions and myElement one after one.

However in kin-openapi when the specification includes a Remote Reference to a document with a Local reference, the local references in the document cannot be found.

Example:

*ext.json*
```json
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "definitions": {
        "a": {
            "type": "string"
        },
        "b": {
            "type": "object",
            "properties": {
                "name": {
                    "$ref": "#/definitions/a"
                }
            }
        }
    }
}
```

*spec.yaml*
```yaml
openapi: 3.0.1
components:
  schemas:
    Test:
      type: object
      properties:
        test:
          $ref: 'ext.json#/definitions/b'
```

Loading the spec (package openapi3) raises an error because it can't find `definitions` when trying to resolve `#/definitions/a` in ext.json. Code to reproduce is available in https://gist.github.com/roee88/626d93f9e73776c7f1ec3aa59d7db988#file-extref_test-go. 

Changing `#/definitions/a` to `ext.json#/definitions/a` in the `ext.json` file bypass the problem but this is not an acceptable solution because often you don't have control of the referenced files (plus I suspect it makes them invalid when used independently but unsure).

fenollp:
Fixes #344


roee88:
Thank you!

PR #348, opened 2021-04-23
[Bugfix] fixed error message when only file is referenced

steffakasid:
@fenollp you were right: The error should include the location. The problem is, that the handling is different if the reference is just a file without a reference (\#....). I added a test case for that and aligned the error message.


fenollp:
Thanks for going the extra mile :)

PR #351, opened 2021-04-26
Clean APIs from trademarked name "Swagger"

fenollp:


mgabeler-lee-6rs:
@fenollp I wonder if you could perhaps comment on the motivation for this change over on https://github.com/deepmap/oapi-codegen/pull/348 ?

PR #355, opened 2021-04-29
Have Header Object follow the structure of the Parameter Object

### LINKED ISSUE #46: Header type should have the same fields as Parameter
Hi,

I'm dinning in OpenAPIv3 spec and just figured out that Header object fields should be the same as in Parameter.

[Header Object](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#headerObject) :
```
Header Object

The Header Object follows the structure of the Parameter Object with the following changes:

    name MUST NOT be specified, it is given in the corresponding headers map.
    in MUST NOT be specified, it is implicitly in header.
    All traits that are affected by the location MUST be applicable to a location of header (for example, style).
```

fenollp:
Fixes #46

@pruser @danielvladco can you confirm this meets your needs? I'm asking specifically about the awkward way to construct a `Header.Value`:
```go
			Headers: map[string]*HeaderRef{
				"someHeader": {
					Ref: "#/components/headers/otherHeader",
				},
				"otherHeader": {
					Value: &Header{Parameter{Schema: &SchemaRef{Value: NewStringSchema()}}},
				},
			},
```

### DIFF
--- a/openapi3/header.go
+++ b/openapi3/header.go
@@ -2,6 +2,7 @@ package openapi3
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/getkin/kin-openapi/jsoninfo"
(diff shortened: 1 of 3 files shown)


danielvladco:
@fenollp the changes look fine, but there is no way to check if this works without cloning the entire repo. 
Also I don't think it is a problem when constructing a header. I don't see anything wrong with it...

fenollp:
> there is no way to check if this works without cloning the entire repo.

https://golang.org/ref/mod#go-mod-file-replace

danielvladco:
Hi @fenollp I tested it and seems to work fine. When are you planning to merge this?

PR #372, opened 2021-06-18
openapi3: resolveComponent should resolve SchemaRef, too

tgulacsi:


fenollp:
If you can share a minimized OpenAPI document that is valid under SoapUI but not with this lib that'd be great.

Also note that your test is only doing deserialization and dereferencing, not [validation](https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3#T.Validate).

tgulacsi:
TestLoadWithSchemaRef contains such an document:
```
openapi: 3.0.1
info:
  title: API
  description: This is a simple API
  contact:
    email: admin@example.com
  license:
    name: Apache 2.0
    url: http://www.apache.org/licenses/LICENSE-2.0.html
  version: 1.0.0
servers:
- url: https://host/xxx/interface/v1
paths:
  /notification:
    post:
      tags:
       - notification
      summary: summary
      description: change notification
      operationId: postMessage
      requestBody:
        description: Notification
        content:
          application/json:
            schema:
              type: array
              items:
               $ref: '#/components/schemas/Message'
          application/xml:
            schema:
              type: array
              xml: 
               name: Message
              items:            
               $ref: '#/components/schemas/Message'              
        required: false
      responses:
        201:
          description: item created
        400:
          description: invalid input, object invalid
        409:
           $ref: '#/components/schemas/error'          
        
   
components:
  schemas:

    Message:
      required:
      - eventTypeName
      - msgID
      type: object
      properties:
        eventTypeName:
          type: string
          example: earthquake
        msgID:
          minimum: 0
          type: integer
          format: int32

    error:
      required:
        - code
        - message
      properties:
        code: 
          type: integer
          format: int32
          example: 100
        message:
          type: string             
```

Without this modification, it fails with `bad data in "#/components/schemas/error"` - resolveComponent meets with a *SchemaRef, which it cannot resolve.

fenollp:
Right so it looks like your response object is malformed: its schema must be nested under
```yaml
        content:
          application/json:
            schema:
```

Tools such as https://editor.swagger.io/ parse your document fine but you'll notice it shows no schema for response 409. My guess is their parser is looking for one of these fields
https://github.com/getkin/kin-openapi/blob/9f8b1acb3de3dc51c4dfe01fc396d695b84fc228/openapi3/response.go#L59-L62
and is disregarding the other ones.
Maybe SoapUI has some settings to show fields that are not expected by the parser? Maybe you'd like to talk to them about adding a warning about this.
I know I'd like to fail validation on such a case :)

PR #379, opened 2021-06-25
Add uint type to openapi3gen

biancarosa:
Didn't open an issue because it was pretty straightforward!

Lemme know if I need to change something, I'm happy to.

### DIFF
--- a/openapi3gen/openapi3gen_test.go
+++ b/openapi3gen/openapi3gen_test.go
@@ -50,6 +50,6 @@ func TestExportUint(t *testing.T) {
 	require.Equal(t, &openapi3.SchemaRef{Value: &openapi3.Schema{
 		Type: "object",
 		Properties: map[string]*openapi3.SchemaRef{
-			"uint": {Value: &openapi3.Schema{Type: "integer"}},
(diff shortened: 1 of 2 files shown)


openapi3gen/openapi3gen_test.go
```diff
@@ -39,3 +39,17 @@ func TestExportedNonTagged(t *testing.T) {
...
+	type UnsignedIntStruct struct {
+		UnsignedInt uint `json:"uint"`
+	}
+
+	schemaRef, _, err := NewSchemaRefForValue(&UnsignedIntStruct{}, UseAllExportedFields())
+	require.NoError(t, err)
+	require.Equal(t, &openapi3.SchemaRef{Value: &openapi3.Schema{
+		Type: "object",
+		Properties: map[string]*openapi3.SchemaRef{
+			"uint": {Value: &openapi3.Schema{Type: "integer"}},
```
fenollp, line 53: More then `type: integer` this should probably also contain `minimum: 0`.
See the other cases such as
```go
	case reflect.Uint16:
		schema.Type = "integer"
		schema.Min = &zeroInt
		schema.Max = &maxUint16
```

fenollp (changes requested):

fenollp (approved):

PR #386, opened 2021-07-27
Detect if a field is anonymous and handle the indirection

DerekStrickland:


fenollp:
Can you add a test with an anonymous struct embedding? Thanks

DerekStrickland:
@fenollp  Absolutely!  Sorry, I was rushing

DerekStrickland:
@fenollp  Done!

PR #392, opened 2021-07-29
Support reference cycles

### LINKED ISSUE #387: [Question]: Intentional cycles in structs
We have a struct in our project that allows us to do recursive diffs on arbitrary objects. It is defined as this.

```
type ObjectDiff struct {
	Type    DiffType
	Name    string
	Fields  []*FieldDiff
	Objects []*ObjectDiff
}
```

This allows users to perform a dry-run type operation against the API to see what would change if they were to actually submit the operation.

When trying to use kin to generate an  OpenAPI spec, this, of course, throws a CycleError.  Is there any existing feature or configuration option that can allow us to detect the cycle and keep on processing without the infinite recursion?

DerekStrickland:
Closes #387


DerekStrickland:
Found a potential bug.  Closing till I fix.

PR #393, opened 2021-07-30
Support reference cycles

### LINKED ISSUE #387: [Question]: Intentional cycles in structs

DerekStrickland:
Closes #387


openapi3gen/openapi3gen_test.go
```diff
@@ -8,20 +8,6 @@ import (
...
 )
 
-type CyclicType0 struct {
-	CyclicField *CyclicType1 `json:"a"`
-}
-type CyclicType1 struct {
-	CyclicField *CyclicType0 `json:"b"`
-}
-
-func TestCyclic(t *testing.T) {
```
fenollp, line 18: Can you add a generator option that keeps the previous default behavior of returning an error? I'd like this capability to still be here, while changing to your work as the default.

openapi3gen/openapi3gen.go
```diff
@@ -104,6 +105,11 @@ func (g *Generator) generateWithoutSaving(parents []*jsoninfo.TypeInfo, t reflec
 		if a && b {
 			vs, err := g.generateSchemaRefFor(parents, v.Type)
 			if err != nil {
+				// TODO: this needs code review
```
fenollp, line 108: LGTM. What are you concerned with?

fenollp (changes requested):

openapi3gen/openapi3gen_test.go
DerekStrickland: For sure

DerekStrickland: For sure

openapi3gen/openapi3gen.go
DerekStrickland: Just wanted your eyes because I didn't feel confident I understood the purpose of the code block correctly.

openapi3gen/openapi3gen_test.go
DerekStrickland: Option added and ready for review @fenollp

Issue #398, opened 2021-08-10
multipart/form-data's part may have no Content-Type


stakme:
Hi, thanks your great work!

I have a trouble to validate requests in `multipart/form-data` format; requests have parts without `Content-Type` Header, but validation refused them.

## multipart requests

According to HTML Living Standard, `Content-Type` Header of parts of `multipart/form-data` must be removed in specific patterns. 

> The parts of the generated multipart/form-data resource that correspond to non-file fields must not have a `Content-Type` header specified.
> https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#multipart-form-data

like this:

```
Content-Type: multipart/form-data; boundary=XXXXX; 
...
--XXXXX
Content-Disposition: form-data; name="id"

123
--XXXXX
Content-Disposition: form-data; name="image"; filename="sample.png"
Content-Type: image/png
...
--XXXXX
```

`image` has Content-Type Header, but `id` doesn't.

## validation

Now `multipartBodyDecoder` depends on `decodeBody` function,

https://github.com/getkin/kin-openapi/blob/fba0a14e84c1efe61f72dd9d5806fc5605fc6f2e/openapi3filter/req_resp_decoder.go#L963

and `decodeBody` returns an error when the part has no `Content-Type` header.

https://github.com/getkin/kin-openapi/blob/fba0a14e84c1efe61f72dd9d5806fc5605fc6f2e/openapi3filter/req_resp_decoder.go#L811-L817

PR #399, opened 2021-08-10
Accept multipart/form-data's part without Content-Type

stakme:
handle non-file parts of `multipart/form-data` as `text/plain`. ref https://github.com/getkin/kin-openapi/issues/398


fenollp:
LGTM. Thanks!

PR #421, opened 2021-09-20
Add support for "application/x-yaml"

luminoso:
Yaml follows the same flow as json, but with a different unmarshaller.

Added both Contet-Types:

- `application/yaml`
- `application/x-yaml`

Unsure if `text/yaml` should also be added

Reused currently yaml.v2 used by the project


fenollp:
LGTM. Only requires a go.mod change to pass CI.

PR #433, opened 2021-10-04
Fix #422 added support for error unwrapping for errors with a single sub-error

### LINKED ISSUE #422: unwrap for internal errors
Implement https://pkg.go.dev/errors#Unwrap on internal errors like `openapi3filter.RequestError` and `openap3.SchemaError` so we can use `errors.Is`/`errors.As` to inspect them

My specific use case is that I have a `*openapi3filter.RequestError` where the Err property is a `*openapi3.SchemaError`, and I want to call JSONPointer() on it to get the path where the error occurred. Right now that looks like this:

```
var e *openapi3filter.RequestError
if goerrors.As(err, &e) {
	var e2 *openapi3.SchemaError
	if goerrors.As(e.Err, &e2) {
		path := e2.JSONPointer()
		fmt.Println(path)
	}
}
```

It could look like this:
```
var e *openapi3.SchemaError
if goerrors.As(err, &e) {
	path := e.JSONPointer()
	fmt.Println(path)
}
```

bionoren:
MultiError and SecurityRequirementsError are trickier - they'd need to be rethought as linked lists of errors. I can do that if you're interested, but I wanted to start with the easy wins.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1489,6 +1489,8 @@ type SchemaError struct {
 	Origin      error
 }
 
+var _ interface{ Unwrap() error } = SchemaError{}
+
 func markSchemaErrorKey(err error, key string) error {
 	if v, ok := err.(*SchemaError); ok {
 		v.reversePath = append(v.reversePath, key)
@@ -1564,7 +1566,7 @@ func (err *SchemaError) Error() string {
 	return buf.String()
 }
 
-func (err *SchemaError) Unwrap() error {
+func (err SchemaError) Unwrap() error {
 	return err.Origin
 }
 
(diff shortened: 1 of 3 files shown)


fenollp:
Sure if you want to you can always add this later. 

Before thought I'd like some way to test this, either by calling As and Is or just by ensuring some interface is implemented (with `var _ iface = ...`)

bionoren:
I've added compile-time checks defining the interface inline, which seems to be the expectation per https://github.com/golang/go/issues/39539. FWIW, go 1.17 added a vet check that if an `error` type has an `Unwrap` method, then it must conform to `errors.Unwrap`'s expectations.

fenollp:
LGTM but I'm not sure why the Go github actions aren't running.

PR #443, opened 2021-11-04
Internalize references

jhwz:
I needed this functionality for some tooling I'm building and thought I would make a pull request in-case anyone would find it useful. If so I can create a few tests too.

All this does is either inline or add to the components section all external references, so that when you write the file back out it is self contained and portable. This makes it easier for other tools to process and removes all dependencies.

If there's a better way of doing this let me know and I will update my code!

### DIFF
--- a/openapi3/internalize_refs.go
+++ b/openapi3/internalize_refs.go
@@ -50,7 +50,7 @@ func isExternalRef(ref string) bool {
 	return ref != "" && !strings.HasPrefix(ref, "#/components/")
 }
 
-func (doc *T) addSchemaToSpec(s *SchemaRef, refNameResolver RefNameResolver) string {
+func (doc *T) addSchemaToSpec(s *SchemaRef, refNameResolver RefNameResolver) {
 	if s == nil || !isExternalRef(s.Ref) {
 		return ""
 	}
--- a/openapi3/internalize_refs.go
+++ b/openapi3/internalize_refs.go
@@ -86,7 +86,7 @@ func (doc *T) addSchemaToSpec(s *SchemaRef, refNameResolver RefNameResolver) {
 	return s.Ref
 }
 
-func (doc *T) addParameterToSpec(p *ParameterRef, refNameResolver RefNameResolver) string {
+func (doc *T) addParameterToSpec(p *ParameterRef, refNameResolver RefNameResolver) {
 	if p == nil || !isExternalRef(p.Ref) {
 		return ""
 	}
--- a/openapi3/internalize_refs.go
+++ b/openapi3/internalize_refs.go
@@ -2,9 +2,33 @@ package openapi3
 
 import (
 	"context"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
+
+type RefNameResolver func(string) string
+
+// DefaultRefResolver is a default implementation of refNameResolver for the
+// InternalizeRefs function.
+//
+// If a reference points to an element inside a document, it returns the last
+// element in the reference using filepath.Base. Otherwise if the reference points
+// to a file, it returns the file name trimmed of all extensions.
+func DefaultRefNameResolver(ref string) string {
+	if ref == "" {
+		return ""
+	}
+	split := strings.SplitN(ref, "#", 2)
+	if len(split) == 2 {
+		return filepath.Base(split[1])
+	}
+	ref = split[0]
+	for ext := filepath.Ext(ref); len(ext) > 0; ext = filepath.Ext(ref) {
+		ref = strings.TrimSuffix(ref, ext)
+	}
+	return filepath.Base(ref)
+}
 
 func schemaNames(s Schemas) []string {
 	out := make([]string, 0, len(s))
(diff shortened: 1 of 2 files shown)


fenollp:
Please add tests, specifically one with recursive references.

jhwz:
Should be all there now. I spent some time and made the implementation complete.

I have tests on an array of files found in the testdata directory.

Cheers!

fenollp:
Ah you'll need your tests to compile for Go 1.14 please (that `os.WriteFile` call)
Also please have these tests define `err` as in
```go
err := os.WriteFile("__internalized.openapi.json", data, 0644)
require.NoError(t, err, "writing internalized spec to new context")
```
instead of:
```go
require.NoError(t, os.WriteFile("__internalized.openapi.json", data, 0644), "writing internalized spec to new context")
```

And finally, no need to write to file, just use `LoadFromData` or `LoadFromDataWithPath`
https://pkg.go.dev/github.com/getkin/kin-openapi@v0.80.0/openapi3#Loader.LoadFromData

jhwz:
Ah right my mistake.

All good now, there is one issue with the recursive references though:
```
              "components": {
                "schemas": {
                  "Bar": {
                    "$ref": "#/components/schemas/Bar"
                  },
                  "Foo": {
                    "$ref": "#/components/schemas/Foo"
                  },
                  "Foo2": {
                    "$ref": "#/components/schemas/Foo2"
                  }
                }
              },
              "info": {
                "title": "Recursive refs example",
                "version": "1.0"
              },
              "openapi": "3.0.3",
              "paths": {
                "/foo": {
                  "get": {
                    "responses": {
                      "200": {
                        "content": {
                          "application/json": {
                            "schema": {
                              "properties": {
                                "foo2": {
                                  "$ref": "#/components/schemas/Foo2"
                                }
                              },
                              "type": "object"
                            }
                          }
                        },
                        "description": "OK"
                      }
                    }
                  }
                }
              }
            }
```

Which is produced by the function on the recursiveRefs schema. I'm not sure you would consider this schema valid, but maybe. The issue is it is not caught by [this](https://github.com/getkin/kin-openapi/blob/e9b36dac1a2890474ae44cfcf6446c00393527b5/openapi3/loader.go#L688-L696) because the value is always nil.

Not sure how you want to deal with that..

fenollp:
Hm this is not how that document should be "internalized". Bar's schema should be 
```yaml
type: string
example: bar
```
I think you're right in pointing out the "visited" clause is ineffectual but I don't see how `component.Value` is `nil`. 
Can you help fix this?

jhwz:
Definitely, I missed that the schema didn't reference itself directly so it must be something I did wrong.

I'll have a look soon and address the issue!

jhwz:
OK! Maybe this time. I found the issue

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,373 @@
...
+	return out
+}
+
+// InternalizeRefs removes all references to external files from the spec and moves them
+// to the components section.
+//
+// refNameResolver takes in references to returns a name to store the reference under locally.
+// Some care should be taken to make sure the resolver does not return duplicate names for different
+// references.
+func (spec *T) InternalizeRefs(ctx context.Context, refNameResolver func(ref string) string) *T {
```
fenollp, line 31: * Please use `doc` identifier instead of `spec` (as is done throughout the source)
* Any reason to output `*T`? This seems redundant as the receiver is a pointer.
* `error` should be outputted, instead of `panic`ing on loops.
* `refNameResolver` should have a default implementation (ie. the one given in tests)

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+		if spec.Components.Schemas == nil {
+			spec.Components.Schemas = make(Schemas)
+		}
+		spec.Components.Schemas[name] = s.Value.NewRef()
+		s.Ref = "#/components/schemas/" + name
+		return s.Ref
+	}
+
+	addParameterToSpec := func(p *ParameterRef) string {
```
fenollp, line 72: Why not define all these closures as functions of `*T`? e.g.
```go
func (doc *T) addParameterToSpec(p *ParameterRef) string { ... }
```
IMO it would be preferable.

openapi3/internalize_refs_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func baseResolver(ref string) string {
```
fenollp, line 13: Export this as the default resolver

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,373 @@
...
+		}
+		spec.Components.Callbacks[name] = &CallbackRef{Value: c.Value}
+		c.Ref = "#/components/callbacks/" + name
+		return c.Ref
+	}
+
+	var derefSchema func(*Schema, []*Schema)
+	derefSchema = func(s *Schema, stack []*Schema) {
+		if len(stack) > 1000 {
+			panic("unresolved circular reference")
```
fenollp, line 212: Return an error here instead.
Do you have an example triggering this clause? If you do I'd like a test for this, if not isn't this dead code?

fenollp (changes requested):

openapi3/internalize_refs.go
jhwz: Yeah I don't, I can remove it

openapi3/internalize_refs.go
jhwz: No problem at all, it did turn into more of a mess than I was anticipating

openapi3/internalize_refs.go
jhwz: - No problem!
- Habit I am in for my APIs to indicate you are mutating the state, but yes its redundant I'll remove it
- As later noted I will just remove it
- Easy!

jhwz:
I've addressed those issues now. Because we don't use the closures we need to pass around the resolver everywhere, but its not too bad

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+// InternalizeRefs removes all references to external files from the spec and moves them
+// to the components section.
+//
+// refNameResolver takes in references to returns a name to store the reference under locally.
+// A default implementation is provided that will suffice for most use cases. See the function
+// documention for more details.
+//
+// Example:
+//
+//   doc.InternalizeRefs(context.Background(), DefaultRefNameResolver)
```
fenollp, line 344: Actually I was thinking of `nil` being valid. ie.
```go
//   doc.InternalizeRefs(context.Background(), nil)
...
	if refNameResolver == nil {
		refNameResolver = DefaultRefNameResolver
	}
```

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+		out = append(out, i)
+	}
+	return out
+}
+
+func isExternalRef(ref string) bool {
+	return ref != "" && !strings.HasPrefix(ref, "#/components/")
+}
+
+func (doc *T) addSchemaToSpec(s *SchemaRef, refNameResolver RefNameResolver) string {
```
fenollp, line 53:
```suggestion
func (doc *T) addSchemaToSpec(s *SchemaRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.Schemas == nil {
+		doc.Components.Schemas = make(Schemas)
+	}
+	doc.Components.Schemas[name] = s.Value.NewRef()
+	s.Ref = "#/components/schemas/" + name
+	return s.Ref
+}
+
+func (doc *T) addParameterToSpec(p *ParameterRef, refNameResolver RefNameResolver) string {
```
fenollp, line 89:
```suggestion
func (doc *T) addParameterToSpec(p *ParameterRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+
+	if doc.Components.Parameters == nil {
+		doc.Components.Parameters = make(ParametersMap)
+	}
+	doc.Components.Parameters[name] = &ParameterRef{Value: p.Value}
+	p.Ref = "#/components/parameters/" + name
+	return p.Ref
+}
+
+func (doc *T) addHeaderToSpec(h *HeaderRef, refNameResolver RefNameResolver) string {
```
fenollp, line 107:
```suggestion
func (doc *T) addHeaderToSpec(h *HeaderRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.Headers == nil {
+		doc.Components.Headers = make(Headers)
+	}
+	doc.Components.Headers[name] = &HeaderRef{Value: h.Value}
+	h.Ref = "#/components/headers/" + name
+	return h.Ref
+}
+
+func (doc *T) addRequestBodyToSpec(r *RequestBodyRef, refNameResolver RefNameResolver) string {
```
fenollp, line 124:
```suggestion
func (doc *T) addRequestBodyToSpec(r *RequestBodyRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.RequestBodies == nil {
+		doc.Components.RequestBodies = make(RequestBodies)
+	}
+	doc.Components.RequestBodies[name] = &RequestBodyRef{Value: r.Value}
+	r.Ref = "#/components/requestBodies/" + name
+	return r.Ref
+}
+
+func (doc *T) addResponseToSpec(r *ResponseRef, refNameResolver RefNameResolver) string {
```
fenollp, line 141:
```suggestion
func (doc *T) addResponseToSpec(r *ResponseRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.Responses == nil {
+		doc.Components.Responses = make(Responses)
+	}
+	doc.Components.Responses[name] = &ResponseRef{Value: r.Value}
+	r.Ref = "#/components/responses/" + name
+	return r.Ref
+}
+
+func (doc *T) addSecuritySchemeToSpec(ss *SecuritySchemeRef, refNameResolver RefNameResolver) string {
```
fenollp, line 158:
```suggestion
func (doc *T) addSecuritySchemeToSpec(ss *SecuritySchemeRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.SecuritySchemes == nil {
+		doc.Components.SecuritySchemes = make(SecuritySchemes)
+	}
+	doc.Components.SecuritySchemes[name] = &SecuritySchemeRef{Value: ss.Value}
+	ss.Ref = "#/components/securitySchemes/" + name
+	return ss.Ref
+}
+
+func (doc *T) addExampleToSpec(e *ExampleRef, refNameResolver RefNameResolver) string {
```
fenollp, line 175:
```suggestion
func (doc *T) addExampleToSpec(e *ExampleRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.Examples == nil {
+		doc.Components.Examples = make(Examples)
+	}
+	doc.Components.Examples[name] = &ExampleRef{Value: e.Value}
+	e.Ref = "#/components/examples/" + name
+	return e.Ref
+}
+
+func (doc *T) addLinkToSpec(l *LinkRef, refNameResolver RefNameResolver) string {
```
fenollp, line 192:
```suggestion
func (doc *T) addLinkToSpec(l *LinkRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

openapi3/internalize_refs.go
```diff
@@ -0,0 +1,389 @@
...
+	}
+	if doc.Components.Links == nil {
+		doc.Components.Links = make(Links)
+	}
+	doc.Components.Links[name] = &LinkRef{Value: l.Value}
+	l.Ref = "#/components/links/" + name
+	return l.Ref
+}
+
+func (doc *T) addCallbackToSpec(c *CallbackRef, refNameResolver RefNameResolver) string {
```
fenollp, line 209:
```suggestion
func (doc *T) addCallbackToSpec(c *CallbackRef, refNameResolver RefNameResolver) {
```
No call to this function actually use the returned value so let's drop that.

fenollp (changes requested):
Sorry I missed all these in my previous review. This should be the last.
Keep up the good work!

jhwz:
All good, thanks for taking the time on this PR!

PR #449, opened 2021-11-18
ClientCredentials conversion to OpenAPI v2

Luukvdm:
ClientCredentials was called application in OpenAPI v2 so this needs to be converted.

Specification:
https://swagger.io/specification/#oauth-flows-object


fenollp:
Thanks. I understand this is a pretty minor change still could you add a regression test?

Luukvdm:
Is this sufficient? I'm new to writing tests in Golang so let me know if I should change or improve something.

Issue #447, opened 2021-11-12
Recursive definitions cause infinite looping


chrisdostert:
If you attempt to generate models for a schema like below, it results in an infinite loop that never completes.

This was originally mentioned in https://github.com/getkin/kin-openapi/issues/309. 

```yaml
openapi: 3.0.1
info:
  title: Recursive refs example
  version: "1.0"
paths: {}
components:
  schemas:
    Complex:
      type: object
      properties:
        parent:
          $ref: '#/components/schemas/Complex'
```

fenollp:
Hi Chris do you mind telling us a bit more about how to produce the bug? 
Loading seems fine: https://github.com/getkin/kin-openapi/pull/448

nguyer:
I think I may have run into this same issue. I have a Go struct that looks like this:

```go
type FFIParam struct {
	Name         string      `json:"name"`
	InternalType string      `json:"internalType"`
	Type         string      `json:"type"`
	Components   []*FFIParam `json:"components,omitempty"`
}
```

And when trying to generate the openapi spec for it, it appears that it goes into an infinite loop and overflows the stack. I've attached the relevant snippet of the log from where this happens. Happy to provide more detail on the code that's generating the openapi if needed.

[log.txt](https://github.com/getkin/kin-openapi/files/7589665/log.txt)

peterbroadhurst:
Created the following UT in `openapi3gen` to reproduce the issue, and work on a fix:

```go
func TestRecursiveSchema(t *testing.T) {

	type RecursiveType struct {
		Field1     string           `json:"field1"`
		Field2     string           `json:"field2"`
		Field3     string           `json:"field3"`
		Components []*RecursiveType `json:"children,omitempty"`
	}

	schemaRef, _, err := NewSchemaRefForValue(&RecursiveType{})
	require.NoError(t, err)

	_, err = schemaRef.MarshalJSON() // *** this blows the stack ***
	require.NoError(t, err)

}
```

fenollp:
@chrisdostert Hi! Are you able to produce a code sample that reproduces your issue?

I think the Kaleido people are experiencing a different issue, from the little code you shared...

PR #451, opened 2021-11-23
Do not clear refs on recursive type references

peterbroadhurst:
Proposed UT and fix for #447 

A reference was being added carefully in the recursion checker while building the struct on this line (note this _also_ includes the `schema` being passed into the `value` of the `$ref` - we leave both):
https://github.com/getkin/kin-openapi/blob/0db2d4c3583875f4fe65b47ccb51662a20ed32a9/openapi3gen/openapi3gen.go#L344

However, it was then being stripped out in this code:
https://github.com/getkin/kin-openapi/blob/0db2d4c3583875f4fe65b47ccb51662a20ed32a9/openapi3gen/openapi3gen.go#L58-L61

As the original `$ref.value` was still there (see note above), this meant the infinite loop occurred on JSON serialization.

### DIFF
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -57,7 +57,9 @@ func NewSchemaRefForValue(value interface{}, opts ...Option) (*openapi3.SchemaRe
 	g := NewGenerator(opts...)
 	ref, err := g.GenerateSchemaRef(reflect.TypeOf(value))
 	for ref := range g.SchemaRefs {
-		ref.Ref = ""
+		if !strings.HasPrefix(ref.Ref, "#/") {
+			ref.Ref = ""
+		}
 	}
 	return ref, g.SchemaRefs, err
 }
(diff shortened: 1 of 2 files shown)


peterbroadhurst:
Converting to draft while I work with @nguyer to verify this in the overall context of the Swagger generation.

fenollp:
Hi! Thanks for the patch. Before reviewing I'd like to note that this lib is v0 so there should not be much care given to backwards compatibility especially in this very case where the default behavior is troublesome - does not bode well with recursive refs.
Could you adapt your patch to actually replace the previous implementation namely `NewSchemaRefForValue`?

peterbroadhurst:
I've made the update to replace `NewSchemaRefForValue`, which did affect some tests so:

- I've updated tests that wanted to inspect internal details, to use `g := NewGenerator()` directly and call the internal implementation method
- As there was an assertion in the `simple_test.go` that could no longer be there, and that I believe is really more of a sample, I added a test that verifies the sample behavior.

nguyer:
Thanks @peterbroadhurst! I just tested this branch and these changes fix the issue I described in https://github.com/getkin/kin-openapi/issues/447

fenollp:
Superseeded by https://github.com/getkin/kin-openapi/pull/454
Please open an issue / a PR if there's anything you want working differently.

PR #462, opened 2021-12-13
Add openapi3 validator middleware

cmars:
This change introduces middleware which wraps an `http.Handler` with OpenAPI 3 request and response validation.

In developing this middleware, I thought it would benefit others trying to do the same. This is my first time contributing here, happy to answer any questions and discuss further.

Thank you for such an awesome OpenAPI library by the way!

### DIFF
--- a/openapi3filter/middleware.go
+++ b/openapi3filter/middleware.go
@@ -1,0 +1,273 @@
+package openapi3filter
+
+import (
+	"bytes"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+
+	"github.com/getkin/kin-openapi/routers"
+)
+
+// Validator provides HTTP request and response validation middleware.
+type Validator struct {
+	router  routers.Router
+	errFunc ErrFunc
+	logFunc LogFunc
+	strict  bool
+}
+
+// ErrFunc handles errors that may occur during validation.
+type ErrFunc func(w http.ResponseWriter, status int, code ErrCode, err error)
+
+// LogFunc handles log messages that may occur during validation.
+type LogFunc func(message string, err error)
+
+// ErrCode is used for classification of different types of errors that may
+// occur during validation. These may be used to write an appropriate response
+// in ErrFunc.
+type ErrCode int
+
+const (
+	// ErrCodeOK indicates no error. It is also the default value.
+	ErrCodeOK = 0
+	// ErrCodeCannotFindRoute happens when the validator fails to resolve the
+	// request to a defined OpenAPI route.
+	ErrCodeCannotFindRoute = iota
+	// ErrCodeRequestInvalid happens when the inbound request does not conform
+	// to the OpenAPI 3 specification.
+	ErrCodeRequestInvalid = iota
+	// ErrCodeResponseInvalid happens when the wrapped handler response does
+	// not conform to the OpenAPI 3 specification.
+	ErrCodeResponseInvalid = iota
+)
+
+func (e ErrCode) responseText() string {
+	switch e {
+	case ErrCodeOK:
+		return "OK"
+	case ErrCodeCannotFindRoute:
+		return "not found"
+	case ErrCodeRequestInvalid:
+		return "bad request"
+	default:
+		return "server error"
+	}
+}
+
+// NewValidator returns a new response validation middlware, using the given
+// routes from an OpenAPI 3 specification.
+func NewValidator(router routers.Router, options ...ValidatorOption) *Validator {
+	v := &Validator{
+		router: router,
+		errFunc: func(w http.ResponseWriter, status int, code ErrCode, _ error) {
+			http.Error(w, code.responseText(), status)
+		},
+		logFunc: func(message string, err error) {
+			log.Printf("%s: %v", message, err)
+		},
+	}
+	for i := range options {
+		options[i](v)
+	}
+	return v
+}
+
+// ValidatorOption defines an option that may be specified when creating a
+// Validator.
+type ValidatorOption func(*Validator)
+
+// OnErr provides a callback that handles writing an HTTP response on a
+// validation error. This allows customization of error responses without
+// prescribing a particular form. This callback is only called on response
+// validator errors in Strict mode.
+func OnErr(f ErrFunc) ValidatorOption {
+	return func(v *Validator) {
+		v.errFunc = f
+	}
+}
+
+// OnLog provides a callback that handles logging in the Validator. This allows
+// the validator to integrate with a services' existing logging system without
+// prescribing a particular one.
+func OnLog(f LogFunc) ValidatorOption {
+	return func(v *Validator) {
+		v.logFunc = f
+	}
+}
+
+// Strict, if set, causes an internal server error to be sent if the wrapped
+// handler response fails response validation. If not set, the response is sent
+// and the error is only logged.
+func Strict(strict bool) ValidatorOption {
+	return func(v *Validator) {
+		v.strict = strict
+	}
+}
+
+// Middleware returns an http.Handler which wraps the given handler with
+// request and response validation.
+func (v *Validator) Middleware(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		route, pathParams, err := v.router.FindRoute(r)
+		if err != nil {
+			v.logFunc("validation error: failed to find route for "+r.URL.String(), err)
+			v.errFunc(w, http.StatusNotFound, ErrCodeCannotFindRoute, err)
+			return
+		}
+		requestValidationInput := &RequestValidationInput{
+			Request:    r,
+			PathParams: pathParams,
+			Route:      route,
+		}
+		if err = ValidateRequest(r.Context(), requestValidationInput); err != nil {
+			v.logFunc("invalid request", err)
+			v.errFunc(w, http.StatusBadRequest, ErrCodeRequestInvalid, err)
+			return
+		}
+
+		var wr responseWrapper
+		if v.strict {
+			wr = &strictResponseWrapper{w: w}
+		} else {
+			wr = newWarnResponseWrapper(w)
+		}
+
+		h.ServeHTTP(wr, r)
+
+		if err = ValidateResponse(r.Context(), &ResponseValidationInput{
+			RequestValidationInput: requestValidationInput,
+			Status:                 wr.statusCode(),
+			Header:                 wr.Header(),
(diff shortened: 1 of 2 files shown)


openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+			v.errFunc(w, http.StatusNotFound, ErrCodeCannotFindRoute, err)
+			return
+		}
+		requestValidationInput := &RequestValidationInput{
+			Request:    r,
+			PathParams: pathParams,
+			Route:      route,
+		}
+		err = ValidateRequest(r.Context(), requestValidationInput)
+		if err != nil {
```
fenollp, line 124:
```suggestion
		if err = ValidateRequest(r.Context(), requestValidationInput); err != nil {
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+
+		h.ServeHTTP(wr, r)
+
+		err = ValidateResponse(r.Context(), &ResponseValidationInput{
+			RequestValidationInput: requestValidationInput,
+			Status:                 wr.statusCode(),
+			Header:                 wr.Header(),
+			Body:                   nopCloser{bytes.NewBuffer(wr.bodyContents())},
+		})
+		if err != nil {
```
fenollp, line 145:
```suggestion
		if err = ValidateResponse(r.Context(), &ResponseValidationInput{
			RequestValidationInput: requestValidationInput,
			Status:                 wr.statusCode(),
			Header:                 wr.Header(),
			Body:                   nopCloser{bytes.NewBuffer(wr.bodyContents())},
		}); err != nil {
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+			v.logFunc("invalid response", err)
+			if v.strict {
+				v.errFunc(w, http.StatusInternalServerError, ErrCodeResponseInvalid, err)
+			}
+			return
+		}
+
+		err = wr.flushBodyContents()
+		if err != nil {
```
fenollp, line 154:
```suggestion
		if err = wr.flushBodyContents(); err != nil {
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	fl     http.Flusher
+	status int
+	body   bytes.Buffer
+}
+
+// Write implements http.ResponseWriter.
+func (l *warnResponseWrapper) Write(b []byte) (int, error) {
+	if l.status == 0 {
+		l.w.WriteHeader(l.status)
+	}
```
fenollp, line 194: Shouldn't this be `!=` instead of `==`?
From https://pkg.go.dev/net/http#ResponseWriter
```go
	// If WriteHeader is not called explicitly, the first call to Write
	// will trigger an implicit WriteHeader(http.StatusOK).
	// Thus explicit calls to WriteHeader are mainly used to
	// send error codes.
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	n, err := l.w.Write(b)
+	if err == nil {
```
fenollp, line 196: Use `!= nil` for readability. I had to double take here.

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+
+		err = wr.flushBodyContents()
+		if err != nil {
+			v.logFunc("failed to write response", err)
+		}
+	})
+}
+
+type nopCloser struct {
+	io.Reader
```
fenollp, line 161: Why not use https://pkg.go.dev/io#ReadCloser directly?

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Write implements http.ResponseWriter.
+func (l *warnResponseWrapper) Write(b []byte) (int, error) {
+	if l.status == 0 {
+		l.w.WriteHeader(l.status)
+	}
+	n, err := l.w.Write(b)
+	if err == nil {
+		l.body.Write(b)
```
fenollp, line 197: WRT `l.body.Write(b)` why not use https://pkg.go.dev/io#MultiWriter

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	return n, err
+}
+
+// WriteHeader implements http.ResponseWriter.
+func (l *warnResponseWrapper) WriteHeader(status int) {
+	if l.status == 0 {
+		l.status = status
+	}
```
fenollp, line 206: Why compare to the empty value when `status` is here?
Reading this makes me think that zero is some special case and `l.status` may have been written to beforehand and so its value should only be overwritten in that case... but it's not clear to me what can mutate `l.status` appart from this place.

This surely alludes to my comment on 
```go
func (l *warnResponseWrapper) Write(b []byte) (int, error) {
	if l.status == 0 {
		l.w.WriteHeader(l.status)
	}
```

In any case, I think you should keep assigning here but drop the conditional; remove the call to `WriteHeader` in `Write`

openapi3filter/middleware_test.go
```diff
@@ -0,0 +1,384 @@
...
+		},
+		response: testResponse{
+			statusCode: 201,
+			body:       `{"id": "42", "contents": {"name": "foo", "expected": 9, "actual": 10}, "extra": true}`,
+		},
+		strict: false,
+	}}
+	for i, test := range tests {
+		t.Logf("test#%d: %s", i, test.name)
+		func() {
```
fenollp, line 329: Use https://pkg.go.dev/testing#T.Run

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	// statusCode returns the response status code, 0 if not set yet.
+	statusCode() int
+
+	// bodyContents returns the buffered
+	bodyContents() []byte
+}
+
+type warnResponseWrapper struct {
+	w      http.ResponseWriter
+	fl     http.Flusher
```
fenollp, line 185: Unused?

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	w      http.ResponseWriter
+	status int
+	body   bytes.Buffer
+}
+
+// Write implements http.ResponseWriter.
+func (wr *strictResponseWrapper) Write(b []byte) (int, error) {
+	if wr.status == 0 {
+		wr.status = http.StatusOK
+	}
```
fenollp, line 244: As per above:
```suggestion
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+	return wr.body.Write(b)
+}
+
+// WriteHeader implements http.ResponseWriter.
+func (wr *strictResponseWrapper) WriteHeader(status int) {
+	if wr.status == 0 {
+		wr.status = status
+	}
```
fenollp, line 252: As per above
```suggestion
	wr.status = status
```

openapi3filter/middleware.go
```diff
@@ -0,0 +1,272 @@
...
+}
+
+// Header implements http.ResponseWriter.
+func (wr *strictResponseWrapper) Header() http.Header {
+	return wr.w.Header()
+}
+
+func (wr *strictResponseWrapper) flushBodyContents() error {
+	wr.w.WriteHeader(wr.status)
```
fenollp, line 261: As per above 
```suggestion
```

openapi3filter/middleware_test.go
```diff
@@ -0,0 +1,384 @@
...
+	"net/http/httptest"
+	"regexp"
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers/gorillamux"
+)
+
+// TODO: simple OpenAPI spec
```
fenollp, line 18: Drop this dead comment.

openapi3filter/middleware_test.go
```diff
@@ -0,0 +1,384 @@
...
+	default:
+		http.Error(w, h.errBody, http.StatusMethodNotAllowed)
+	}
+}
+
+func TestValidator(t *testing.T) {
+	doc, err := openapi3.NewLoader().LoadFromData([]byte(validatorSpec))
+	if err != nil {
+		t.Error("failed to load test fixture spec:", err)
+	}
```
fenollp, line 166: Please use `require.NoError(t, err)` here on out, it reduces the line count.

openapi3filter/middleware_test.go
```diff
@@ -0,0 +1,384 @@
...
+				t.Error("failed to read response body:", err)
+			}
+			if test.response.body != string(body) {
+				t.Errorf("response body expect %q got %q", test.response.body, string(body))
+			}
+		}()
+	}
+}
+
+// TODO: example
```
fenollp, line 384: Yes please :)
Something extra simple so people can copy/paste and build on.

fenollp (changes requested):
Great work! I thought I already reviewed this but either it got lost to the ether or I'm going mad. Or you know, both. Or neither...

openapi3filter/middleware.go
cmars: Updated this behavior here and in the "strict-mode" wrapper below. We need to intercept and keep track of the status code for response validation, but using a separate flag to indicate whether the header has been written is much clearer.

openapi3filter/middleware.go
cmars: I'd forgotten about `ioutil.NopCloser` -- updated. Thought it was only available >=1.16 but that's in the `io` package.

openapi3filter/middleware.go
cmars: I think separating the status we record, from the state of whether we've "written the header" (or captured it) should hopefully clear this up, updated.

openapi3filter/middleware_test.go
cmars: :100: On it!

cmars: Added

openapi3filter/middleware.go
cmars: Not sure if this was clear (and if godoc comments could explain it better, please let me know!): we don't want to flush the response through to the underlying ResponseWriter in strict mode, because the response may be out of compliance with OpenAPI and invalid; we'll want to `wr.w.WriteHeader(500)` (or give our errFunc the chance to) when that happens in strict mode. If we pass through the call to WriteHeader, we won't be able to override it.

cmars:
@fenollp Thanks for the review! I believe I've addressed comments, open to discuss further. PTAL :pray:

fenollp (approved):

PR #463, opened 2021-12-14
Update rfc422 regex as per spec: 'case insensitive on input'

karl-dau:
From https://datatracker.ietf.org/doc/html/rfc4122 section 3:

>       Each field is treated as an integer and has its value printed as a
>       zero-filled hexadecimal digit string with the most significant
>       digit first.  The hexadecimal values "a" through "f" are output as
>       lower case characters and are **case insensitive on input.**

And the ABNF:

```
hexDigit =
            "0" / "1" / "2" / "3" / "4" / "5" / "6" / "7" / "8" / "9" /
            "a" / "b" / "c" / "d" / "e" / "f" /
            "A" / "B" / "C" / "D" / "E" / "F"
```

This pull request just allows the capitalised A-F.


fenollp:
Please add a test showing both casing cases.

karl-dau:
I added some test cases for all caps and mixed case input.

PR #469, opened 2021-12-21
feat: cache resolved refs, improve URI reader extensibility

### LINKED ISSUE #334: Resolving refs is not cached
From what I can see in swagger_loader.go, if a document has 10 references to the same document, the document will be fetched and parsed 10 times. If we were to cache by URL, we could significantly speed up document loading.

cmars:
This introduces simple URI-based caching.

A basic map-based cache implementation is provided, which may cover many
simple use cases. This caching is now used by default in the Loader.

File and HTTP URI readers are extracted from the loader internals so
that they may be extended, replaced, or configured. ReadFromHTTP allows
specifying the http.Client used to read HTTP $refs, for example.

Supersedes #117 and fixes #334.

### DIFF
--- a/openapi3/loader_uri_reader.go
+++ b/openapi3/loader_uri_reader.go
@@ -1,0 +1,104 @@
+package openapi3
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path/filepath"
+)
+
+// ReadFromURIFunc defines a function which reads the contents of a resource
+// located at a URI.
+type ReadFromURIFunc func(loader *Loader, url *url.URL) ([]byte, error)
+
+// ErrURINotSupported indicates the ReadFromURIFunc does not know how to handle a
+// given URI.
+var ErrURINotSupported = errors.New("unsupported URI")
+
+// ReadFromURIs returns a ReadFromURIFunc which tries to read a URI using the
+// given reader functions, in the same order. If a reader function does not
+// support the URI and returns ErrURINotSupported, the next function is checked
+// until a match is found, or the URI is not supported by any.
+func ReadFromURIs(readers ...ReadFromURIFunc) ReadFromURIFunc {
+	return func(loader *Loader, url *url.URL) ([]byte, error) {
+		for i := range readers {
+			buf, err := readers[i](loader, url)
+			if err == ErrURINotSupported {
+				continue
+			} else if err != nil {
+				return nil, err
+			}
+			return buf, nil
+		}
+		return nil, ErrURINotSupported
+	}
+}
+
+// DefaultReadFromURI returns a caching ReadFromURIFunc which can read remote
+// HTTP URIs and local file URIs.
+var DefaultReadFromURI = URIMapCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))
+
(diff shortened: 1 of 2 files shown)


openapi3/loader_uri_reader.go
```diff
@@ -0,0 +1,115 @@
...
+	"net/url"
+)
+
+// ReadFromURIFunc defines a function which reads the contents of a resource
+// located at a URI.
+type ReadFromURIFunc func(loader *Loader, url *url.URL) ([]byte, error)
+
+// ErrURINotSupported indicates the ReadFromURIFunc does not know how to handle a
+// given URI.
+var ErrURINotSupported = fmt.Errorf("unsupported URI")
```
fenollp, line 16: use errors.New

openapi3/loader_uri_reader.go
```diff
@@ -0,0 +1,115 @@
...
+			}
+			return buf, nil
+		}
+		return nil, ErrURINotSupported
+	}
+}
+
+// DefaultReadFromURI returns a ReadFromURIFunc which can read remote HTTP URIs and
+// local file URIs.
+var DefaultReadFromURI = ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile)
```
fenollp, line 39: I'd like caching to be enabled by default.

openapi3/loader_uri_reader.go
```diff
@@ -0,0 +1,115 @@
...
+		return nil, ErrURINotSupported
+	}
+	if location.Scheme != "" && location.Scheme != "file" {
+		return nil, ErrURINotSupported
+	}
+	return ioutil.ReadFile(location.Path)
+}
+
+// URIReadCache defines a cache for the contents read from URI references.
+type URIReadCache interface {
```
fenollp, line 77: What does this provide that `ReadFromURIFunc` cannot already do?
A simple cache:
```go
var DefaultReadFromURI = SimplestCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))

var readURICache map[string][]byte
func SimplestCache(reader ReadFromURIFunc) ReadFromURIFunc {
	if readURICache == nil {
		readURICache = make(map[string][]byte)
	}
	return func(loader *Loader, url *url.URL) (buf []byte, err error) {
		uri := url.String()
		var ok bool
		if buf, ok = readURICache[uri]; ok {
			return
		}
		if buf, err = reader(loader, url); err != nil {
			return
		}
		readURICache[uri] = buf
		return
	}
}
```

fenollp (changes requested):
I like these changes however I don't see how `URIReadCache` is needed when there is already `ReadFromURIFunc`.

openapi3/loader_uri_reader.go
cmars: Good point, the interface was overkill, walking that back.

cmars:
Gentle reminder on this one, could I please get another review?

I'm seeing a significant speedup in a [downstream tool of mine](https://github.com/snyk/vervet) with this change, when it processes OpenAPI that reuses a lot of http $ref'd components.

Without caching:
```
real 61.28
user 3.75
sys 2.75
```

With caching:
```
real 0.89
user 0.39
sys 0.06
```

Thanks!

fenollp (approved):

PR #470, opened 2022-01-04
extensible-paths

general-kroll-4-life:
## Summary

The existing path sorting approach is flawed, because calls to `Router.FindRoute()` invariably lead to `routers.ErrMethodNotAllowed` where both `/{X}` and `/{X}:<someSuffix>` are present.  This occurs because `gorillamux` does not preclude `:` inside parameter substitutions and the shorter templated path is evaluated first.  The suggested fix is simply to reverse the lexicographical ordering.  Thus, the longer path is evaluated first.

### Note on unit tests

The prior unit tests leverage lexicographical sorting on the parameter name itself, coupled with opportune parameter name selection, to ensure passing tests.  However, according to https://swagger.io/specification/#patterned-fields:

> Templated paths with the same hierarchy but different templated names MUST NOT exist as they are identical.

This is also exemplified in https://swagger.io/specification/#path-templating-matching 

This implies we should not leverage parameter names in this way.

While I can well imagine future support for "same hierarchy but different templated names" (eg: parameter schema or format different), then in such cases the paths should be evaluated in an ordering that guarantees the core problem does not occur.

### DIFF
--- a/routers/gorillamux/router.go
+++ b/routers/gorillamux/router.go
@@ -142,7 +142,7 @@ func orderedPaths(paths map[string]*openapi3.PathItem) []string {
 	ordered := make([]string, 0, len(paths))
 	for c := 0; c <= max; c++ {
 		if ps, ok := vars[c]; ok {
-			sort.Strings(ps)
+			sort.Sort(sort.Reverse(sort.StringSlice(ps)))
 			ordered = append(ordered, ps...)
 		}
 	}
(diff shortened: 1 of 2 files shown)


fenollp:
Could you rephrase the issue you're solving here? I'm not sure I understand what `X:action` is in this context.
Also I have trouble interpreting the tests. Why rename the `bookid2` parameter and how did that fix the tests?
Could you add regression test(s) around the specific issue this is solving?

Please also amend the note in that function: https://github.com/getkin/kin-openapi/blob/06620d24bbe9ccdfdb4d3925f0e3727030ee8f42/routers/gorillamux/router.go#L131-L132

PR #478, opened 2022-01-21
Fix slash removing in `documentPathForRecursiveRef` method

bozaro:
Unfortunately `path.Join` remove last slash from result file.

For example:
 ```
path.Join("foo/bar/", "") == "foo/bar"
```

As result `documentPathForRecursiveRef(current, "")` returns `rootDir` without directory as file path and next relative path lost last `rootDir` element.

In this workaround I try to restore last slash if needed.


fenollp (changes requested):
Please add regression tests.

PR #483, opened 2022-02-01
openapi2conv: Convert response headers

SVilgelm:
v2 <-> v3 conversion of the Response.Headers objects

### DIFF
--- a/openapi2conv/openapi2_conv.go
+++ b/openapi2conv/openapi2_conv.go
@@ -424,7 +424,31 @@ func ToV3Response(response *openapi2.Response) (*openapi3.ResponseRef, error) {
 	if schemaRef := response.Schema; schemaRef != nil {
 		result.WithJSONSchemaRef(ToV3SchemaRef(schemaRef))
 	}
+	if headers := response.Headers; len(headers) > 0 {
+		result.Headers = ToV3Headers(headers)
+	}
 	return &openapi3.ResponseRef{Value: result}, nil
+}
+
+func ToV3Headers(defs map[string]*openapi2.Header) openapi3.Headers {
+	headers := make(openapi3.Headers)
+	for name, header := range defs {
+		stripNonCustomExtensions(header.Extensions)
+		if ref := header.Ref; ref != "" {
+			headers[name] = &openapi3.HeaderRef{Ref: ToV3Ref(ref)}
+		} else {
+			headers[name] = &openapi3.HeaderRef{Value: &openapi3.Header{
+				Parameter: openapi3.Parameter{
+					ExtensionProps: header.ExtensionProps,
+					Description:    header.Description,
+					Schema: &openapi3.SchemaRef{Value: &openapi3.Schema{
+						Type: header.Type,
+					}},
+				},
+			}}
+		}
+	}
+	return headers
 }
 
 func ToV3Schemas(defs map[string]*openapi3.SchemaRef) map[string]*openapi3.SchemaRef {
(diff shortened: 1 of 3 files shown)


openapi2conv/openapi2_conv.go
```diff
@@ -424,9 +424,33 @@ func ToV3Response(response *openapi2.Response) (*openapi3.ResponseRef, error) {
...
+		stripNonCustomExtensions(header.Extensions)
+		if ref := header.Ref; ref != "" {
+			headers[name] = &openapi3.HeaderRef{Ref: ToV3Ref(ref)}
+		} else {
+			headers[name] = &openapi3.HeaderRef{Value: &openapi3.Header{
+				Parameter: openapi3.Parameter{
+					ExtensionProps: header.ExtensionProps,
+					Description:    header.Description,
+					Schema: &openapi3.SchemaRef{Value: &openapi3.Schema{
+						Type: header.Type,
```
fenollp, line 445: Why only translate type?

SVilgelm: because the i is only `type` parameter in the v2 struct: https://github.com/getkin/kin-openapi/blob/master/openapi2/openapi2.go#L234-L239

We can change the v2 Header struct to use Parameter as it is done in v3

fenollp: I see. This is missing plenty: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#:~:text=If%20in%20is%20any%20value%20other%20than%20%22body%22%3A 

If you don't feel like adding the missing fields in this PR it can always be done in a follow up PR.

SVilgelm: I will add in this PR, np

SVilgelm: ok, I adde second commit here with using the Parameter for Header struct

openapi2conv/openapi2_conv.go
```diff
@@ -424,9 +424,29 @@ func ToV3Response(response *openapi2.Response) (*openapi3.ResponseRef, error) {
...
 		result.WithJSONSchemaRef(ToV3SchemaRef(schemaRef))
 	}
+	if headers := response.Headers; len(headers) > 0 {
+		result.Headers = ToV3Headers(headers)
+	}
 	return &openapi3.ResponseRef{Value: result}, nil
 }
 
+func ToV3Headers(defs map[string]*openapi2.Header) openapi3.Headers {
+	headers := make(openapi3.Headers)
```
fenollp, line 434: Please reserve space `len(defs)`

openapi2conv/openapi2_conv.go
```diff
@@ -1048,9 +1069,31 @@ func FromV3Response(ref *openapi3.ResponseRef, components *openapi3.Components)
...
+		var err error
+		if result.Headers, err = FromV3Headers(headers, components); err != nil {
+			return nil, err
+		}
+	}
 	return result, nil
 }
 
+func FromV3Headers(defs openapi3.Headers, components *openapi3.Components) (map[string]*openapi2.Header, error) {
+	headers := make(map[string]*openapi2.Header)
```
fenollp, line 1082: Reserve space.

fenollp (changes requested):
A couple notes WRT memory allocation, otherwise LGTM.

openapi2conv/openapi2_conv.go
SVilgelm: sure

SVilgelm: done

openapi2conv/openapi2_conv.go
SVilgelm: done

fenollp (approved):

SVilgelm:
cool, thank you so much, and especial for releasing new version!

fenollp:
Of course :) Thanks for the patch!

PR #318, opened 2021-03-03
WIP: Switch to gojsonschema

fenollp:
Fixes incorrect schema validation:
* [x] https://github.com/getkin/kin-openapi/issues/82
* [x] https://github.com/getkin/kin-openapi/issues/171
* [x] https://github.com/getkin/kin-openapi/issues/203
* [x] https://github.com/getkin/kin-openapi/issues/267
* [ ] https://github.com/getkin/kin-openapi/issues/280
* [ ] https://github.com/getkin/kin-openapi/issues/293

Changes `format` support:
* [x] https://github.com/getkin/kin-openapi/issues/107

Also:
* works [towards OpenAPIv3.1](https://github.com/getkin/kin-openapi/issues/230)

PR #487, opened 2022-02-11
WIP: Use santhosh-tekuri/jsonschema

lkm:
This is based of this PR https://github.com/getkin/kin-openapi/pull/318 to remove internal validation and originally replace with [xeipuuv/gojsonschema](https://github.com/xeipuuv/gojsonschema) and then in the newest commits replace with [santhosh-tekuri/jsonschema](https://github.com/santhosh-tekuri/jsonschema)

I'm using an extension ([openapi3/components_extension.go](https://github.com/getkin/kin-openapi/compare/master...lkm:jsonschema?expand=1#diff-5940f5499a122afd5ad162529ce7e45f974d02add4554d49405ebfe0ce381649) to validate all `components.schemas` in one go as this seems to be the only way to keep internal document referencing working, e.g. `#/components.schemas.ExampleObject`. 

This is very much still a work in progress and I'd need to resolve the fact the for this to work fully within the openapi3filter we'll need the complete `openapi3.T` document to resolve request and response references too. However, that very much seems to go against the grain of the current design.

I could use some input or opinions on direction. @fenollp


lkm:
Another idea would be to use the OAS json schema to validate the entire document, see https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.json - however, that may be too much tie-in and too big of a diversion as we'd have to write a few extensions to handle the different aspects.

fenollp:
Superseeded by #1125
Please hang in tight while that PR gets merged :)

PR #489, opened 2022-02-15
WIP: Skip adding index in error for enum validation

kandaaaaa:
Enum value validation has confused error messages like below where `Error at "0"` is like hardcoded because it is validating with enum values, which is unnecessary and confusing. So this PR suggests removing this. error will be `value is not one of the allowed values`
```
Error at \"/0\": value is not one of the allowed values
```


kandaaaaa:
I think I misunderstand the original reason for the index... Closing

PR #490, opened 2022-02-16
Fix openapi3 validation: path param must be required

vasayxtx:
Hi,

Looks like path parameter always should be marked as `required: true` explicitly according OpenAPI 3.0 spec.
Quote from the [section](https://spec.openapis.org/oas/v3.0.3#fixed-fields-9) about `required` field of the parameter object:
> Determines whether this parameter is mandatory. If the [parameter location](https://spec.openapis.org/oas/v3.0.3#parameterIn) is "path", this property is REQUIRED and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

So, perhaps it would be better to check this on validation.

### DIFF
--- a/openapi3/loader_test.go
+++ b/openapi3/loader_test.go
@@ -412,6 +412,7 @@ paths:
       parameters:
         - name: id,
           in: path
+          required: true
           schema:
             type: string
       responses:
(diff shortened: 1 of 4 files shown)


fenollp:
Good catch. Thanks!

PR #491, opened 2022-02-16
Fix oauth2 in openapi2conv.FromV3SecurityScheme

ckaznocha:
This change addresses two issues.
* When scopes are present in an openapi3.OAuthFlow the conversion panics
  due to an npe because the result.Scopes map had not been initialized.
* The conversion of the flow is incomplete because each oauth2 flow
  type has different required fields. The requirements for each type can
  be found here https://github.com/OAI/OpenAPI-Specification/blob/main/versions/2.0.md#securityDefinitionsObject


fenollp:
Thanks!

PR #496, opened 2022-02-24
added negative test case for validating discriminator

ohasanliUW:
Request body validation does not work correctly when `discriminator` and `oneOf` is used. This PR is adding a test case that expects validation to fail but it passes.


ohasanliUW:
perhaps this test case should be added to `func TestValidateRequestBody(t *testing.T)` ?

PR #498, opened 2022-02-26
Fix OpenAPI 3 validation: request body content is required

vasayxtx:
According to [request body object  spec section](https://spec.openapis.org/oas/v3.0.3#request-body-object) content field is required. Seems like it would be better to validate this.

P.S.: we use this cool library for strict OpenAPI v3 specs validation in one of our project, that's why I dive into this:)


fenollp:
Well thank you very much for diving into this, I'm glad!

PR #501, opened 2022-03-01
issue/500

### LINKED ISSUE #500: openapi3gen.NewSchemaRefForValue panics for embedded pointer types
**Summary**

`openapi3gen.NewSchemaRefForValue` panics if it encounters an embedded pointer type when a schema customizer is being used.

See [here](https://go.dev/play/p/A4gXt0Hk1m9) for a minimal example that reproduces the problem.

**Explanation**

I believe the panic is originating [here](https://github.com/getkin/kin-openapi/blob/5352767564ece15c85c3d0a1ea1143d1f4206b04/openapi3gen/openapi3gen.go#L134) in `getStructField`, because `t.Field` panics if the `reflect.Type`'s `Kind` is not `Struct` (and in this case, it's `Ptr`). The panic only occurs when using a schema customizer because `getStructField` is only called (in this case) when a schema customizer is present (see [here](https://github.com/getkin/kin-openapi/blob/5352767564ece15c85c3d0a1ea1143d1f4206b04/openapi3gen/openapi3gen.go#L312-L313)).

I believe this could be addressed by doing something like [this](https://github.com/getkin/kin-openapi/blob/5352767564ece15c85c3d0a1ea1143d1f4206b04/jsoninfo/field_info.go#L24-L26) in [getStructField](https://github.com/getkin/kin-openapi/blob/5352767564ece15c85c3d0a1ea1143d1f4206b04/openapi3gen/openapi3gen.go#L130).

Here's a pull request that implements the fix: https://github.com/getkin/kin-openapi/pull/501.

cochran-at-niche:
Fixes #500 by updating [getStructField](https://github.com/nicheinc/kin-openapi/blob/issue/500/openapi3gen/openapi3gen.go#L130-L141) to handle embedded pointer types, which were previously causing panics (but only when a schema customizer was being used). The solution involved adding these [three lines](https://github.com/nicheinc/kin-openapi/blob/issue/500/openapi3gen/openapi3gen.go#L136-L138), to mirror the logic happening [here](https://github.com/getkin/kin-openapi/blob/5352767564ece15c85c3d0a1ea1143d1f4206b04/jsoninfo/field_info.go#L24-L26).

Also includes a new unit test that should help prevent regressions (I verified that the unit test triggered the panic before these changes).

Please let me know if you have any questions!

### DIFF
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -133,6 +133,9 @@ func getStructField(t reflect.Type, fieldInfo jsoninfo.FieldInfo) reflect.Struct
 	for i := 0; i < len(fieldInfo.Index); i++ {
 		ff = t.Field(fieldInfo.Index[i])
 		t = ff.Type
+		if t.Kind() == reflect.Ptr {
+			t = t.Elem()
+		}
 	}
 	return ff
 }
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -133,7 +133,7 @@ func getStructField(t reflect.Type, fieldInfo jsoninfo.FieldInfo) reflect.Struct
 	for i := 0; i < len(fieldInfo.Index); i++ {
 		ff = t.Field(fieldInfo.Index[i])
 		t = ff.Type
-		if t.Kind() == reflect.Ptr {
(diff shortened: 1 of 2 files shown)


fenollp:
Stellar work. Thanks!

PR #509, opened 2022-03-13
discriminator value should verify the type is string to avoid panic

kandaaaaa:
With out this check, the request with body like `{"discriminator": null}` or `{"discriminator": 1}` will simply get panic 
```
interface conversion: interface {} is nil, not string
```

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -871,11 +871,12 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 					return errors.New("input does not contain the discriminator property")
 				}
 
-				if _, okcheck = discriminatorVal.(string); !okcheck {
+				discriminatorValString, okcheck := discriminatorVal.(string)
+				if !okcheck {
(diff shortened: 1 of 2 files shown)


openapi3/schema.go
```diff
@@ -871,6 +871,11 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 					return errors.New("input does not contain the discriminator property")
 				}
 
+				_, okcheck = discriminatorVal.(string)
+				if !okcheck {
```
fenollp, line 875:
```suggestion
				if _, okcheck = discriminatorVal.(string); !okcheck {
```

fenollp (changes requested):
Thanks for this!
Please add regression tests with at least the two example values you mention.

kandaaaaa:
Will add some regression tests soon

kandaaaaa:
@fenollp PR updated for a new test case, and minor code update. Thanks

Issue #416, opened 2021-09-03
Order of request validation


nmeheus:
I noticed that the validation of the security schemes is only the last action when validating a request (see: https://github.com/getkin/kin-openapi/blob/master/openapi3filter/validate_request.go#L84). While I haven't yet found a ground truth for this online, I was wondering if it wouldn't make more sense to check security schemes before the others, as one might not want to expose information about what the request should look like to unauthorized users, if one were to return the request validation errors in the response. If anyone has any sources documenting the intended behaviour, I would love to know!

fenollp:
Sure validating security schemes first makes sense. Feel free to send a PR (along with a test, please)!

PR #514, opened 2022-03-24
Change the order of request validation to validate the Security schemas first before all other paramters

Yarn-e:
## Problem statement

The security schemas are being checked after all other parameters are checked, this could possibly give a user who is not authorised to see validation errors that expose how the request works, instead of giving them an error that they are not authorised.

## What changed

- Moved the security check to be before all the other parameters
- Added tests for ValidateRequest to show that this solution fixes the issue

## Related issue
- Fixes https://github.com/getkin/kin-openapi/issues/416


fenollp:
Thanks! Good work!

Issue #389, opened 2021-07-29
Non required parameters with empty values are not caught


nmeheus:
According to the specifications an empty valued parameter is not equal to the parameter not being present. For required parameters one can use the `allowEmptyValue: true` to allow `?queryparam=`, but for non required query parameters there seems no way to prevent `?queryparam=` and `?queryparam`(*) from passing the validation. 

A small snippet to illustrate the expected behaviour, and how it fails with the current implementation: https://goplay.space/#T58uhnOPA-t

(*) Note that this behaviour (without the `=`) is also not working as expected for required parameters.

fenollp:
Thanks for the demo code! Amazing!

So yes there is no handling of `allowEmptyValue` appart from un/marshaling. PRs very welcomed!

PR #515, opened 2022-03-24
Add support for allowEmptyValue

Yarn-e:
## Problem statement

The library currently does not support AllowEmptyValue to be validated correctly, this PR adds the support for it.
This also fixes the issue that when an empty valued parameter is being send that the library will treat it as a valid parameter. It will now check if theres an actual value present.

For example, this was possible, which should not be allowed
```
my.cool.host/api/test/?queryParam=
```

## What changed

- Added support for AllowEmptyValue
- Checks if an actual value is send when the parameter is present

## Related issue
- Fixes https://github.com/getkin/kin-openapi/issues/389


openapi3filter/internal.go
```diff
@@ -11,3 +12,14 @@ func parseMediaType(contentType string) string {
 	}
 	return contentType[:i]
 }
+
+func IsNilValue(value interface{}) bool {
```
fenollp, line 16: Why make this public?

fenollp (changes requested):

openapi3filter/internal.go
Yarn-e: It should not be public 🙈.

fenollp (approved):

PR #516, opened 2022-03-28
Fix ExampleValidator test for 32-bit architectures

anthonyfok:
Fix an integer overflow in ExampleValidator on 32-bit architectures (where `65536 * 65536` with `int` returns `0`)
by using int64 for square calculations.

Tested with `GOARCH=386 go test ./...`

See Debian Bug report log "autopkgtest regression on armhf and i386" at https://bugs.debian.org/1007733

Many thanks!

### DIFF
--- a/openapi3filter/middleware_test.go
+++ b/openapi3filter/middleware_test.go
@@ -420,7 +420,7 @@ paths:
 	// requests.
 	squareHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		xParam := path.Base(r.URL.Path)
-		x, err := strconv.Atoi(xParam)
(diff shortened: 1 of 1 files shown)


anthonyfok:
> Thanks for upstreaming this!

You are very welcome!  It is our duty as Debian developers to do so!

> Also I had no idea Debian packaged golang libs

We Debian (and Fedora) too package Go libraries usually when they are needed as dependencies for Go programs.  For example, github.com/getkin/kin-openapi is needed to build [Hugo](https://github.com/gohugoio/hugo), so packaging github.com/getkin/kin-openapi (as golang-github-getkin-kin-openapi) is a prerequisite.  :-)

Thank you for creating and maintaining github.com/getkin/kin-openapi!

PR #522, opened 2022-03-31
Add ExcludeSchema sentinel error for schemaCustomizer

nguyer:
This PR extends the functionality of the schemaCustomizer to allow the customizer to indicate that the schema for a specific field should actually be ignored in the final schema output.

This is useful for cases where might be building a schema for a struct that has internal fields which should not be exposed in the generated Open API spec. It can also be useful in cases where the Go code uses a single struct as input and output JSON, but some fields (say an `id` for example) are only present on the HTTP response, and should not be part of the schema for a request.

Using a sentinel error seemed to be the least intrusive change to the code to enable this, and also still maintain backwards compatibility by not changing the function signature, but I'm open to feedback.

### DIFF
--- a/openapi3gen/openapi3gen_test.go
+++ b/openapi3gen/openapi3gen_test.go
@@ -468,6 +468,17 @@ func TestSchemaCustomizerError(t *testing.T) {
 	require.EqualError(t, err, "test error")
 }
 
+func TestSchemaCustomizerExcludeSchema(t *testing.T) {
+	customizer := openapi3gen.SchemaCustomizer(func(name string, ft reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error {
+		return &openapi3gen.ExcludeSchemaSentinel{}
+	})
+
+	type Bla struct{}
+	schema, err := openapi3gen.NewSchemaRefForValue(&Bla{}, nil, openapi3gen.UseAllExportedFields(), customizer)
+	require.NoError(t, err)
+	require.Nil(t, schema)
+}
+
 func ExampleNewSchemaRefForValue_recursive() {
 	type RecursiveType struct {
 		Field1     string           `json:"field1"`
--- a/openapi3gen/openapi3gen_test.go
+++ b/openapi3gen/openapi3gen_test.go
@@ -469,12 +469,25 @@ func TestSchemaCustomizerError(t *testing.T) {
 }
 
 func TestSchemaCustomizerExcludeSchema(t *testing.T) {
+	type Bla struct {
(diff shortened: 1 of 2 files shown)


fenollp (changes requested):
LGTM please rename the error to `ExcludeSchemaSentinel` (also it has a typo)

nguyer:
> LGTM please rename the error to `ExcludeSchemaSentinel` (also it has a typo)

Thanks! 🤦‍♂️ not sure why spell check didn't catch that one...

Name has been updated

openapi3gen/openapi3gen_test.go
```diff
@@ -468,6 +468,17 @@ func TestSchemaCustomizerError(t *testing.T) {
...
 
+func TestSchemaCustomizerExcludeSchema(t *testing.T) {
+	customizer := openapi3gen.SchemaCustomizer(func(name string, ft reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error {
+		return &openapi3gen.ExcludeSchemaSentinel{}
+	})
+
+	type Bla struct{}
+	schema, err := openapi3gen.NewSchemaRefForValue(&Bla{}, nil, openapi3gen.UseAllExportedFields(), customizer)
+	require.NoError(t, err)
+	require.Nil(t, schema)
```
fenollp, line 479:
```suggestion
	type Bla struct {
		Str string
	}

	customizer := openapi3gen.SchemaCustomizer(func(name string, ft reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error {
		return nil
	})
	schema, err := openapi3gen.NewSchemaRefForValue(&Bla{}, nil, openapi3gen.UseAllExportedFields(), customizer)
	require.NoError(t, err)
	require.Equal(t, &openapi3.SchemaRef{Value: &openapi3.Schema{
		Type: "object",
		Properties: map[string]*openapi3.SchemaRef{
			"Str": {Value: &openapi3.Schema{Type: "string"}},
		}}}, schema)

	customizer = openapi3gen.SchemaCustomizer(func(name string, ft reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error {
		return &openapi3gen.ExcludeSchemaSentinel{}
	})
	schema, err = openapi3gen.NewSchemaRefForValue(&Bla{}, nil, openapi3gen.UseAllExportedFields(), customizer)
	require.NoError(t, err)
	require.Nil(t, schema)
```

fenollp (approved):

nguyer:
Thanks for merging this!

Issue #367, opened 2021-06-09
gorillamux.NewRouter fails if any servers url contain variable for port number of the url


karupanerura:
e.g.

```yaml
openapi: "3.0.0"
servers:
  - url: http://127.0.0.1:{port}/
    variables:
      port:
        default: '8000'
```

It throws error at https://github.com/getkin/kin-openapi/blob/v0.63.0/routers/gorillamux/router.go#L41

```
parse "http://127.0.0.1:--------------------------------------------------port__________________________________________________": invalid port ":--------------------------------------------------port__________________________________________________" after host
```

fenollp:
Ah yes this is related to https://github.com/getkin/kin-openapi/pull/337 and the issue it attempts to solve.
In my mind there isn't a perfect solution to this as it relies on `url.Parse`'s leniency. A good enough solution might be to use the default value of variables in the host but then your `port` will only ever match `8000`. Which is fine as long as it is documented I guess.

karupanerura:
Thank you for the kind reply. My understanding is almost exactly the same as yours.

In my opinion, `url.Parse` is only intended to validate the URL format. So, I think that creating a URL string using the default value of variables in the host and validating it will serve the purpose.
Next, `(*mux.Route).Host` method [can accept the OpenAPI 3.0 servers URL template syntax(by accident)](https://pkg.go.dev/github.com/gorilla/mux#Route.Host). So, I think it can pass the host section of the template directly to this method.
I think it can solve this issue.

If this idea is good for you, I can send you a patch. What do you think?

slessard:
Now I've hit this problem. @karupanerura I'd be interested in trying out a patch if you have one

fenollp:
I've opened https://github.com/getkin/kin-openapi/pull/524 that handles this specific issue.

@slessard Your PR (at this time) may solve this issue but it uses defaults for all variables not just this one specific case of port vars and so is not backwards compatible.

@karupanerura @slessard please give feedback
If you guys are happy with it I'll merge this and close https://github.com/getkin/kin-openapi/pull/523

slessard:
Sorry for the delay. I am looking at this now

PR #523, opened 2022-03-31
Add minimalistic support for resolving server variables

slessard:
each server variable is replaced with its default value.


fenollp:
Continuing discussion from https://github.com/getkin/kin-openapi/issues/367#issuecomment-1085106442 here:
```
--- FAIL: TestRouter (0.00s)
    router_test.go:93: 'GET https://domain0.domain1.com/api/v1/hello': should have returned an operation, but it returned an error: no matching operation was found
```
This is because with your patch 
```go
	doc.Servers = []*openapi3.Server{
		{URL: "https://www.example.com/api/v1"},
		{URL: "{scheme}://{d0}.{d1}.com/api/v1/", Variables: map[string]*openapi3.ServerVariable{
			"d0":     {Default: "www"},
			"d1":     {Default: "example", Enum: []string{"example"}},
			"scheme": {Default: "https", Enum: []string{"https", "http"}},
		}},
	}
```
`d0` and `d1` get set to their defaults and thus
```go
	expect(r, http.MethodGet, "https://domain0.domain1.com/api/v1/hello", helloGET, map[string]string{
		"d0": "domain0",
		"d1": "domain1",
	})
```
the given URL `"https://domain0.domain1.com/api/v1/hello"` cannot match that host.

slessard:
I've closed this pull request in favor of the solution that is being worked on in #524

PR #532, opened 2022-04-03
Don't validate readonly field for write requests

CTrando:
## What

Don't validate readonly field for write requests

## Why

In my work, we were using the same struct for many of our openapi endpoints, and one of those struct fields was `readOnly`. We were getting validation errors whenever we tried to make a post request with that struct because we weren't filling in the field.

For example, we were using the common struct:

```
Account:
  id: readonly
```
For all our endpoints.

When we tried to create an account, we wouldn't fill in the ID, but we were still getting validation errors. The issue is because in the library there's no way (to my knowledge) to give the validator context on whether the endpoint is a read endpoint or a write endpoint. The behavior should be if the endpoint is a write endpoint, then we shouldn't validate a readonly field.

cc @codyaray who knows more about this than me.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -547,10 +547,14 @@ func (schema *Schema) WithAdditionalProperties(v *Schema) *Schema {
 	return schema
 }
 
-func (schema *Schema) IsEmpty() bool {
+func (schema *Schema) IsEmpty(settings *schemaValidationSettings) bool {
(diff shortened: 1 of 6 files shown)


codyaray:
@CTrando can you share the exact error output you receive before this change? Just trying to make sure we have all the context here

fenollp:
@CTrando Kind greetings :) Is there anything you'd like to add to this PR / explain about the issue you're solving? Thanks!

Issue #521, opened 2022-03-30
Load yaml file for openapi2


ilteoood:
Hi guys,
thank you so much for this awesome library.

I'm opening this issue just to ask a question: is a way to load an openapi2 specification from a yaml file?

Currently I'm:
- loading it;
- converting it to JSON;
- using the `UnmarshalJSON` method to fill the openapi2.T object.

Is there a better way?
Because the convertion yaml 2 json is a little bit heavy for large specifications.

fenollp:
Have a look at https://github.com/ghodss/yaml

And if you have the time I'd happily review a pull request that solves this for others, be it only an example.

fenollp:
See https://github.com/getkin/kin-openapi/blob/0a4abfc91ea65332d247587bda355ab2ed461ef0/openapi2/openapi2_test.go#L45

PR #533, opened 2022-04-04
Unmarshal YAML

ilteoood:
Fix for #521

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,5 +7,6 @@ require (
 	github.com/go-openapi/jsonpointer v0.19.5
 	github.com/gorilla/mux v1.8.0
 	github.com/stretchr/testify v1.5.1
-	gopkg.in/yaml.v2 v2.3.0
+	gopkg.in/yaml.v2 v2.4.0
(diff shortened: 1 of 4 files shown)


fenollp (changes requested):
This lib uses https://github.com/ghodss/yaml already so bringing a fork of (k8s sigs) to me should replace ghodss's throughout the project and so be discussed in another PR (which I welcome, by the way).

Please try this again solely using https://github.com/ghodss/yaml for now.

Your tests seem to miss the fact that `UnmarshalJSON` / `UnmarshalYAML` implement a certain interface that is to be used with `(encoding/json).Unmarshal` and `(ghodss/yaml).Unmarshal`. Your calls should reflect this usage instead.
The test should also be turned into an Example so kin-openapi users of openapiv2 have a documented example of how to read YAML documents.

fenollp:
Closing this because we already can deserialize OpenAPIv2 documents formatted as YAML: https://github.com/getkin/kin-openapi/blob/0a4abfc91ea65332d247587bda355ab2ed461ef0/openapi2/openapi2_test.go#L45

PR #540, opened 2022-05-11
Internalize parameter references in the path as well

hikhvar:
We use externalized parameters in our OpenAPI spec. While running `InternalizeSpec` those were not internalized. This is a fix for that bug.


fenollp:
Thanks!

Issue #136, opened 2019-10-31
validation of default value conformity


bonnyr:
The OpenAPI Spec states that the default property, when applied to a schema must use the same values as the underlying property type, thus, a string property cannot have an integer value specified etc.

> https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schema-object
> default - The default value represents what would be assumed by the consumer of the input as the value of the schema if one is not provided. Unlike JSON Schema, the value MUST conform to the defined type for the Schema Object defined at the same level. For example, if type is string, then default can be "foo" but cannot be 1.

This is currently not covered by the project as best I can tell.

Is this on anyone's radar to fix? I'm currently working on a private version that tackles that for our use case, but will be happy to upstream this if needed

fenollp:
No idea if it's on anyone's radar. This project is about implementing the whole OA3 spec so I'd gladly review your PRs :)

PR #544, opened 2022-05-19
feat: handling `default` in request body and parameter schema

### LINKED ISSUE #206: Handling "default" in request body schema
(Apologies if I'm getting the nature of "default" completely wrong, but...)

I have an openapi3 request body schema which includes:

```json
          "group_by_time": {
            "default": true,
            "description": "Aggregate data over the provided time range, using the provided resolution. Default: true",
            "type": "boolean"
          },
```

For me, the ideal result of RequestBodyValidation would be a changed body that has `"group_by_time": true` if it is not otherwise set. I can think of many of reasons why this is a bad idea: req.Body -> JSON -> to new JSON -> to new Body is icky.

Also it doesn't appear that the json visitor would make the changes, as currently written.

So what is the correct or preferred way of achieving this? Should I simply be checking/setting the default in my code.

If there is a way to make this possible in the kin-openapi, I'm happy to work on the code if someone pushes me in the right direction.

nic-6443:
Based on #285
Close #206


nic-6443:
Hi, @fenollp Do you have time to review this PR, thanks.

fenollp:
Thank you very much! 
Closes #285 
Closes #136 
Closes #162

PR #556, opened 2022-06-01
Unit tests

slessard:
Add three new test cases for the TestServerPath test in routers/gorillamux/router_test.go
The last of the three test cases, use a variable for the port number, fails in v0.95.0 but passes in v0.96.0


fenollp:
Thanks!

PR #565, opened 2022-07-05
Match on overridden servers at the path level, fixes #564

### LINKED ISSUE #564: gorillamux router does not match on overridden servers
The gorillamux router does not match on overridden servers at the path and operation level.

These are valid in the OpenAPI spec: https://swagger.io/docs/specification/api-host-and-base-path/

sonu27:
This adds support for overriding servers at the path level.

Support for overriding servers at the operation level is still not supported.

### DIFF
--- a/routers/gorillamux/router.go
+++ b/routers/gorillamux/router.go
@@ -34,6 +34,13 @@ type routeMux struct {
 	varsUpdater varsf
 }
 
+type srv struct {
+	schemes     []string
+	host, base  string
+	server      *openapi3.Server
+	varsUpdater varsf
+}
+
 var singleVariableMatcher = regexp.MustCompile(`^\{([^{}]+)\}$`)
 
 // TODO: Handle/HandlerFunc + ServeHTTP (When there is a match, the route variables can be retrieved calling mux.Vars(request))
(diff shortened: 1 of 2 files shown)


routers/gorillamux/router.go
```diff
@@ -42,78 +49,23 @@ var singleVariableMatcher = regexp.MustCompile(`^\{([^{}]+)\}$`)
...
-	}
 	muxRouter := mux.NewRouter().UseEncodedPath()
 	r := &Router{}
 	for _, path := range orderedPaths(doc.Paths) {
+		servers := servers
+
 		pathItem := doc.Paths[path]
+		if len(pathItem.Servers) > 0 {
+			servers, err = makeServers(pathItem.Servers)
+			if err != nil {
```
fenollp, line 65:
```suggestion
			if servers, err = makeServers(pathItem.Servers); err != nil {
```

routers/gorillamux/router_test.go
```diff
@@ -254,6 +254,46 @@ func TestServerPath(t *testing.T) {
...
+	require.NoError(t, err)
+
+	req, err := http.NewRequest(http.MethodGet, "https://another.com/hello", nil)
+	require.NoError(t, err)
+	route, _, err := router.FindRoute(req)
+	require.Equal(t, "/hello", route.Path)
+
+	req, err = http.NewRequest(http.MethodGet, "https://example.com/hello", nil)
+	require.NoError(t, err)
+	route, _, err = router.FindRoute(req)
```
fenollp, line 293: Please do something with `route`.

fenollp: For instance compare both `route.PathItem`s for equality?

fenollp (changes requested):
Thanks for your patch! Please amend by my review and I'll merge it :)

routers/gorillamux/router_test.go
sonu27: It's meant to be nil as overriding in the spec replaces the default servers for the path.

fenollp:
Thanks :)

PR #569, opened 2022-07-15
Add DropResolvedRefs loader option to enable Value over Ref marshalling

john-dev:


fenollp:
Hi! Thanks for your patch. Would you say you'll only ever want to zero-out all the Refs or would there be cases where you'd like to keep some of them before serialization?

Also please include a test on a document that has ref cycles and handle these with an error.

openapi3/loader_test.go
```diff
@@ -283,6 +284,63 @@ func TestLoadFileWithExternalSchemaRefSingleComponent(t *testing.T) {
...
+    test:
+      description: 'This is a test description'
+      properties:
+        id:
+          type: string
+        test:
+          $ref: '#/components/schemas/test2'
+      type: object
+      additionalProperties: false
+    test2:
```
fenollp, line 304: Hi! Thanks for your patch. Would you say you'll only ever want to zero-out all the Refs or would there be cases where you'd like to keep some of them before serialization?

Also please include a test on a document that has ref cycles and handle these with an error.

fenollp (changes requested):

openapi3/loader_test.go
tjasko: I am also looking for a capability like this, though I agree, this may be too restrictive. Perhaps something like the below would be better:
```go
type Loader struct {
	// DropResolvedRefsFunc enables marshalling the (resolved) Value over the Ref.
	// When nil, the specification will not be flattened.
	DropResolvedRefsFunc DropResolvedRefsFunc
}

// DropResolvedRefsFunc defines a function that determines how Ref objects are marshalled. For the
// given reference, a boolean value should be returned that defines its marshalling strategy.
//
// Given an example specification of:
//  paths:
//    /pets/{petId}:
//      get:
//        parameters:
//          - name: petId
//            in: path
//            required: true
//            schema: {type: string}
//        responses:
//          200:
//            content:
//              application/json:
//                schema: {$ref: "#/components/schemas/Pet"}
//  components:
//    schemas:
//      Pet:
//        required: [id, name]
//        properties:
//          id: {type: integer, format: int64}
//          name: {type: string}
//          tag: {type: string}
//
// When the returned value of DropResolvedRefsFunc for the `#/components/schemas/Pet`
// Ref is false, the specification would be left untouched when marshalled.
// 
// When true, the specification would replace the reference with its (resolved) Value
// when marshalled. Additionally, all components are left untouched. Given the example
// above, such specification would be marshalled as such: 
//  paths:
//    /pets/{petId}:
//      get:
//        parameters:
//          - name: petId
//            in: path
//            required: true
//            schema: {type: string}
//        responses:
//          200:
//            content:
//              application/json:
//                schema:
//                  required: [id, name]
//                  properties:
//                    id: {type: integer, format: int64}
//                    name: {type: string}
//                    tag: {type: string}
//
//  components: [...]
type DropResolvedRefsFunc func(ref string) (dropResolvedRef bool)
```

That way, a user can do something like:
```go
l := openapi3.NewLoader()

// Honor the resolved Value for the Pet schema and use Ref for all other schemas.
l.DropResolvedRefsFunc = func(ref string) (dropResolvedRef bool) {
     if ref == "#/components/schemas/Pet" {
        return true
    }
    return false
}
```

Additionally, would it be worth implementing a solution for the `openapi2` package too?

fenollp: I don't think this capabilty should be tied to the `Loader`. Instead I picture a visitor function that takes a predicate, akin to https://pkg.go.dev/github.com/getkin/kin-openapi@v0.113.0/openapi3#T.InternalizeRefs

Feel free to take a try at it :)

fenollp:
Hi @john-dev do you need help moving this forward?

john-dev:
This pull_request is quite old. I am no longer investigating here so i'd vote for closing it.

Issue #570, opened 2022-07-21
Recursive definitions causes the parser to run indefinitely


shlomi-dr:
We are parsing a bunch of 3rd party openapi specs, not all are perfectly valid.. We came across the following example today:

```yaml
  "ReplicationSnapshotInfo": {
      "type": "object",
      "required": [
        "snappableId",
        "snapshotId"
      ],
       .....
        "childSnapshotInfos": {
          "type": "array",
          "description": "An array of child snapshots information.",
          "items": {
            "$ref": "#/definitions/ReplicationSnapshotInfo"
          }
        }
      }
    },
```
Where `ReplicationSnapshotInfo` is defined in terms of itself.

(See the original file here https://rubrikinc.github.io/api-doc-internal-6.0/openapi.json, line 28681, to reproduce in context)

When `childSnapshotInfos` was removed from this section, the parser works as expected, when its there, it goes into an infinite recursion and never halts.

I think it shouldn't be difficult for the parser to recognize such recursion by keeping a "visited" set, and checking that we are not recursing into something we already recursed in higher up the stack. Otherwise, as a plain user of this client, I don't know how else to avoid this, short of making a simplified parser that only looks for these loopy definitions.. 

Thing is that this runs on its own on many such files, so we can't manually inspect each (we couldn't even if we wanted to), and it just freezes our process without any way to recover.

fenollp:
This may be related to https://github.com/getkin/kin-openapi/issues/542

TristanSpeakEasy:
@shlomi-dr how would you expect this to be handled? An error returned that recursion was detected, so at least you can skip these (instead of having your process hang) or it to be handled in a particular way so you can actually work with the file?

shlomi-dr:
Hi @TristanSpeakEasy, thanks for taking this up! What I did in my own little "fix" for this (https://github.com/getkin/kin-openapi/pull/571/files) is to just return an error, which is sufficient for my use-case. 

If we want to be more "optimistic" about this, we could perhaps drop these "bad segments" and try to parse the rest of the file, but that may cause other troubles or inconsistencies further down the line, so perhaps this could be an option to the parser ("ErrorOnDeepRecursion" or something)

TristanSpeakEasy:
@shlomi-dr yeah in the PR you will see linked above I took the error approach for now as at list it gives you a way of being able to recover from the issue.

[...]

PR #571, opened 2022-07-21
Mark component.Ref as visited in resolveSchemRef

shlomi-dr:
First attempt at solving #570

### DIFF
--- a/openapi3/loader_recursive_ref_test.go
+++ b/openapi3/loader_recursive_ref_test.go
@@ -1,6 +1,7 @@
 package openapi3
 
 import (
+	"net/url"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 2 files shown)


shlomi-dr:
This fix seemed to work for that file I linked to in the issue

openapi3/loader_recursive_ref_test.go
```diff
@@ -47,3 +48,9 @@ components:
 		// type
 		Value.Type)
 }
+
+func TestIssue570(t *testing.T) {
+	loader := NewLoader()
+	_, err := loader.LoadFromURI(&url.URL{RawPath: "https://rubrikinc.github.io/api-doc-internal-6.0/openapi.json"})
```
fenollp, line 54: I'd suggest you first expand this link into its JSON then prune parts of that document until the infinite loop issue disappears. Then you'll have a minimized regression test and should be able to actually solve this from here :)

fenollp (changes requested):

openapi3/loader_recursive_ref_test.go
shlomi-dr: Yeah, you're absolutely right. I spent over 3 hours trying to do just that prior to submitting this commit (pruning to a minimum) but unfortunately I was unable to succeed, and could spend more time on this.. This is a gigantic file, and every time I thought it should happen, the problem didn't occur.. I left this in to at least show the problem. 

I know that as it is, this pr cannot be merged, hoping we could locate a better solution from the other pr you marked this as duplicate of. 

Thanks!!

Issue #482, opened 2022-01-31
integer support broken with yaml


cboitel:
Two use cases:
1. use openapi3 to load a document and ensure some data unarmshaled from a file complies with one of the schema (using VisitJSON)
2. use openapi3.filter to validate a request complies with a loaded openapi document

In both use cases, if a schema references a type integer and data was unmarshaled from YAML, we end up with " unhandled value of type int" error. If file or request is using JSON, no error is reported.
=> for a file, one can workaround the issue by marshaling the read YAML to JSON and unmarshal it back from JSON
=> for openapi3filter, no workaround found as of today.

I believe this is due to the wrong assumption that all numbers will be using float type as this is the case for JSON (no integers, only floats => json unmarshaling returns float only)
=> https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L795 only references float
=> https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L998 checks for integers but not its format (int32/64)

Support for integers type and its associated formats (int32/int64) is required for non-JSON content-types.

fenollp:
Sounds to me you're experiencing unintuitive YAML integer parsing but it might be something else.
Please provide minimal code reproducing the above mentioned error.

cboitel:
Here is an example:

```
package main

import (
	"bytes"
	"context"
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
	legacyrouter "github.com/getkin/kin-openapi/routers/legacy"
)

var openapiDoc = []byte(`
openapi: 3.0.3
info:
  version: v0.1.0
  title: Integer Sample API
paths:
  /samples:
    post:
      description: Sample operation
      requestBody:
        content:
          application/yaml:
            schema:
              $ref: '#/components/schemas/sample'
          application/json:
            schema:
              $ref: '#/components/schemas/sample'
      responses:
        "200":
          description: ok
components:
  schemas:
    sample:
      type: object
      properties:
        integer:
          type: integer
        number:
          type: number
`)

func main() {

	loader := openapi3.NewLoader()
	doc, err := loader.LoadFromData(openapiDoc)
	if err != nil {
		panic(err)
	}
	err = doc.Validate(loader.Context)
	if err != nil {
		panic(err)
	}

	router, err := legacyrouter.NewRouter(doc)
	if err != nil {
		panic(err)
	}
[...]

PR #577, opened 2022-08-01
#482 integer support broken with yaml

cboitel:
This PR aims at closing issue #482. 

I also added support for int32/64 integer formats during data validation.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -814,6 +814,12 @@ func (schema *Schema) visitJSON(settings *schemaValidationSettings, value interf
 	switch value := value.(type) {
 	case bool:
 		return schema.visitJSONBoolean(settings, value)
+	case int:
+		return schema.visitJSONNumber(settings, float64(value))
+	case int32:
+		return schema.visitJSONNumber(settings, float64(value))
+	case int64:
+		return schema.visitJSONNumber(settings, float64(value))
 	case float64:
 		return schema.visitJSONNumber(settings, value)
 	case string:
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -24,6 +24,12 @@ const (
 	TypeNumber  = "number"
 	TypeObject  = "object"
 	TypeString  = "string"
+
+	// constants for integer formats
+	formatMinInt32 = float64(math.MinInt32)
+	formatMaxInt32 = float64(math.MaxInt32)
+	formatMinInt64 = float64(math.MinInt64)
+	formatMaxInt64 = float64(math.MaxInt64)
 )
 
 var (
@@ -1019,7 +1025,7 @@ func (schema *Schema) VisitJSONNumber(value float64) error {
 func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value float64) error {
 	var me MultiError
 	schemaType := schema.Type
-	if schemaType == "integer" {
+	if schemaType == TypeInteger {
 		if bigFloat := big.NewFloat(value); !bigFloat.IsInt() {
 			if settings.failfast {
 				return errSchema
@@ -1037,6 +1043,39 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value 
 		}
 	} else if schemaType != "" && schemaType != TypeNumber {
 		return schema.expectedType(settings, "number, integer")
+	}
+
+	// formats
+	if schemaType == TypeInteger && schema.Format != "" {
+		formatMin := float64(0)
+		formatMax := float64(0)
+		switch schema.Format {
+		case "int32":
+			formatMin = formatMinInt32
+			formatMax = formatMaxInt32
+		case "int64":
+			formatMin = formatMinInt64
+			formatMax = formatMaxInt64
+		default:
+			if !SchemaFormatValidationDisabled {
+				return unsupportedFormat(schema.Format)
+			}
+		}
+		if formatMin != 0 && formatMax != 0 && !(formatMin <= value && value <= formatMax) {
+			if settings.failfast {
+				return errSchema
+			}
+			err := &SchemaError{
+				Value:       value,
+				Schema:      schema,
+				SchemaField: "format",
+				Reason:      fmt.Sprintf("number must be an %s", schema.Format),
+			}
+			if !settings.multiError {
+				return err
+			}
+			me = append(me, err)
+		}
 	}
 
 	// "exclusiveMinimum"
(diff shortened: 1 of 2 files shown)


openapi3/schema.go
```diff
@@ -1039,6 +1051,32 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
...
+		}
+		if !(value >= formatMin) || !(value <= formatMax) {
+			if settings.failfast {
+				return errSchema
+			}
+			err := &SchemaError{
+				Value:       value,
+				Schema:      schema,
+				SchemaField: "format",
+				Reason:      fmt.Sprintf("number must be a %s", schema.Format),
```
fenollp, line 1071:
```suggestion
				Reason:      fmt.Sprintf("number must be an %s", schema.Format),
```

openapi3/schema.go
```diff
@@ -1039,6 +1051,32 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
...
+	// formats
+	if schema.Format != "" {
+		formatMin := formatMinInt64
+		formatMax := formatMaxInt64
+		switch schema.Format {
+		case "int32":
+			formatMin = formatMinInt32
+			formatMax = formatMaxInt32
+		}
+		if !(value >= formatMin) || !(value <= formatMax) {
```
fenollp, line 1063:
```suggestion
		if !(formatMin <= value && value <= formatMax) {
```

openapi3/schema.go
```diff
@@ -1039,6 +1051,32 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
...
 
+	// formats
+	if schema.Format != "" {
+		formatMin := formatMinInt64
+		formatMax := formatMaxInt64
+		switch schema.Format {
+		case "int32":
+			formatMin = formatMinInt32
+			formatMax = formatMaxInt32
+		}
```
fenollp, line 1062: This wrongly assumes that only 2 formats are possible. Please default `formatMin,formatMat := 0, 0` (for instance) and from there set their values then check against default.

openapi3/schema_test.go
```diff
@@ -4,13 +4,13 @@ import (
...
 	"encoding/base64"
 	"encoding/json"
-	"fmt"
 	"math"
 	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
```
fenollp, line 13:
```suggestion
	"github.com/invopop/yaml"
```

openapi3/schema.go
```diff
@@ -1039,6 +1051,32 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
 		return schema.expectedType(settings, "number, integer")
 	}
 
+	// formats
+	if schema.Format != "" {
```
fenollp, line 1055:
```suggestion
	if schema.Type == "integer" && schema.Format != "" {
```

fenollp (changes requested):

openapi3/schema.go
cboitel: Yes but used TypeInteger instead of "integer". Also changed  line 1034 to also use TypeInteger

openapi3/schema.go
cboitel: yes

openapi3/schema.go
cboitel: I wrongly assumed the default format was int64. Will fix

openapi3/schema.go
cboitel: yes it make it more readable

openapi3/schema_test.go
cboitel: Well i disagree:
1. we want to use the yaml unmarshaller used in body decoder used for yaml: gopkg.in/yaml.v3 is used
2. invovop converts YAML to JSON before unmarshalling : as a consequence, we would end up with float instead of integers as we would have with yaml.v3 package

openapi3/schema.go
cboitel: Also will report an error if another format is provided

cboitel:
@fenollp : modified according to your requested changes except for sticking to use YAML parser used as body decoder.

Checks fails for it means you upgraded to GO v1.9 which does a bit of format differently. 

Do you want me to perform the required 1.9 changes in my PR ?

openapi3/schema.go
```diff
@@ -1039,6 +1053,37 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
...
+		formatMax := float64(0)
+		switch schema.Format {
+		case "int32":
+			formatMin = formatMinInt32
+			formatMax = formatMaxInt32
+		case "int64":
+			formatMin = formatMinInt64
+			formatMax = formatMaxInt64
+		default:
+			return ErrSchemaInvalidIntegerFormat
```
fenollp, line 1068: There's already https://github.com/getkin/kin-openapi/blob/fd4bae81cff0aae76fbea0f57ee8be5b1d9ec4f0/openapi3/schema.go#L668-L677

openapi3/schema.go
```diff
@@ -1039,6 +1053,37 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
...
+		}
+		if !(formatMin <= value && value <= formatMax) {
```
fenollp, line 1070: This is missing a defaults check: `formatMin, formatMax != float64(0), float64(0)`

fenollp (changes requested):
Just a couple things left :) 
Sure feel free to send a PR with go19 fmt changes.

openapi3/schema.go
cboitel: fixed

openapi3/schema.go
cboitel: yes, but there is only two values supported and any other value will lead to an error being returned. If you prefer, i may as well not return an error and check the default values instead.

cboitel:
@fenollp : fixes performed

openapi3/schema.go
fenollp: Ah sorry I forgot to mention that the OpenAPI specification does not forbid foreign formats so this code cannot just error here. 

Please wrap this error return on the condition of `SchemaFormatValidationDisabled`. See string formats. 

Then this'll get merged!

fenollp: Also please rebase on top of https://github.com/getkin/kin-openapi/pull/584

cboitel: @fenollp : rebased and added defaults check as well as SchemaFormatValidationDisabled

fenollp:
Thanks!!

PR #578, opened 2022-08-08
Protect from recursion in openapi3.InternaliseRefs

wtertius:
Fixes stack overflow panic when schema includes link to itself, like:

```
components:
  schemas:
    Item:
      type: object
      properties:
        Name:
          type: string
        ParentLink:
          $ref: '#/components/schemas/Item'
```

### DIFF
--- a/openapi3/visited.go
+++ b/openapi3/visited.go
@@ -22,15 +22,15 @@ type ctxKey struct{}
 // Returned ctx can be passed to children function calls and used to check whether
 // component already visited. For instance:
 //
-//     ctx := context.Background()
-//     ctx = newVisited().withContext(ctx)
-//     ...
-//     doc.deferSchemaRecursively(ctx, schema)
-//     func (doc *T) deferSchemaRecursively(ctx context.Context, s *Schema) {
-//         if s == nil || isVisitedSchema(ctx, s) {
-//             return
-//         }
-//     }
+//	ctx := context.Background()
+//	ctx = newVisited().withContext(ctx)
+//	...
+//	doc.deferSchemaRecursively(ctx, schema)
+//	func (doc *T) deferSchemaRecursively(ctx context.Context, s *Schema) {
+//	    if s == nil || isVisitedSchema(ctx, s) {
+//	        return
+//	    }
+//	}
 func (v *visitedComponent) withContext(ctx context.Context) context.Context {
 	if visited, ok := ctx.Value(ctxKey{}).(*visitedComponent); ok {
 		if visited == v {
--- a/openapi3/visited.go
+++ b/openapi3/visited.go
@@ -2,8 +2,8 @@ package openapi3
 
 import "context"
 
-func newVisited() *visitedComponent {
-	return &visitedComponent{
+func newVisited() visitedComponent {
+	return visitedComponent{
 		header: make(map[*Header]struct{}),
 		schema: make(map[*Schema]struct{}),
 	}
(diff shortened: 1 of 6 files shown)


openapi3/visited.go
```diff
@@ -0,0 +1,81 @@
...
+//	ctx := context.Background()
+//	ctx = newVisited().withContext(ctx)
+//	...
+//	doc.deferSchemaRecursively(ctx, schema)
+//	func (doc *T) deferSchemaRecursively(ctx context.Context, s *Schema) {
+//	    if s == nil || isVisitedSchema(ctx, s) {
+//	        return
+//	    }
+//	}
+func (v *visitedComponent) withContext(ctx context.Context) context.Context {
```
fenollp, line 34: I'm not sure we should be adding things to ctx. What made you store these maps in here? Could this be moved to the root document type T instead?

Thanks

fenollp (changes requested):

openapi3/visited.go
wtertius: It can be moved to document type T, but it have one minus: changing document we change it globally. It means, that it's not thread safe. Context allows us to restrict recursion only for this exact call of `InternalizeRef`.

If you ok with that minus, I can add the maps to root document type T and clean them before exiting `InternalizeRef`.
What do you decide?

Thanks

wtertius: Moved the visitedComponent to doc *T.
Could you please check?

fenollp (approved):
LGTM. I'm just wondering why you're covering just schema and parameter when all these should match the pattern you're going for?
```
openapi3/internalize_refs.go:202:func (doc *T) derefSchema(s *Schema, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:229:func (doc *T) derefHeaders(hs Headers, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:239:func (doc *T) derefExamples(es Examples, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:245:func (doc *T) derefContent(c Content, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:258:func (doc *T) derefLinks(ls Links, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:264:func (doc *T) derefResponses(es Responses, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:275:func (doc *T) derefParameter(p Parameter, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:283:func (doc *T) derefRequestBody(r RequestBody, refNameResolver RefNameResolver) {
openapi3/internalize_refs.go:287:func (doc *T) derefPaths(paths map[string]*PathItem, refNameResolver RefNameResolver) {
```

wtertius:
> LGTM. I'm just wondering why you're covering just schema and parameter when all these should match the pattern you're going for?
> 
> ```
> openapi3/internalize_refs.go:202:func (doc *T) derefSchema(s *Schema, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:229:func (doc *T) derefHeaders(hs Headers, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:239:func (doc *T) derefExamples(es Examples, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:245:func (doc *T) derefContent(c Content, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:258:func (doc *T) derefLinks(ls Links, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:264:func (doc *T) derefResponses(es Responses, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:275:func (doc *T) derefParameter(p Parameter, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:283:func (doc *T) derefRequestBody(r RequestBody, refNameResolver RefNameResolver) {
> openapi3/internalize_refs.go:287:func (doc *T) derefPaths(paths map[string]*PathItem, refNameResolver RefNameResolver) {
> ```

Actually, that was what I planned firstly. But when I started to implement it, I discovered, that
1) the recursion problem now is possible only for schema and parameters and there is no way to write a test for methods without recursion
2) many of methods are for slices of items - so the recursion protection should be in the place where it is iterates over and call some method down - that is not written yet
3) For many of the items recursion doesn't have physical meaning: like header probably will not include responses
That's why I limited this MR with only two of items.

If you'd like, I can try to write a syntetic test for every item that try to call method for the same item twice and checks that the second time is protected from recursion. Then I can write the protection for every item.
Should I?

fenollp:
> If you'd like, I can try to write a syntetic test for every item that try to call method for the same item twice and checks that the second time is protected from recursion. Then I can write the protection for every item.

Sounds great :)

fenollp:
Please rebase on top of https://github.com/getkin/kin-openapi/pull/584

fenollp:
Thanks :)

wtertius:
You too ))

Issue #542, opened 2022-05-16
Loading openapi schema with loop references ends-up with endless loop and crashing with "stack overflow" error


panush:
I'm using kin-openapi v0.89.0.
when trying to load the below openapi (with a reference loop in it), kin-openapi is crashing after getting into an endless loop (see attached file)
[kin-openapi-endless-loop-error.txt](https://github.com/getkin/kin-openapi/files/8701649/kin-openapi-endless-loop-error.txt)

```yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
servers:
  - url: http://petstore.swagger.io/v1
paths:
  /pets:
    patch:
      requestBody:
        content:
          application/json:
            schema:
              oneOf:
                - $ref: '#/components/schemas/Cat'
                - $ref: '#/components/schemas/Kitten'
              discriminator:
                propertyName: pet_type
      responses:
        '200':
          description: Updated
components:
  schemas:
    Cat:
      anyOf: 
        - $ref: '#/components/schemas/Kitten'
        - type: object
          properties:
            hunts:
              type: boolean
            age:
              type: integer
            offspring:
    Kitten: 
      $ref: '#/components/schemas/Cat'
```

fenollp:
Hi @panush please use backticks to share your YAML. See https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#quoting-code

Am I correct in assuming that components' Cat's `offspring` field is empty?

chanced:
I'm trying to define the json schema's [recommended response structure](https://json-schema.org/draft/2019-09/output/schema) but I'm running into the stack overflow. Unless I'm missing something, the `allOf` workaround isn't working for it.

```yaml
components:
    schemas:
       ValidationError:
            type: object
            properties:
                parameters:
                    allOf:
                        - $ref: "#/components/schemas/ValidationErrorOutputUnit"
                body:
                    allOf:
                        - $ref: "#/components/schemas/ValidationErrorOutputUnit"
        ValidationErrorOutputUnitArray:
            type: array
            items:
                allOf:
                    - $ref: "#/components/schemas/ValidationErrorOutputUnit"
        ValidationErrorOutputUnit:
            description: A schema that validates the minimum requirements for validation output
            properties:
                valid:
                    type: boolean
                keywordLocation:
                    type: string
                    format: uri-reference
[...]

PR #581, opened 2022-08-13
reproduce #542

fenollp:


fenollp:
Fixed long ago. See https://github.com/getkin/kin-openapi/commits/master/openapi3/issue542_test.go

PR #583, opened 2022-08-23
Fix `resolveSchemaRef()` to load correctly an other spec. file referenced by `$ref`

masu-mi:
# Fix `resolveSchemaRef()` to load correctly an other spec. file referenced by `$ref`

I can't find contribution guidelines. If you know it, would you let me know it?
At the following case, `resolveSchemaRef()` fails to load other spec. files.

* A schema participate a circular reference.
* The schema has references to component's schema in external spec. files.

## Changes

* Add test case: `TestLoadCircularRefFromFile`
* Change `resolveSchemaRef()` to fix it.

## Cause analysis
`resolveSchemaRef()` miss file path's information during recursive resolving schema reference.
For describe, I'm going use following example.

```directory tree
./testdata
    |- a.yml
    |- b.yml
```

```a.yml
# a.yml
components:
  schemas:
    Foo:
      properties:
        bar:
          $ref: "#/components/schemas/Bar"
        foo2:
          $ref: "b.yml#/components/schemas/Foo"
    Bar:
      properties:
        foo:
          $ref: "#/components/schemas/Foo"
```

```b.yml
# b.yml
components:
  schemas:
    Foo:
      properties:
        id:
          type: string
```

Resolving `Foo`, if process load `foo2` before `bar`, it loads with an order as `Foo` -> `Bar` -> `Foo`.
During 2nd loading `Foo` schema, because `loader.resolveComponent()` set a instance with empty Ref field to `resolved`, the process lost file path information.

Because of golang specification, order of for-loop over map is random, I fail to resolve Foo randomly.

### DIFF
--- a/openapi3/load_cicular_ref_with_external_file_test.go
+++ b/openapi3/load_cicular_ref_with_external_file_test.go
@@ -1,0 +1,30 @@
+//go:build go1.16
+// +build go1.16
+
+package openapi3_test
+
+import (
+	"embed"
+	"fmt"
+	"net/url"
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+//go:embed testdata/circularRef/*
+var circularResSpecs embed.FS
+
+func TestLoadCircularRefFromFile(t *testing.T) {
+	loader := openapi3.NewLoader()
+	loader.IsExternalRefsAllowed = true
+	loader.ReadFromURIFunc = func(loader *openapi3.Loader, uri *url.URL) ([]byte, error) {
+		return circularResSpecs.ReadFile(uri.Path)
+	}
+
+	doc, err := loader.LoadFromFile("testdata/circularRef/base.yml")
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("%v\n", doc)
+}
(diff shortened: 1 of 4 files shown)


openapi3/load_cicular_ref_with_external_file_test.go
```diff
@@ -0,0 +1,30 @@
...
+	loader.IsExternalRefsAllowed = true
+	loader.ReadFromURIFunc = func(loader *openapi3.Loader, uri *url.URL) ([]byte, error) {
+		return circularResSpecs.ReadFile(uri.Path)
+	}
+
+	doc, err := loader.LoadFromFile("testdata/circularRef/base.yml")
+	if err != nil {
+		t.Error(err)
+	}
+	fmt.Printf("%v\n", doc)
```
fenollp, line 29: Please use package `require` to assert for absence of error above and presence of correctly parsed data here. 

I'll take care of the go119 fmt changes.

fenollp (changes requested):

fenollp:
Please rebase on top of https://github.com/getkin/kin-openapi/pull/584

masu-mi:
@fenollp Thank you for you feedback. I've rebased this PR and update test code to add assertion with `require` package.

fenollp:
Thanks!

PR #592, opened 2022-09-08
Examples validation

### LINKED ISSUE #588: Examples not being validated
As per #526 I'd expect schema `examples` to be validated as well, but I can't get these tests to fail validation using either `examples` or `example` (this one not implemented afaik). Am I missing something in the code below?

```go
package openapi3

import (
	"testing"

)

func TestExamplesValidation(t *testing.T) {
	spec := []byte(`
openapi: 3.0.0

paths:
  /user:
    post:
      description: User creation.
      operationId: createUser
      requestBody:
        content:
          application/json:
            schema:
              $ref: "#/components/schemas/CreateUserRequest"
            examples:
              BadUser:
                $ref: '#/components/examples/BadUser'
        description: Created user object
        required: true
      responses:
        "201":
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/CreateUserResponse"
          description: successful operation
      # security:
      #   - api_key: []
      summary: creates a new user
      tags:
        - user
components:
  schemas:
    CreateUserRequest:
      description: represents a new User
      example:
        username: "]bad["
        email: bad
        password: short
      required:
        - username
        - email
        - password
      properties:
        username:
          type: string
          pattern: "^[ a-zA-Z0-9_-]+$"
          minLength: 3
        email:
          type: string
          pattern: "^[A-Za-z0-9+_.-]+@(.+)$"
        password:
          type: string
          minLength: 7
      title: a User
      type: object
    CreateUserResponse:
      description: represents the response to a User creation
      properties:
        access_token:
          type: string
        user_id:
          format: int64
          type: integer
      type: object
# implemented in #526
  examples:
    BadUser:
      value:
        username: "]bad["
        email: bad
        password: short
info:
  title: An API
  version: 1.2.3.4
`)

	loader := NewLoader()

	doc, err := loader.LoadFromData(spec)
	require.NoError(t, err)

	err = doc.Validate(loader.Context)
	require.Error(t, err) // FAIL
}
```

**EDIT**: well this explains it:

```go
// Validate returns an error if Example does not comply with the OpenAPI spec.
func (example *Example) Validate(ctx context.Context) error {
	return nil // TODO
}
```

Is this being worked on/planned/would accept a PR for it?

danicc097:
First draft to fix #588. Let me know what you think @fenollp before digging more into it and having proper tests.

Should this be non-configurable? Based on the spec: 

> Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.`


openapi3/example.go
```diff
@@ -55,5 +55,9 @@ func (example *Example) UnmarshalJSON(data []byte) error {
 
 // Validate returns an error if Example does not comply with the OpenAPI spec.
 func (example *Example) Validate(ctx context.Context) error {
-	return nil // TODO
+	if v := example.Value; v == nil && example.ExternalValue == "" {
+		return fmt.Errorf("example has no value field")
+	}
+
```
fenollp, line 61: I'd also add some code that ensures
>  The value field and externalValue field are mutually exclusive.

openapi3/example_validation_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	doc, err := loader.LoadFromData(spec)
+	require.NoError(t, err)
+
+	err = doc.Validate(loader.Context)
+	t.Logf("error: %v", err)
+	require.Error(t, err)
+}
+
+// XXX should this be a validation error or skipped?
```
fenollp, line 247: This should be an error complaining that either `value` or `externalValue` should be set.

openapi3/example_validation_test.go
```diff
@@ -0,0 +1,268 @@
...
+      properties:
+        access_token:
+          type: string
+        user_id:
+          format: int64
+          type: integer
+      type: object
+info:
+  title: An API
+  version: 1.2.3.4
```
fenollp, line 234: I see these docs are mostly the same and are quite large in that it's not easy to notice their differences. Maybe use string concatenation to show the reusable parts and the differing parts.

openapi3/example_validation.go
```diff
@@ -0,0 +1,18 @@
+package openapi3
+
+import (
+	"context"
+)
+
+func ValidateExampleValue(ctx context.Context, input interface{}, schema *Schema) error {
+	opts := make([]SchemaValidationOption, 0, 3) // 3 potential opts here
```
fenollp, line 8: I only see two...

openapi3/example_validation.go
```diff
@@ -0,0 +1,18 @@
...
+	opts = append(opts, VisitAsRequest())
```
fenollp, line 9: Then what about examples in responses?

openapi3/media_type.go
```diff
@@ -80,7 +81,22 @@ func (mediaType *MediaType) Validate(ctx context.Context) error {
 		if err := schema.Validate(ctx); err != nil {
 			return err
 		}
+		if example := mediaType.Example; example != nil {
+			if err := ValidateExampleValue(ctx, example, schema.Value); err != nil {
+				return err
+			}
+		} else if examples := mediaType.Examples; examples != nil {
```
fenollp, line 88: This needs another check that ensures
>  The example object is mutually exclusive of the examples object.

openapi3/refs.go
```diff
@@ -290,6 +291,12 @@ func (value *SchemaRef) UnmarshalJSON(data []byte) error {
 // Validate returns an error if SchemaRef does not comply with the OpenAPI spec.
 func (value *SchemaRef) Validate(ctx context.Context) error {
 	if v := value.Value; v != nil {
+		if x := v.Example; x != nil {
+			if err := ValidateExampleValue(ctx, x, v); err != nil {
+				return fmt.Errorf("invalid schema example: %s", err)
+			}
+		}
+
```
fenollp, line 299: Actually this must go within `func (schema *Schema) Validate(ctx context.Context) error {...}` otherwise validating schemas without dereferencing a `$ref` won't go through this code.

fenollp (changes requested):
Thanks for taking the time to do this :)

---

> Should this be non-configurable? Based on the spec:
>>    Tooling implementations MAY choose to validate compatibility automatically, and reject the example value(s) if incompatible.`

We recently merged #590 which added https://github.com/getkin/kin-openapi/blob/46603c3b350c28a74763bda34fd4fe2142a76145/openapi3/validation_options.go#L9-L12
so I'd suggest to add an option there with strict defaults.

openapi3/example_validation.go
danicc097: Very true + added tests for this

openapi3/example_validation_test.go
danicc097: It certainly was a bad idea to choose YAML for this. Let me know if now it's decent enough or should refactor to JSON/build it from structs

danicc097:
I think that should cover just about everything (valid request, response and schema examples or disabled validation flag raise no error at all, else it errors out). Thanks for the inputs and if you have any further comments PLMK

openapi3/media_type.go
```diff
@@ -73,14 +74,37 @@ func (mediaType *MediaType) UnmarshalJSON(data []byte) error {
 
 // Validate returns an error if MediaType does not comply with the OpenAPI spec.
 func (mediaType *MediaType) Validate(ctx context.Context) error {
+	validationOpts := getValidationOptions(ctx)
```
fenollp, line 77: Think of the hot path. Move this where it's used `if validationOpts := getValidationOptions(ctx); validationOpts.ExamplesValidationDisabled { ...`

openapi3/example_validation.go
```diff
@@ -0,0 +1,19 @@
+package openapi3
+
+import (
+	"context"
+)
+
+func ValidateExampleValue(ctx context.Context, input interface{}, schema *Schema) error {
```
fenollp, line 7: Why is this public?

openapi3/example.go
```diff
@@ -55,5 +56,17 @@ func (example *Example) UnmarshalJSON(data []byte) error {
 
 // Validate returns an error if Example does not comply with the OpenAPI spec.
 func (example *Example) Validate(ctx context.Context) error {
-	return nil // TODO
+	validationOpts := getValidationOptions(ctx)
+
+	if validationOpts.ExamplesValidationDisabled {
```
fenollp, line 61: Not that kind of validation. The validation of the example against the schema should be opt-out, the validation of the example object should not.

openapi3/example_validation.go
```diff
@@ -0,0 +1,19 @@
...
+	opts := make([]SchemaValidationOption, 0, 3) // 3 potential opts here
+	opts = append(opts, VisitAsResponse())
+	opts = append(opts, VisitAsRequest())
```
fenollp, line 10: What are you trying to do with these options here and how do you think this gets tested?

fenollp (changes requested):

danicc097:
I added parameters validation which I completely forgot and implemented all changes requested. Do you prefer the `mustErr` field  trick for example object validation or duplicate to its own `TestExamplesValidation` and rename the rest to `TestExamplesSchemaValidation` or something like that?

openapi3/example_validation_test.go
```diff
@@ -8,22 +8,51 @@ import (
...
-		errContains            string
-		raisesValidationErr    bool
+		name                    string
+		requestSchemaExample    string
+		responseSchemaExample   string
+		mediaTypeRequestExample string
+		parametersExample       string
+		componentExamples       string
+		errContains             string
+		mustErr                 bool
```
fenollp, line 19: To me `mustErr` is redundant with `errContains`: `mustErr = errContains != ""`.

openapi3/parameter.go
```diff
@@ -307,6 +307,26 @@ func (parameter *Parameter) Validate(ctx context.Context) error {
 		if err := schema.Validate(ctx); err != nil {
 			return fmt.Errorf("parameter %q schema is invalid: %v", parameter.Name, err)
 		}
+		if parameter.Example != nil && parameter.Examples != nil {
+			return fmt.Errorf("%s: example and examples are mutually exclusive", schema.Ref)
```
fenollp, line 311:
```suggestion
			return fmt.Errorf("parameter %q example and examples are mutually exclusive", parameter.Name)
```

openapi3/media_type.go
```diff
@@ -80,7 +81,28 @@ func (mediaType *MediaType) Validate(ctx context.Context) error {
 		if err := schema.Validate(ctx); err != nil {
 			return err
 		}
+		if mediaType.Example != nil && mediaType.Examples != nil {
+			return fmt.Errorf("%s: example and examples are mutually exclusive", schema.Ref)
```
fenollp, line 85: `schema.Ref` can be empty. You'll have to use something else :)

openapi3/parameter.go
```diff
@@ -307,6 +307,26 @@ func (parameter *Parameter) Validate(ctx context.Context) error {
...
+		}
+		if validationOpts := getValidationOptions(ctx); validationOpts.ExamplesValidationDisabled {
+			return nil
```
fenollp, line 314: Can't return here. There's non-Example code below this block...

fenollp (changes requested):

openapi3/example_validation_test.go
```diff
@@ -0,0 +1,426 @@
...
+					} else {
+						require.NoError(t, err)
+					}
+				})
+			}
+		})
+	}
+}
+
+func TestExampleObjectValidation(t *testing.T) {
```
danicc097, line 261: to get rid of `mustErr` and make intent clearer

openapi3/example_validation.go
```diff
@@ -0,0 +1,9 @@
+package openapi3
+
+import (
+	"context"
+)
+
+func validateExampleValue(ctx context.Context, input interface{}, schema *Schema) error {
```
fenollp, line 7:
```suggestion
func validateExampleValue(input interface{}, schema *Schema) error {
```

openapi3/media_type.go
```diff
@@ -80,7 +82,28 @@ func (mediaType *MediaType) Validate(ctx context.Context) error {
...
 			return err
 		}
+		if mediaType.Example != nil && mediaType.Examples != nil {
+			return errors.New("example and examples are mutually exclusive")
+		}
+		if validationOpts := getValidationOptions(ctx); validationOpts.ExamplesValidationDisabled {
+			return nil
+		}
+		if example := mediaType.Example; example != nil {
+			if err := validateExampleValue(ctx, example, schema.Value); err != nil {
```
fenollp, line 92:
```suggestion
			if err := validateExampleValue(example, schema.Value); err != nil {
```

openapi3/media_type.go
```diff
@@ -80,7 +82,28 @@ func (mediaType *MediaType) Validate(ctx context.Context) error {
...
+				return err
+			}
+		} else if examples := mediaType.Examples; examples != nil {
+			for k, v := range examples {
+				if err := v.Validate(ctx); err != nil {
+					return fmt.Errorf("%s: %s", k, err)
+				}
+				if err := validateExampleValue(ctx, v.Value.Value, schema.Value); err != nil {
```
fenollp, line 100:
```suggestion
				if err := validateExampleValue(v.Value.Value, schema.Value); err != nil {
```

openapi3/parameter.go
```diff
@@ -303,16 +303,38 @@ func (parameter *Parameter) Validate(ctx context.Context) error {
...
+			return fmt.Errorf("parameter %q schema is invalid: %v", parameter.Name, err)
+		}
+		if parameter.Example != nil && parameter.Examples != nil {
+			return fmt.Errorf("parameter %q example and examples are mutually exclusive", parameter.Name)
+		}
+		if validationOpts := getValidationOptions(ctx); validationOpts.ExamplesValidationDisabled {
+			return nil
+		}
+		if example := parameter.Example; example != nil {
+			if err := validateExampleValue(ctx, example, schema.Value); err != nil {
```
fenollp, line 324:
```suggestion
			if err := validateExampleValue(example, schema.Value); err != nil {
```

openapi3/parameter.go
```diff
@@ -303,16 +303,38 @@ func (parameter *Parameter) Validate(ctx context.Context) error {
...
+				return err
+			}
+		} else if examples := parameter.Examples; examples != nil {
+			for k, v := range examples {
+				if err := v.Validate(ctx); err != nil {
+					return fmt.Errorf("%s: %s", k, err)
+				}
+				if err := validateExampleValue(ctx, v.Value.Value, schema.Value); err != nil {
```
fenollp, line 332:
```suggestion
				if err := validateExampleValue(v.Value.Value, schema.Value); err != nil {
```

openapi3/schema.go
```diff
@@ -753,6 +753,12 @@ func (schema *Schema) validate(ctx context.Context, stack []*Schema) (err error)
 		}
 	}
 
+	if x := schema.Example; x != nil && !validationOpts.ExamplesValidationDisabled {
+		if err := validateExampleValue(ctx, x, schema); err != nil {
```
fenollp, line 757:
```suggestion
		if err := validateExampleValue(x, schema); err != nil {
```

fenollp (approved):

fenollp:
Thanks a lot for this work! 

I note support for `readOnly` and `writeOnly` schema fields is missing (see [spec the here](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#fixed-fields-20)). This is where `VisitAsRequest()` should be employed when validating examples that go in requests and respectively `VisitAsResponse()` responses. Let's do this as another PR!

PR #596, opened 2022-09-16
use %w to wrap the errors

SVilgelm:
Update all usage of the `fmt.Errorf` to use `%w` instead of `%s` or `%v` for the wrapped errors

### DIFF
--- a/jsoninfo/unmarshal.go
+++ b/jsoninfo/unmarshal.go
@@ -25,7 +25,7 @@ type ObjectDecoder struct {
 func NewObjectDecoder(data []byte) (*ObjectDecoder, error) {
 	var remainingFields map[string]json.RawMessage
 	if err := json.Unmarshal(data, &remainingFields); err != nil {
-		return nil, fmt.Errorf("failed to unmarshal extension properties: %v (%s)", err, data)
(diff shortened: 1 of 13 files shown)


fenollp:
I guess tests can be fixed with `errors.Is`.

How can we lint for this? So CI can look for these cases.

SVilgelm:
> I guess tests can be fixed with `errors.Is`.

yeah, I will fix the errors in the unit tests

> How can we lint for this? So CI can look for these cases.

I remember there was a linter for this in golangci-lint, I will check and add to the workflows

SVilgelm:
@fenollp I have fixed the test by using `require.EqualError` that compares an error with a string. We cannot use `errors.Is|As` because Is function can compare the dedicated types only, Is returns false for two errors with same text: https://go.dev/play/p/DrWaY0cQQDx

second, I found `https://github.com/polyfloyd/go-errorlint`, it can validate the usage of %w: `non-wrapping format verb for fmt.Errorf. Use `%w` to format errors`. But the linter found bunch of different errors, like `comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error` or `type assertion on error will fail on wrapped errors. Use errors.As to check for specific errors`, I will prepare another PR to fix all those issues and integrate the linter

PR #604, opened 2022-09-19
fix: only inject default value for matched oneOf or anyOf

nic-6443:
Signed-off-by: Nic <qianyong@api7.ai>


nic-6443:
**This PR is an improvement for https://github.com/getkin/kin-openapi/pull/544, please take a look.** @fenollp

openapi3/schema.go
```diff
@@ -956,17 +960,22 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 
 			return e
 		}
+
+		_ = v[matchedOneOfIdx].Value.visitJSON(settings, value)
```
fenollp, line 964: Please wrap these operations and the deepcopy inside 
```go
	if settings.asreq || settings.asrep {
```
the same way that the code that mutates `value` does
```go
	if settings.asreq || settings.asrep {
		for propName, propSchema := range schema.Properties {
			if value[propName] == nil {
				if dlft := propSchema.Value.Default; dlft != nil {
					value[propName] = dlft
					if f := settings.defaultsSet; f != nil {
						settings.onceSettingDefaults.Do(f)
					}
				}
			}
		}
	}
```
This way the general case isn't costlier because of this PR.

openapi3/schema.go
```diff
@@ -921,11 +923,13 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 				continue
 			}
 
-			if err := v.visitJSON(settings, value); err != nil {
+			tempValue := deepcopy.Copy(value)
```
fenollp, line 926: Move this outside the loop.

openapi3/schema.go
```diff
@@ -956,17 +960,22 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
...
 		ok := false
-		for _, item := range v {
+		matchedAnyOfIdx := 0
+		for idx, item := range v {
 			v := item.Value
 			if v == nil {
 				return foundUnresolvedRef(item.Ref)
 			}
-			if err := v.visitJSON(settings, value); err == nil {
+			tempValue := deepcopy.Copy(value)
```
fenollp, line 975: Move this outside the loop.

openapi3/schema.go
```diff
@@ -980,6 +989,8 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 				SchemaField: "anyOf",
 			}
 		}
+
+		_ = v[matchedAnyOfIdx].Value.visitJSON(settings, value)
 	}
 
 	for _, item := range schema.AllOf {
```
fenollp, line 996: And what about `allOf`?

fenollp (changes requested):

openapi3/schema.go
nic-6443: Make sense, fixed.

openapi3/schema.go
nic-6443: Done.

openapi3/schema.go
nic-6443: Done.

pstefanovic: just fyi @nic-6443 @fenollp (ran into this PR while debugging)

Mutating `tempValue` across schemas listed under an `allOf` or `anyOf` brings about an issue in case two schemas declare a property with the same name but say a different default value, validation rules, etc.

PR #605, opened 2022-09-20
Improve error message when path validation fails

praneetloke:
With the introduction of validation of examples in an OpenAPI spec in #592 there is now a high chance that many will have specs that fail validation. I did! It was hard to debug what was wrong with the spec as the errors were too generic. The change in this PR will hopefully make it easier for folks to debug the problem in their OpenAPI docs.

### DIFF
--- a/openapi3/example_validation_test.go
+++ b/openapi3/example_validation_test.go
@@ -26,7 +26,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
             param1example:
               value: abcd
    `,
-			errContains: "param1example",
+			errContains: "invalid paths: invalid path /user: invalid operation POST: param1example",
 		},
 		{
 			name: "valid_parameter_examples",
@@ -64,7 +64,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
         email: bad
         password: short
    `,
-			errContains: "BadUser",
+			errContains: "invalid paths: invalid path /user: invalid operation POST: BadUser",
 		},
 		{
 			name: "valid_component_examples",
--- a/openapi3/example_validation_test.go
+++ b/openapi3/example_validation_test.go
@@ -26,7 +26,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
             param1example:
               value: abcd
    `,
-			errContains: "invalid paths: param1example",
+			errContains: "param1example",
 		},
 		{
 			name: "valid_parameter_examples",
@@ -64,7 +64,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
         email: bad
         password: short
    `,
-			errContains: "invalid paths: BadUser",
+			errContains: "BadUser",
 		},
 		{
 			name: "valid_component_examples",
--- a/openapi3/loader_test.go
+++ b/openapi3/loader_test.go
@@ -91,7 +91,7 @@ func TestResolveSchemaRef(t *testing.T) {
 }
 
 func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
-	source := []byte(`{"openapi":"3.0.0","info":{"title":"MyAPI","version":"0.1","description":"An API"},"paths":{"/foo":{"post":{"summary":"Create Foo","requestBody":{"content":{"application/json":{"schema":null}}}}}}}`)
+	source := []byte(`{"openapi":"3.0.0","info":{"title":"MyAPI","version":"0.1","description":"An API"},"paths":{"/foo":{"post":{"requestBody":{"content":{"application/json":{"schema":null}}}}}}}`)
 	loader := NewLoader()
 	doc, err := loader.LoadFromData(source)
 	require.NoError(t, err)
--- a/openapi3/loader_test.go
+++ b/openapi3/loader_test.go
@@ -96,7 +96,7 @@ func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
 	doc, err := loader.LoadFromData(source)
 	require.NoError(t, err)
 	err = doc.Validate(loader.Context)
-	require.EqualError(t, err, `invalid paths: path /foo failed validation for operation POST: found unresolved ref: ""`)
+	require.EqualError(t, err, `invalid paths: path /foo failed validation: validation failed for operation POST: found unresolved ref: ""`)
 }
 
 func TestResolveResponseExampleRef(t *testing.T) {
--- a/openapi3/loader_test.go
+++ b/openapi3/loader_test.go
@@ -96,7 +96,7 @@ func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
 	doc, err := loader.LoadFromData(source)
 	require.NoError(t, err)
 	err = doc.Validate(loader.Context)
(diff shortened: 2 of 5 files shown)


openapi3/path_item.go
```diff
@@ -123,9 +123,9 @@ func (pathItem *PathItem) SetOperation(method string, operation *Operation) {
 
 // Validate returns an error if PathItem does not comply with the OpenAPI spec.
 func (pathItem *PathItem) Validate(ctx context.Context) error {
-	for _, operation := range pathItem.Operations() {
+	for method, operation := range pathItem.Operations() {
 		if err := operation.Validate(ctx); err != nil {
-			return err
+			return fmt.Errorf("path (summary: %s) failed validation for operation %s: %v", pathItem.Summary, method, err)
```
praneetloke, line 128: I didn't want to modify the signature of the `Validate` method to take the `path` as input as that might mean more changes. I didn't check if it would be. This seemed like an acceptable minimal change to improve the error messaging. Let me know what you think.

praneetloke:
Hmm I looked at the failing tests and tried to update them. But something funky is going on with the unmarshal code. It seems that the value receiver method `Validate` is losing the value for the string properties in the struct somehow. Is this because it's a value receiver instead of a [pointer receiver](https://go.dev/tour/methods/4)? Basically, `Summary` is always blank in the `Validate` method for a `PathItem` and so is `Description` even when I update the test OpenAPI JSON docs that are in each of those tests. I do see the value in those properties just before I enter the `Validate` method. That's when it loses them.

openapi3/path_item.go
```diff
@@ -122,10 +122,10 @@ func (pathItem *PathItem) SetOperation(method string, operation *Operation) {
 }
 
 // Validate returns an error if PathItem does not comply with the OpenAPI spec.
-func (pathItem *PathItem) Validate(ctx context.Context) error {
-	for _, operation := range pathItem.Operations() {
+func (pathItem *PathItem) Validate(ctx context.Context, path string) error {
```
fenollp, line 125: I appreciate the effort & I think this is worthwhile! However I'd like to not change the prototype of the Validate here.

Can't you just add the path in case of error in the caller to this Validate?
```go
// somewhere in callback.go and paths.go...
		if err := v.Validate(ctx); err != nil {
			return fmt.Errorf("path %s failed validation %v", path, err)
		}

// ...and in here path_item.go:
		if err := operation.Validate(ctx); err != nil {
			return fmt.Errorf("for operation %s: %v", method, err)
		}
```
What do you think? This is probably complexifying things for not much gain actually

praneetloke: Yes, indeed. I can wrap the error when it bubbles up to the caller and add the path information there. I'll make the change.

praneetloke: @fenollp I've made the change.

praneetloke: I just saw https://github.com/getkin/kin-openapi/pull/596 in the latest release. I don't think there is a need to unwrap the inner errors in this case so I think using `%v` is fine here. Let me know what you think.

openapi3/loader_test.go
```diff
@@ -96,7 +96,7 @@ func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
 	doc, err := loader.LoadFromData(source)
 	require.NoError(t, err)
 	err = doc.Validate(loader.Context)
-	require.EqualError(t, err, `invalid paths: path /foo failed validation for operation POST: found unresolved ref: ""`)
+	require.EqualError(t, err, `invalid paths: path /foo failed validation: validation failed for operation POST: found unresolved ref: ""`)
```
fenollp: Well if you could get the message to be thus that'd be easier to read I think:
> invalid paths: path /foo failed validation: for operation POST: found unresolved ref: ""

fenollp (changes requested):

openapi3/response_issue224_test.go
```diff
@@ -457,5 +457,5 @@ func TestEmptyResponsesAreInvalid(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, doc.ExternalDocs.Description, "See AsyncAPI example")
 	err = doc.Validate(context.Background())
-	require.EqualError(t, err, `invalid paths: the responses object MUST contain at least one response code`)
+	require.Contains(t, err.Error(), `the responses object MUST contain at least one response code`)
```
fenollp, line 460: Please keep `EqualError`. 
> `invalid paths: invalid path /pet: invalid operation POST: the responses object MUST contain at least one response code`

openapi3/loader_test.go
```diff
@@ -91,12 +91,12 @@ func TestResolveSchemaRef(t *testing.T) {
 }
 
 func TestResolveSchemaRefWithNullSchemaRef(t *testing.T) {
-	source := []byte(`{"openapi":"3.0.0","info":{"title":"MyAPI","version":"0.1","description":"An API"},"paths":{"/foo":{"post":{"requestBody":{"content":{"application/json":{"schema":null}}}}}}}`)
+	source := []byte(`{"openapi":"3.0.0","info":{"title":"MyAPI","version":"0.1","description":"An API"},"paths":{"/foo":{"post":{"summary":"Create Foo","requestBody":{"content":{"application/json":{"schema":null}}}}}}}`)
```
fenollp, line 94: I believe this can be reverted.

openapi3/example_validation_test.go
```diff
@@ -26,7 +26,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
             param1example:
               value: abcd
    `,
-			errContains: "invalid paths: param1example",
+			errContains: "param1example",
```
fenollp, line 29: I'd like to see the whole following be matched:
> invalid paths: invalid path /user: invalid operation POST: param1example: Field must be set to integer or not be present

openapi3/example_validation_test.go
```diff
@@ -64,7 +64,7 @@ func TestExamplesSchemaValidation(t *testing.T) {
         email: bad
         password: short
    `,
-			errContains: "invalid paths: BadUser",
+			errContains: "BadUser",
```
fenollp, line 67: > invalid paths: invalid path /user: invalid operation POST: BadUser

openapi3/loader_test.go
praneetloke: Sure. I went with:

> invalid paths: path /foo failed validation: operation POST: found unresolved ref: ""

(dropped the `for` from your suggestion)

openapi3/response_issue224_test.go
praneetloke: I would have except the sample spec used in this test actually has multiple operations that can fail. It's not deterministic either so I don't know which one will fail validation first.

openapi3/example_validation_test.go
praneetloke: Done. I've updated it.

openapi3/response_issue224_test.go
fenollp: Ah yes I have a branch that's working towards deterministic validation. I'll try and finish it soon.

praneetloke: Rebased and updated this test.

PR #607, opened 2022-09-20
fix: detects circular references that can't be handled at the moment to avoid infinite loops loading documents

TristanSpeakEasy:
fixes infinite loops caused by https://github.com/getkin/kin-openapi/issues/542 and https://github.com/getkin/kin-openapi/issues/570.

This particular implementation will detect recursion at a depth of 3 or more (tuned based on other recursive tests in the repo that could resolve recursions before this depth) and return an error if detected, at least allowing these situations to be detected and skipped instead of entering infinite recursion in the loading of a document


TristanSpeakEasy:
@fenollp I took the above approach to detecting infinite loops in the loading of documents as a dumb first pass at removing the problem.

I would imagine at some point the repo might want to address some of the valid cases this will be marking as errors, such as  references that are in properties, arrays, anyOf/OneOf/allOf etc that could be terminated by not being included at some point in the chain https://github.com/OAI/OpenAPI-Specification/issues/822 but that might require a much more sophisticated solution then this.

But at the bare minimum this removes an issues that is causing various projects that rely on this repo from entering into unrecoverable scenarios

TristanSpeakEasy:
I have found some other schemas in https://github.com/APIs-guru/openapi-directory that are still causing infinite loops even with the above fix, investigating them now

openapi3/loader.go
```diff
@@ -693,12 +694,18 @@ func (loader *Loader) resolveSchemaRef(doc *T, component *SchemaRef, documentPat
 			}
 			component.Value = &schema
 		} else {
+			if visitedLimit(visited, ref, 3) {
+				visited = append(visited, ref)
+				return fmt.Errorf("invalid schema: circular schema reference detected - %s", strings.Join(visited, " -> "))
```
fenollp, line 699: Not an invalid schema but a bug in this lib! (that, by the way, I think has to do with `loader.visitedSchema` keeping track of `*Schema` but is missing the pointers to schemas that are embedded in a schema such as `items`, `not`, `oneOf`, `allOf` and the one that appears in the minified test of 542 `anyOf`. I'd dig that way :) )

I'm down to merge this safeguarding PR, just modify the error message to not say the document is invalid but that there is a bug with ongoing investigation. Introduce a package-editable var if you prefer.

If `3` is too small I'd be fine with `300`. As long as it returns before exhausting the stack and looping for too long, this PR is an improvement!

I'm interested in bringing https://github.com/APIs-guru/openapi-directory/tree/main/APIs into the CI test suite by the way. Let's finish this and bring this in another PR if that's fine with you :)

TristanSpeakEasy: updated the error message and made it configurable.

Also 3 I found was the sweet spot for the depth ( I didn't find any schemas that succeeded loading with 4 that couldn't load with 3 for example).

fenollp (approved):

PR #608, opened 2022-09-20
Expose request/response validation options in the middleware Validator

sorintm:
Added new ValidatorOptions funtion to pass the validation options.

The context is that while ValidateRequest and ValidateResponse have configurable Options exposed through their input objects, the same Options cannot be passed when using the middleware/Validator approach.
The usecase that triggered the change was the validation of a response status code that is not in the API spec. ValidateResponse would check for it if the right option was passed, but there was no way to do it through the Validator.


fenollp:
Thanks!

Issue #613, opened 2022-09-23
Allow to add ValidationOptions to NewRouter()


yclybouw:
In routers/legacy/route, we have the function:
```go
func NewRouter(doc *openapi3.T) (routers.Router, error)
```
Which calls:
```go
doc.Validate(context.Background())
```

As `doc.Validate` accepts validation options (signature: `func (doc *T) Validate(ctx context.Context, opts ...ValidationOption) error`) I would like to modify the `NewRouter` function (or add a new one named `NewRouterWithValidationOptions)`, allowing to provide validation option:
```go
func NewRouter(doc *openapi3.T, opts ...ValidationOption) (routers.Router, error)
```

If this is fine for you, I can create a pull request for it.

PR #614, opened 2022-09-23
Allow validations options when creating legace Router

yclybouw:
My proposal to implement https://github.com/getkin/kin-openapi/issues/613

### DIFF
--- a/routers/legacy/router.go
+++ b/routers/legacy/router.go
@@ -58,14 +58,7 @@ type Router struct {
 //
 // If the given OpenAPIv3 document has servers, router will use them.
 // All operations of the document will be added to the router.
-func NewRouter(doc *openapi3.T) (routers.Router, error) {
-	return NewRouterWithValidationOptions(doc)
-}
-
-// NewRouterWithValidationOptions creates a new router with specific validation options before creation.
-//
-// Does the same thing as NewRouter, but when calling `doc.Validate()`, the validation options `opts` will be provided.
-func NewRouterWithValidationOptions(doc *openapi3.T, opts ...openapi3.ValidationOption) (routers.Router, error) {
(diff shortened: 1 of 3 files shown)


routers/legacy/router.go
```diff
@@ -59,7 +59,14 @@ type Router struct {
...
 // All operations of the document will be added to the router.
 func NewRouter(doc *openapi3.T) (routers.Router, error) {
-	if err := doc.Validate(context.Background()); err != nil {
+	return NewRouterWithValidationOptions(doc)
+}
+
+// NewRouterWithValidationOptions creates a new router with specific validation options before creation.
+//
+// Does the same thing as NewRouter, but when calling `doc.Validate()`, the validation options `opts` will be provided.
+func NewRouterWithValidationOptions(doc *openapi3.T, opts ...openapi3.ValidationOption) (routers.Router, error) {
```
fenollp, line 68: Good catch! Do feel free to overload today's implementation of `NewRouter` instead of introducing a new function.

Also do the same with `gorillamux`'s router please.

fenollp (changes requested):

routers/legacy/router.go
yclybouw: Hey, `gorillamux`'s  mentions "Assumes spec is .Validate()d" and indeed, `doc.Validate()` is not called.

What do you propose?

Note: I first tried to only change `func NewRouter` of `legacy`, but one test assumed both `legacy` and `gorillamux` to use the same signature.

fenollp: Ah well my bad sorry. Forget about `gorillamux`'s then.

Yes, just modify that test then. You'll just need to wrap `gorillamux.NewRouter` inside a func.

yclybouw: Done!

fenollp:
Thanks!

PR #620, opened 2022-10-01
fix #582 enhance format support

### LINKED ISSUE #582: Enhance format support
**Current regex used for `date` and `date-time` formats alllows date with day 00**

Extract `openapi3/schema_formats.go`:
```go
	// date
	DefineStringFormat("date", `^[0-9]{4}-(0[0-9]|10|11|12)-([0-2][0-9]|30|31)$`)

	// date-time
	DefineStringFormat("date-time", `^[0-9]{4}-(0[0-9]|10|11|12)-([0-2][0-9]|30|31)T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?(Z|(\+|-)[0-9]{2}:[0-9]{2})?$`)
```

Target formats would simply be:
```
"date":      `[0-9]{4}-(0[1-9]|10|11|12)-(0[1-9]|[12][0-9]|30|31)`,
"date-time": `[0-9]{4}-(0[1-9]|10|11|12)-(0[1-9]|[12][0-9]|30|31)T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?(Z|(\+|-)[0-9]{2}:[0-9]{2})?`,
```

**missing formats used in OpenAPI specs**

By default `uuid` format isn't enabled by default as well as IPv44/V6 and `hostname` is missing although this is defined in 3.1 spec.

Proposed changes in `init()` of `openapi3/schema_formats.go`:
```
DefineStringFormat(`uuid`, FormatOfStringForUUIDOfRFC4122)
DefineIPv4format()
DefineIPv6format()
DefineStringFormat( `hostname`, `[a-zA-Z0-9][a-zA-Z0-9-.]+[a-zA-Z0-9]` // https://www.rfc-editor.org/rfc/rfc1123#section-2.1
```

cboitel:


cboitel:
@fenollp ready for review

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
 	DefineStringFormat("byte", `(^$|^[a-zA-Z0-9+/\-_]*=*$)`)
 
 	// date
```
fenollp, line 90: Please add to comment: 
> defined as full-date in https://www.rfc-editor.org/rfc/rfc3339#section-5.6

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
...
-	DefineStringFormat("date", `^[0-9]{4}-(0[0-9]|10|11|12)-([0-2][0-9]|30|31)$`)
+	DefineStringFormat("date", `^[0-9]{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])$`)
 
 	// date-time
```
fenollp, line 93: Please add to comment: 
> defined as date-time in https://www.rfc-editor.org/rfc/rfc3339#section-5.6

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
...
-	DefineStringFormat("date-time", `^[0-9]{4}-(0[0-9]|10|11|12)-([0-2][0-9]|30|31)T[0-9]{2}:[0-9]{2}:[0-9]{2}(\.[0-9]+)?(Z|(\+|-)[0-9]{2}:[0-9]{2})?$`)
+	DefineStringFormat("date-time", `^[0-9]{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3])(:[0-5][0-9]){2}(\.[0-9]+)?(Z|(\+|-)[0-9]{2}:[0-9]{2})?$`)
 
+	// uuid
+	DefineStringFormat(`uuid`, FormatOfStringForUUIDOfRFC4122)
```
fenollp, line 97: This isn't in [OpenAPIv3.0](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.0.md#dataTypeFormat) so you need to find a way to define these only when the document's version is `>=3.1`.

I'd say we should keep the formats registry as a package var, just add an enum (int) field that can represent a default "AnyVersion" and a "SinceVersion3Point1" (or a better name). Then `Validate` and the `Visit*` funcs should consider these formats only if document version matches. Maybe pass the version info through context, maybe set an un-serializable pointer to the root of the doc in each struct...

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
...
+
+	// ipv4/v6
+	DefineIPv4Format()
```
fenollp, line 100: Not part of OpenAPIv3.0

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
...
+	DefineIPv6Format()
```
fenollp, line 101: Not part of OpenAPIv3.0

openapi3/schema_formats.go
```diff
@@ -88,11 +88,20 @@ func init() {
...
+
+	// hostname as defined in https://www.rfc-editor.org/rfc/rfc1123#section-2.1
+	DefineStringFormat(`hostname`, `^[a-zA-Z0-9][a-zA-Z0-9-.]+[a-zA-Z0-9]$`)
```
fenollp, line 104: Not part of OpenAPIv3.0

openapi3/schema_test.go
```diff
@@ -387,11 +408,15 @@ var schemaExamples = []schemaExample{
 			"2017-12-31T11:59:59\n",
 			"2017-12-31T11:59:59.+11:30",
 			"2017-12-31T11:59:59.Z",
+			"2017-12-00T11:59:59",
+			"2017-12-31T23:59:60",
```
fenollp, line 412: >    time-second     = 2DIGIT  ; 00-58, 00-59, 00-60 based on leap second rules

I don't think your code takes leap seconds into account?

fenollp (changes requested):

openapi3/schema_formats.go
cboitel: From a library user point of view, I don't see the point of having to specify OpenAPI version again if schema was loaded from an OpenAPI document. On the other side, library allows to create OpenAPI 3 Schemas out of an OpenAPI document (cf NewSchema) so adding a link to OpenAPI root document isn't a viable option.

Here is my proposal:
1. add type for OpenAPI 3 version (type int) with functions allowing to convert from/to strings
    > 3.0.0 will convert to 300, 3.1.x to 310, anything past x.y.z will be ignored
2. T structure will get a new un-serializable value matching the string read as OpenAPI version
     > If it fails, loading will also: there is currently no check on OpenAPI version string so this will somewhat add a breaking change but were are still in v0 so no harm yet
3. Schema structure will evolve to allow to attach to it an un-serializable OpenAPI Version:
     - it will default to 3.0 (nil value)
     - An WithOpenAPIVersion will be provided to allow to specify for other versions
4. After document is loading, we will scan the loaded document and call the WithOpenAPIVersion with version loaded from Document thus ensuring proper version is propagated to schemas
5. the package var for string formats will evolve:
        - we will able to define formats that would apply regardless of version (as of today => to maintain compatibility with earlier usage)
        - we will allow to define formats for specific versions of major or minor OpenAPI 
        - the internal implementation of the package var will evolve to allow quick access to format values for performance may be an issue when using library to validate requests or responses.

I will ensure existing library usage tests do not have to be modified to ensure compatibility with prior versions and will duplicate those tests for 3.1  (ensure 3.0 formats do work with 3.1) and add 3.1 specific ones. 

This will make a bulk of changes but will maintain library usage as earlier and adding capability to support new/modified formats in 3.x upcoming versions.

@fenollp : let me know if this sounds good to you and I also would prefer to close this PR and submit a new one since initial changes won't be kept as is.

fenollp: Sounds great to me. Let's open a new PR :) they're very cheap!

PR #622, opened 2022-10-05
Fix second level relative ref in property resolving

wtertius:
Before this fix the ref path was resolved incorrectly:

    openapi3/testdata/refInRefInProperty/components/errors.yaml:
        $ref: "../common-data-objects/problem-details-0.0.1.schema.json"

and I got error "no such file or directory".


d-sauer:
I got similar error on resolving spec folder as well: [611](https://github.com/getkin/kin-openapi/pull/611)
I think it’s slightly different situation.

Issue #632, opened 2022-10-12
Bug in the fix to 624 issue


orensolo:
Hi,

There is a bug in the fix to https://github.com/getkin/kin-openapi/issues/624 issue.
Using a special character in the parameter causes unmarshalling error.

Running the following:

```
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
	"github.com/getkin/kin-openapi/routers/gorillamux"
)

func main() {
	ctx := context.Background()
	loader := &openapi3.Loader{Context: ctx, IsExternalRefsAllowed: true}
	schema :=
		`
openapi: 3.0.0
info:
 version: 1.0.0
 title: Sample API
paths:
 /items:
  get:
   parameters:
    - description: "test non object"
      explode: true
      style: form
      in: query
      name: test
      required: false
      content:
       application/json:
         schema:
          anyOf:
           - type: string
           - type: integer
   description: Returns a list of stuff              
   responses:
    '200':
     description: Successful response`

	doc, err := loader.LoadFromData([]byte(schema))

	if err != nil {
		fmt.Println("failed to load schema. " + err.Error())
		return
	}
	// Validate document

	if err = doc.Validate(ctx); err != nil {
		fmt.Println("failed to validate schema. " + err.Error())
		return
	}

	router, _ := gorillamux.NewRouter(doc)
	httpReq, _ := http.NewRequest(http.MethodGet, `/items?test=test[1`, nil)  // <== note the input parameter contains [

	// Find route
	route, pathParams, _ := router.FindRoute(httpReq)

	// Validate request
	requestValidationInput := &openapi3filter.RequestValidationInput{
		Request:    httpReq,
		PathParams: pathParams,
		Route:      route,
	}
	err = openapi3filter.ValidateRequest(ctx, requestValidationInput)
	if err != nil {
		fmt.Println("failed to validate request. " + err.Error())
		return
	}

}
```

Returns:

> failed to validate request. parameter "test" in query has an error: error unmarshaling parameter "test"

The fix assumes that there cannot be special characters in the input parameters - which is a wrong assumption.
The correct fix is to use the **decodeStyledParameter** instead of decodeContentParameter. (with some relevant changes)

Please change the fix to the following fix:

In validate_request.go:

```
func ValidateParameter(ctx context.Context, input *RequestValidationInput, parameter *openapi3.Parameter) error {
[...]

Issue #624, opened 2022-10-06
query parameter validation unmarshalling error


orensolo:
Hi,

I am using   github.com/getkin/kin-openapi v0.103.0 (latest version).

 If the parameter content  is application/json,  then  kin-open api tries to unmarshel it (using json.unmarshal).
 But the parameter value need not be a json even if its content is application/json.

For example, when I run the following code :
```
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
	"github.com/getkin/kin-openapi/routers/gorillamux"
)

func main() {
	ctx := context.Background()
	loader := &openapi3.Loader{Context: ctx, IsExternalRefsAllowed: true}
	schema :=
		`
openapi: 3.0.0
info:
 version: 1.0.0
 title: Sample API
paths:
 /items:
  get:
   parameters:
    - description: "test non object"
      explode: true
      style: form
      in: query
      name: test
      required: false
      content:
       application/json:
         schema:
          anyOf:
           - type: string
           - type: integer
   description: Returns a list of stuff              
   responses:
    '200':
     description: Successful response`

	doc, err := loader.LoadFromData([]byte(schema))

	if err != nil {
		fmt.Println("failed to load schema. " + err.Error())
		return
	}
	// Validate document

	if err = doc.Validate(ctx); err != nil {
		fmt.Println("failed to validate schema. " + err.Error())
		return
	}

	router, _ := gorillamux.NewRouter(doc)
	httpReq, _ := http.NewRequest(http.MethodGet, `/items?test=test1`, nil)

	// Find route
	route, pathParams, _ := router.FindRoute(httpReq)

	// Validate request
	requestValidationInput := &openapi3filter.RequestValidationInput{
		Request:    httpReq,
		PathParams: pathParams,
		Route:      route,
	}
	err = openapi3filter.ValidateRequest(ctx, requestValidationInput)
	if err != nil {
		fmt.Println("failed to validate request. " + err.Error())
		return
	}

}
```

I got the following error:

> failed to validate request. parameter "test" in query has an error: error unmarshaling parameter "test"

But test=test1 is a valid parameter according to the schema.
Please run the above code to reproduce  the bug.

Thanks in advance,
Oren

PR #634, opened 2022-10-13
Actually  #624, thanks to @orensolo

fenollp:
cc https://github.com/getkin/kin-openapi/issues/632


orensolo (commented):
This fix has bugs. Please see my latest comment in https://github.com/getkin/kin-openapi/issues/632 for another suggested fix

PR #649, opened 2022-10-26
fix: yaml marshal output

### LINKED ISSUE #241: Marshaling to yaml doesn't inline ExtensionProps
I don't have a runable example handy as I'm basing this off of some results in my work, but I've pasted 2 differing yaml files and how they were generated from `Swagger` objects in these libraries.

### Expected
Results from `openapiv3.Swagger.MarshalJSON` and converting to yaml in https://editor.swagger.io/#
```
components:
  schemas:
    FooBar:
      properties:
        type_url:
          type: string
        value:
          format: byte
          type: string
      type: object
info:
  title: sample
  version: version not set
openapi: 3.0.3
paths: {}
```

### Actual (invalid openapi spec due to the presence of various struct fields in results)
- `extensionprops`
- `value`

Results from a `yaml.Marshal` of an `openapiv3.Swagger`
```
extensionprops: {}
openapi: 3.0.3
components:
  extensionprops: {}
  schemas:
    FooBar:
      ref: ""
      value:
        extensionprops: {}
        type: object
        properties:
          type_url:
            ref: ""
            value:
              extensionprops: {}
              type: string
          value:
            ref: ""
            value:
              extensionprops: {}
              type: string
              format: byte
info:
  extensionprops: {}
  title: sample
  version: version not set
paths: {}
```

### Questions
- Perhaps there's a more appropriate way to use this library and output the resulting yaml?

### How to fix

From what I can tell this libary needs to add additional `yaml` struct tags (`inline`) to the swagger objects to get this functioning correctly (`encoding/json` behaves this way by default). See https://godoc.org/gopkg.in/yaml.v2#Marshal

Adding `inline` to usages of the embedded `ExtensionProps` struct and to usages of the `Schema`, `SecuritySchema`, and `Example` ref types should cover a lot of this, I'm not sure what the full set is.

### Workaround
Something like the following can be done to work around this issue, by leveraging the correct output from `UnmarshalJSON` to get valid yaml output.

```
	jsonBuf, err := v3Spec.MarshalJSON()
	if err != nil {
		return err
	}
	tmp := map[string]interface{}{}
	err = json.Unmarshal(jsonBuf, &tmp)
	if err != nil {
		return err
	}
	return yaml.NewEncoder(w).Encode(tmp)
```

stevenh:
Fix the output when encoding the YAML by:
* Excluding ExtensionProps, the Extensions were already excluded resulting in empty extensionProps only.
* Add custom marshaler to all ref types so it doesn't output both values in the same way the JSON marshaler does.

Fixes #241

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.yml text eol=lf
--- a/openapi3/components.go
+++ b/openapi3/components.go
@@ -12,7 +12,7 @@ import (
 // Components is specified by OpenAPI/Swagger standard version 3.
 // See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#componentsObject
 type Components struct {
-	ExtensionProps
+	ExtensionProps `json:"-" yaml:"-"`
 
 	Schemas         Schemas         `json:"schemas,omitempty" yaml:"schemas,omitempty"`
 	Parameters      ParametersMap   `json:"parameters,omitempty" yaml:"parameters,omitempty"`
--- a/openapi3/discriminator.go
+++ b/openapi3/discriminator.go
@@ -9,7 +9,7 @@ import (
 // Discriminator is specified by OpenAPI/Swagger standard version 3.
 // See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#discriminatorObject
 type Discriminator struct {
-	ExtensionProps
+	ExtensionProps `json:"-" yaml:"-"`
 
 	PropertyName string            `json:"propertyName" yaml:"propertyName"`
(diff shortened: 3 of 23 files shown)


fenollp (changes requested):
Hi! You're probably hitting https://github.com/getkin/kin-openapi/issues/241 
I encourage you to use the same yaml package mentioned there.

If that's not suitable for you and this PR fixes your issues, please add a test that uses the yaml package you are using (or just not the one mentioned in the issue but gopkgin's one directly) showing explicitly what serialization returns.

stevenh:
Yes looks very similar, test added to prove it fixes that issue.

fenollp (approved):

PR #653, opened 2022-11-04
only encode defaults for supported content type

hutm:
there is a problem with openapi spec when the following are true:
- content type is multipart/form-data 
- defaults are specified


fenollp:
Could you add a regression test for this please?

fenollp (changes requested):
Please add a regression test.

PR #654, opened 2022-11-05
fix panic slice out of range error #652

nk2ge5k:
Fixes panic from the issue #652, that occurs when trying to load specification from file that contains reference to the file insed of parent directory.

### DIFF
--- a/openapi3/issue652_test.go
+++ b/openapi3/issue652_test.go
@@ -3,8 +3,9 @@ package openapi3_test
 import (
 	"testing"
 
-	"github.com/getkin/kin-openapi/openapi3"
 	"github.com/stretchr/testify/require"
+
+	"github.com/getkin/kin-openapi/openapi3"
 )
 
 func TestIssue652(t *testing.T) {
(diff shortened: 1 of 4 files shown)


fenollp (approved):
LGTM!
And if that's not too much to ask for, could you add a test that produces that error message?
Thanks

nk2ge5k:
Reproducing error message here #659

fenollp:
Ah sorry no I meant another test that would hit this line: https://github.com/getkin/kin-openapi/pull/654/files#diff-035b395ae81006bbe6cd7144c686238113762fccdb051ddaef5d507560039ee6R747
```go
			foundPath, rerr := loader.getResolvedRefPath(ref, &resolved, documentPath, componentPath)
			if rerr != nil {
				return fmt.Errorf("failed to resolve file from reference %q: %w", ref, rerr)
			}
```

nk2ge5k:
Actually, I cannot reproduce this error message - for such error message to return we need to make one of this calls
```go
	filepath.Rel("..", ".")
	filepath.Rel{"..", "a")
	filepath.Rel{"../..", "..")
	filepath.Rel{"a", "/a")
	filepath.Rel{"/a", "a")
```
which is impossible because every path is checked before `getResolvedRefPath` call.

It is possible that I am just not seeing some possibility.

PR #660, opened 2022-11-08
Fixed recurive reference resolving when property referencies local co…

derbylock:
Fixed recurive reference resolving when property referencies local component schema type, which referencies json schema in another dir, in which property referencies another json in some non-root (not in the same dir as the specification with local reference) dir json schema.

Please, look at added test cases for more details about the fixed issue.


fenollp (changes requested):
LGTM! Just you can remove the two `definitions.json` files. They are unused if I'm not mistaken.

derbylock:
> LGTM! Just you can remove the two `definitions.json` files. They are unused if I'm not mistaken.

Thanks, done

Issue #625, opened 2022-10-06
arrays of objects as query parameters cause a crush


orensolo:
Hi,

I am using the latest kin-openapi  (v0.103.0)
This issue is similar to https://github.com/getkin/kin-openapi/issues/619.
Query parameters that are defined as arrays of objects (i.e not primitives) causing kin-openapi to crush with panic (schema has non primitive type ...)

I prepared a code that you can run in order to reproduce the bug, please try it out to verify the bug:

```
package main

import (
	"context"
	"fmt"
	"net/http"

	"github.com/getkin/kin-openapi/openapi3"
	"github.com/getkin/kin-openapi/openapi3filter"
	"github.com/getkin/kin-openapi/routers/gorillamux"
)

func main() {
	ctx := context.Background()
	loader := &openapi3.Loader{Context: ctx, IsExternalRefsAllowed: true}
	schema :=
		`
openapi: 3.0.0
info:
 version: 1.0.0
 title: Sample API
paths:
 /items:
  get:
   parameters:
    - description: "test object"
      explode: false
      in: query
      name: test
      required: false
      schema:
       items:
        properties:
         name:
          type: string
        type: object
       type: array
   description: Returns a list of stuff              
   responses:
    '200':
     description: Successful response`

	doc, err := loader.LoadFromData([]byte(schema))

	if err != nil {
		fmt.Println("failed to load schema. " + err.Error())
		return
	}
	// Validate document

	if err = doc.Validate(ctx); err != nil {
		fmt.Println("failed to validate schema. " + err.Error())
		return
	}

	router, _ := gorillamux.NewRouter(doc)
	httpReq, _ := http.NewRequest(http.MethodGet, "/items?test=3", nil)

	// Find route
	route, pathParams, _ := router.FindRoute(httpReq)

	// Validate request
	requestValidationInput := &openapi3filter.RequestValidationInput{
		Request:    httpReq,
		PathParams: pathParams,
		Route:      route,
	}
	err = openapi3filter.ValidateRequest(ctx, requestValidationInput)
	if err != nil {
		fmt.Println("failed to validate request. " + err.Error())
		return
	}

}

```

When this code is executed, the following panic happen:

```
panic: schema has non primitive type "object"

goroutine 1 [running]:
github.com/getkin/kin-openapi/openapi3filter.parsePrimitive({0x919bd9, 0x1}, 0xc00000ea80)
[...]

PR #664, opened 2022-11-15
Issue 625 array crush

orensolo:
This pull request solves the issue in https://github.com/getkin/kin-openapi/issues/625

### DIFF
--- a/openapi3filter/issue625_test.go
+++ b/openapi3filter/issue625_test.go
@@ -1,4 +1,4 @@
-package openapi3_test
+package openapi3filter_test
 
 import (
 	"net/http"
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -526,8 +526,105 @@ func (d *urlValuesDecoder) DecodeArray(param string, sm *openapi3.SerializationM
 		}
 		values = strings.Split(values[0], delim)
 	}
-	val, err := parseArray(values, schema)
+	val, err := d.parseArray(values, sm, schema)
 	return val, ok, err
+}
+
+// parseArray returns an array that contains items from a raw array.
+// Every item is parsed as a primitive value.
+// The function returns an error when an error happened while parse array's items.
+func (d *urlValuesDecoder) parseArray(raw []string, sm *openapi3.SerializationMethod, schemaRef *openapi3.SchemaRef) ([]interface{}, error) {
+	var value []interface{}
+
+	for i, v := range raw {
+		item, err := d.parseValue(v, schemaRef.Value.Items)
+		if err != nil {
+			if v, ok := err.(*ParseError); ok {
+				return nil, &ParseError{path: []interface{}{i}, Cause: v}
+			}
+			return nil, fmt.Errorf("item %d: %w", i, err)
+		}
+
+		// If the items are nil, then the array is nil. There shouldn't be case where some values are actual primitive
+		// values and some are nil values.
+		if item == nil {
+			return nil, nil
+		}
+		value = append(value, item)
+	}
+	return value, nil
+}
+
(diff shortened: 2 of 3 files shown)


openapi3filter/req_resp_decoder.go
```diff
@@ -526,10 +526,107 @@ func (d *urlValuesDecoder) DecodeArray(param string, sm *openapi3.SerializationM
...
+		var err error
+		for _, sr := range schema.Value.AllOf {
+			var f bool
+			value, err = d.parseValue(v, sr)
+			found = found || f
+			if value == nil || err != nil {
+				break
+			}
+		}
+		return value, err
```
fenollp, line 572: Returning here implies allOf and siblings are mutually exclusive. They aren't.

openapi3filter/issue625_test.go
```diff
@@ -0,0 +1,242 @@
...
+			route, pathParams, err := router.FindRoute(httpReq)
+			require.NoError(t, err)
+
+			requestValidationInput := &openapi3filter.RequestValidationInput{
+				Request:    httpReq,
+				PathParams: pathParams,
+				Route:      route,
+			}
+			err = openapi3filter.ValidateRequest(ctx, requestValidationInput)
+			require.Equal(t, testcase.isErr, err != nil)
```
fenollp, line 238: Let's try matching the error we expect a bit more.

openapi3filter/issue625_test.go
```diff
@@ -0,0 +1,242 @@
...
+		{
+			name:  "successful any item array",
+			spec:  anyItemSpec,
+			req:   "/items?test=3",
+			isErr: false,
+		},
+		{
+			name:  "successful any item object array",
+			spec:  anyItemSpec,
+			req:   `/items?test={"name": "test1"}`,
```
fenollp, line 159: This doesn't seem to be url-encoded. Are you sure this is a valid URL?

openapi3filter/issue625_test.go
```diff
@@ -0,0 +1,242 @@
...
+    description: Returns a list of stuff
+    parameters:
+    - description: test object
+      explode: false
+      in: query
+      name: test
+      required: false
+      schema:
+        type: array
+        items: {}   ###
```
fenollp, line 33: Please make it obvious the parts of these blobs you reuse and the parts you duplicate. 
Maybe introduce a function that takes the missing body as argument and simply returns header + body + footer.

fenollp (changes requested):

openapi3filter/issue625_test.go
orensolo: fixed. I use "Contains" to verify the error

openapi3filter/issue625_test.go
orensolo: fixed. I using strings.ReplaceAll

openapi3filter/issue625_test.go
orensolo: You are right. This is not a valid URL. 
Lets focus the fix on array with items of polymorphism type  (i.e anyOf, oneOf, allOf ). This is the needed use case.
I removed the test case and relevant code

openapi3filter/req_resp_decoder.go
orensolo: I follow the same logic as in decodeValue function (https://github.com/getkin/kin-openapi/blob/871e029d62a4d47239d11006431407e08edb1259/openapi3filter/req_resp_decoder.go#L262)

The purpose of this function is to decode the value according to the schema type.
And not to validate it.

If the decoding is successful, then that's it. There is no need to further process the value.
If the value does not match the rest of the siblings, then it will fail in the validation

orensolo:
@fenollp I fixed all your comments. Can you please review again and merge if all OK?

The-Murdock:
@fenollp Hi, can you help us to close this PR?

PR #669, opened 2022-11-17
fix: example ref to schemas cause error

dominicqi:
if example ref to '#/components/schemas/Pet'   ,will cause load error 
`: bad data in "#/components/schemas/Pet"% `

Fixed by check if ref is schemas

### DIFF
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -683,6 +683,24 @@ func (loader *Loader) resolveResponseRef(doc *T, component *ResponseRef, documen
 		sort.Strings(examples)
 		for _, name := range examples {
 			example := contentType.Examples[name]
+			if exampleRef := example.Ref; exampleRef != "" {
+				parse, err := url.Parse(exampleRef)
+				if err != nil {
+					return err
+				}
+				if refType := strings.Split(parse.Fragment[1:], "/")[1]; refType == "schemas" {
+					var resolved SchemaRef
+					_, err := loader.resolveComponent(doc, exampleRef, parse, &resolved)
+					if err != nil {
+						return err
+					}
+					contentType.Examples[name] = &ExampleRef{
+						Ref:   exampleRef,
+						Value: NewExample(resolved.Value),
+					}
+					continue
+				}
+			}
 			if err := loader.resolveExampleRef(doc, example, documentPath); err != nil {
 				return err
 			}
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -683,23 +683,22 @@ func (loader *Loader) resolveResponseRef(doc *T, component *ResponseRef, documen
 		sort.Strings(examples)
 		for _, name := range examples {
 			example := contentType.Examples[name]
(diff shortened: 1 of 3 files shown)


openapi3/loader_exmaple_schema_ref_test.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func TestLoadSpecWithExampleRefToSchemas(t *testing.T) {
+	loader := NewLoader()
+	loader.IsExternalRefsAllowed = true
+	api, err := loader.LoadFromFile("testdata/example-schemas-ref.json")
+	exampleValue := api.Paths["/pet"].Put.Responses.Get(400).Value.Content.Get("*/*").Examples["200"].Value.Value
+	schemaRef := api.Components.Schemas["Pet"].Value
+	require.Equal(t, exampleValue, schemaRef)
+	require.NoError(t, err)
```
fenollp, line 16: Check err where it is set.

Also, use `doc` instead of `api` as is done in the other tests.

Also, ensure document is valid per `doc.Validate(ctx)` as is done in the other tests.

openapi3/loader_exmaple_schema_ref_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLoadSpecWithExampleRefToSchemas(t *testing.T) {
+	loader := NewLoader()
+	loader.IsExternalRefsAllowed = true
```
fenollp, line 11: This does not seem required for this test.

openapi3/loader.go
```diff
@@ -683,6 +683,24 @@ func (loader *Loader) resolveResponseRef(doc *T, component *ResponseRef, documen
 		sort.Strings(examples)
 		for _, name := range examples {
 			example := contentType.Examples[name]
+			if exampleRef := example.Ref; exampleRef != "" {
+				parse, err := url.Parse(exampleRef)
+				if err != nil {
+					return err
+				}
+				if refType := strings.Split(parse.Fragment[1:], "/")[1]; refType == "schemas" {
```
fenollp, line 691: I see jobs fail around here. I'm just skimming here but maybe try using the `ReadFromURIFunc`s instead of unrolling your own.

fenollp (changes requested):

openapi3/loader_exmaple_schema_ref_test.go
dominicqi: fixed

openapi3/loader_exmaple_schema_ref_test.go
dominicqi: yes,removed

openapi3/loader.go
dominicqi: sorry, i not found `ReadFromURIFunc` can do this, i fixed by check the ref prefix

PR #672, opened 2022-11-18
Update content length after replacing request body

borodiychuk:
Update content length after replacing the request body when setting default values. Otherwise, the request gets inconsistent, and further operation on it, like proxying it, causes an error.


fenollp:
Hey thanks for this. Please add a regression test. Then I'll merge this.

borodiychuk:
@fenollp thanks, added tests.

fenollp:
Thanks!

Issue #674, opened 2022-11-21
errors from `FormatCallback` are not thrown, so it is impossible to determine their type


micronull:
Errors defined with `openapi3.DefineStringFormatCallback()` are forcibly converted into a string, which makes it impossible to determine their type when validating a query with `openapi3filter.ValidateRequest()`.

PR #675, opened 2022-11-21
fix: wrap the error that came back from the callback (#674)

micronull:
see #674 and TestFormatCallback_WrapError


fenollp:
Thanks!

Issue #678, opened 2022-11-22
`openapi3.SchemaError` message customize


micronull:
Hi. I need to be able to change the text of the error message in the schema. 

The functionality has been implemented. I will attach it as a separate PR.

PR #679, opened 2022-11-22
fix: openapi3.SchemaError message customize (#678)

micronull:
See #678.
You can, of course, try to use i18n, but then we won't be able to substitute the values we need from the schema fields, such as title.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1315,29 +1315,31 @@ func (schema *Schema) visitJSONString(settings *schemaValidationSettings, value 
 	}
 
 	// "format"
-	var formatErr string
+	var formatStrErr string
+	var formatErr error
 	if format := schema.Format; format != "" {
 		if f, ok := SchemaStringFormats[format]; ok {
 			switch {
 			case f.regexp != nil && f.callback == nil:
 				if cp := f.regexp; !cp.MatchString(value) {
-					formatErr = fmt.Sprintf(`string doesn't match the format %q (regular expression "%s")`, format, cp.String())
+					formatStrErr = fmt.Sprintf(`string doesn't match the format %q (regular expression "%s")`, format, cp.String())
 				}
 			case f.regexp == nil && f.callback != nil:
 				if err := f.callback(value); err != nil {
-					formatErr = err.Error()
+					formatErr = err
 				}
 			default:
-				formatErr = fmt.Sprintf("corrupted entry %q in SchemaStringFormats", format)
+				formatStrErr = fmt.Sprintf("corrupted entry %q in SchemaStringFormats", format)
 			}
 		}
 	}
-	if formatErr != "" {
+	if formatStrErr != "" || formatErr != nil {
 		err := &SchemaError{
 			Value:                 value,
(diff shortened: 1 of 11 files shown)


openapi3/schema_validation_settings.go
```diff
@@ -50,6 +52,11 @@ func DefaultsSet(f func()) SchemaValidationOption {
 	return func(s *schemaValidationSettings) { s.defaultsSet = f }
 }
 
+// SetSchemaErrorCustomMessage allows to override the schema error message.
+func SetSchemaErrorCustomMessage(f func(err *SchemaError) string) SchemaValidationOption {
```
fenollp, line 56: `SetSchemaErrorMessageCustomizer`

openapi3/schema_test.go
```diff
@@ -1089,6 +1090,33 @@ func TestSchemaErrors(t *testing.T) {
 	}
 }
 
+func TestSchemaError_CustomMessage(t *testing.T) {
```
fenollp, line 1093: This should be an Example.

openapi3/schema.go
```diff
@@ -1687,6 +1714,12 @@ func (err *SchemaError) JSONPointer() []string {
 }
 
 func (err *SchemaError) Error() string {
+	if err.customizeMessageError != nil {
+		if msg := err.customizeMessageError(err); msg != "" {
```
fenollp, line 1718: The documentation of this option should mention that returning an empty string falls back to the upstream `Error()` impl.

openapi3filter/validate_request_test.go
```diff
@@ -215,3 +216,53 @@ components:
...
+
+	route, pathParams, err := router.FindRoute(req)
+	require.NoError(t, err)
+
+	validationInput := &RequestValidationInput{
+		Request:    req,
+		PathParams: pathParams,
+		Route:      route,
+		Options: &Options{
+			CustomSchemaErrorFunc: func(err *openapi3.SchemaError) string {
```
fenollp, line 260: Use the function calls to set the options slice instead of the struct.

fenollp (changes requested):
The ask and approach seem valid to me, let's fix some naming / docs and we'll be ready to go!

openapi3filter/validate_request_test.go
```diff
@@ -215,3 +216,53 @@ components:
 		})
 	}
 }
+
+func TestValidateRequest_CustomSchemaErrorMessage(t *testing.T) {
```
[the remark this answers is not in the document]
micronull: turned

openapi3/schema_test.go
micronull: made an example of.

openapi3/schema.go
micronull: added

openapi3filter/validate_request_test.go
micronull: made by calling the `WithCustomSchemaErrorFunc` function

fenollp (approved):

Issue #681, opened 2022-11-25
lost format error path


micronull:
Line format errors no longer contain "Error at".

```yml
components:
  schemas:
    Something:
      type: object
      properties:
        ip:
          type: string
          format: ipv4
```

```go
err = doc.Components.Schemas["Something"].Value.VisitJSON(map[string]interface{}{
		`ip`: `123.0.0.11111`,
	})
```

Error: `Not an IP address`
Must be: `Error at "/ip": Not an IP address`

PR #682, opened 2022-11-25
fix: error path is lost (#681)

micronull:
see #681 and `TestReversePathInMessageSchemaError`.

### DIFF
--- a/openapi3/schema_formats_test.go
+++ b/openapi3/schema_formats_test.go
@@ -75,3 +75,33 @@ func TestFormatCallback_WrapError(t *testing.T) {
 
 	delete(SchemaStringFormats, "foobar")
 }
+
+func TestReversePathInMessageSchemaError(t *testing.T) {
+	DefineIPv4Format()
+
+	SchemaErrorDetailsDisabled = true
+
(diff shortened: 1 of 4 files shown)


openapi3/schema_formats_test.go
```diff
@@ -75,3 +75,33 @@ func TestFormatCallback_WrapError(t *testing.T) {
 
 	delete(SchemaStringFormats, "foobar")
 }
+
+func TestReversePathInMessageSchemaError(t *testing.T) {
+	DefineIPv4Format()
+
+	SchemaErrorDetailsDisabled = true
```
fenollp, line 82: From a quick glance I believe tests fail due to tests being run in parallel and multiple times. This is to detect race conditions. 

Here, there this obvious racey code. So I recommend to move this test to another package (as is done with a few other tests). Something with `_test` IIRC. 

Otherwise LGTM!

micronull: The tests will run in parallel if we explicitly specify this with `t.Parallel()`: https://pkg.go.dev/testing#T.Parallel

PR #690, opened 2022-12-01
fix: setting defaults for oneOf and anyOf

tomato0111:
When setting default values for oneOf or anyOf, a temporary value (tempValue) is used to prevent original value from being modified. However, tempValue should to be reset on the next iteration otherwise default value set from previous iteration can modify tempValue which can lead to failure.

The fix is to move the deepcopy inside the loop, which will reset tempValue to the original input.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -933,10 +933,6 @@ func (schema *Schema) visitSetOperations(settings *schemaValidationSettings, val
 			matchedOneOfIdx  = 0
 			tempValue        = value
 		)
-		// make a deep copy to protect origin value from being injected default value that defined in mismatched oneOf schema
-		if settings.asreq || settings.asrep {
-			tempValue = deepcopy.Copy(value)
-		}
 		for idx, item := range v {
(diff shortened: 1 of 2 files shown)


fenollp:
Hi! Please make it more obvious how this patch fixes what it claims.

How does moving the copy inside the loop help?

Can you add minimal a regression test?

Thanks

tomato0111:
tempValue should to be reset on the next iteration otherwise default value set from previous iteration can modify the input which can lead to failure.

Thanks!

fenollp:
Cool thanks, that makes sense.

Changes to `openapi3filter/validate_set_default_test.go` don't seem to matter for the fix you're providing. Can you explain why you pushed them? 

Please add a regression test.

Thanks

tomato0111:
Setting "additionalProperties" to false in the test schema actually exposes the issue. I can add another test.

tomato0111:
I have added a new test case. Let me know if this is sufficient.

Thanks!

PR #694, opened 2022-12-04
draft: fix: handle security requirements as OpenAPI 3 specifies (I think!!!)

gward:
From https://swagger.io/specification/#security-requirement-object:

> Security Requirement Objects that contain multiple schemes require
> that all schemes MUST be satisfied for a request to be authorized.
> This enables support for scenarios where multiple query parameters or
> HTTP headers are required to convey security information.

I'm fairly sure there was a bug in ValidateSecurityRequirements(),
which I've fixed. And I added a test case to TestValidateRequest() to
show the problem and prove that I've fixed it.

Unfortunately, this also breaks TestAnySecurityRequirementMet(), and
I'm not yet sure why. Not sure if the test is incorrect or if my fix
is incorrect.

### DIFF
--- a/openapi3filter/validate_request.go
+++ b/openapi3filter/validate_request.go
@@ -315,8 +315,13 @@ func ValidateSecurityRequirements(ctx context.Context, input *RequestValidationI
 			errs = append(errs, err)
 			continue
 		}
+	}
+	if len(errs) == 0 {
+		// all security requirements have been met: success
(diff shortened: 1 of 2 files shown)


gward:
Ooops, my mistake. I have read (and re-read) the OpenAPI spec more carefully, and I think I was wrong about this.

PR #698, opened 2022-12-12
openapi3: patch YAML serialization of dates

cosnicolaou:


fenollp:
I'm down with this but aren't there more places this should be handled? Such as the `examples` field?

Also, have you looked at a way to fix this upstream and what were your findings?

Thanks!

cosnicolaou:
It's clearly an ugly fix and I'm open to suggestions for a betterone! It can affect other fields for sure, but I'm not sure it matters for those, then again, I'm not an experienced openapi user. This particular problem seems very specific to openapi, that is, the spec contains a literal date which gets turned into a date-time by json/yaml marshal/unmarshaling which is considered invalid since the format field is a date. Another option would be to change the validation to treat the 00:00:00Z suffix as not being present, or to issue a warning rather than an error. An alternative is to edit the original input specification to use a string rather than a literal, but that is also far from ideal since it's tedious to find all of the occurrences that need to be fixed.

When you say 'aren't there more places' - do you have any that you're concerned about?

cosnicolaou:
As an aside, I just wrote a small package to allow 'walking' of an openapi3.T, https://github.com/cosnicolaou/openapi. If you wish, it's probably better to integrate the 'walk' portion (not necessarily the transformations) into your core package.

fenollp:
If there's a way to disable date parsing in the yaml package, that'd be a cleaner solution to me. Can you enquire? Or at least open an issue there that links back here?

Also please add a comment that points to this issue in your `schema.go` code.
Also these can be just one line:
```go
			eg = strings.TrimSuffix(eg, "T00:00:00Z")
			schema.Example = eg
// =>
			schema.Example = strings.TrimSuffix(eg, "T00:00:00Z")
```


WRT your walker package: I'm down to integrate this in here especially if it can be used in v3tov2. I'll review any PR you send my way :)

openapi3/schema.go
```diff
@@ -180,7 +181,14 @@ func (schema *Schema) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON sets Schema to a copy of data.
 func (schema *Schema) UnmarshalJSON(data []byte) error {
-	return jsoninfo.UnmarshalStrictStruct(data, schema)
+	err := jsoninfo.UnmarshalStrictStruct(data, schema)
+	if schema.Format == "date" {
+		if eg, ok := schema.Example.(string); ok {
+			eg = strings.TrimSuffix(eg, "T00:00:00Z")
+			schema.Example = eg
```
fenollp, line 188:
```suggestion
			schema.Example = strings.TrimSuffix(eg, "T00:00:00Z")
```

openapi3/schema.go
```diff
@@ -180,7 +181,14 @@ func (schema *Schema) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON sets Schema to a copy of data.
 func (schema *Schema) UnmarshalJSON(data []byte) error {
-	return jsoninfo.UnmarshalStrictStruct(data, schema)
+	err := jsoninfo.UnmarshalStrictStruct(data, schema)
+	if schema.Format == "date" {
+		if eg, ok := schema.Example.(string); ok {
```
fenollp, line 186: please add a comment that points to the issue

fenollp (changes requested):

PR #699, opened 2022-12-12
openapi3filter: Include schema ref or title in response body validation errors

slessard:
Currently schema validation error messages are extremely verbose. For example, when validating a response body against a schema, specifically when `additionalProperties` is set to `false` for each additional property the entire response body and the entire schema for that response is dumped in the error message. This can result in loads and loads and loads of log messages. 

A partial solution to the above mentioned issue is to use the recently introduced ability to add an error message customizer function. With the error message customizer it is possible to remove the response body and the response scheme from the error message, but then the error message is a bit out of context. The error message might look something like this:

```
response body doesn't match the schema: property "cooldown_period" is unsupported | property "end_time" is unsupported | property "extra_field_compute_profile" is unsupported | property "extra_field_compute_profile_response" is unsupported | property "group_id" is unsupported | property "group_name" is unsupported | property "initially_suspended" is unsupported | property "instance" is unsupported | property "instance_name" is unsupported | property "instance_type" is unsupported | property "max_instance_count" is unsupported | property "min_instance_count" is unsupported | property "name" is unsupported | property "profile_id" is unsupported | property "scaling_policy" is unsupported | property "start_time" is unsupported
```

Where it lacks context is that you cannot easily tell which schema is being evaluated. Sure, fancier logging in the consuming application could do a better job of associating this error message with a request URL thus giving a clue as to which response schema is being evaluated. But a better, more user-friendly solution would be to include a short reference to the schema right there in the error message, like this...

```
response body doesn't match the schema #/components/schemas/ComputeProfileResponse: property "cooldown_period" is unsupported | property "end_time" is unsupported | property "extra_field_compute_profile" is unsupported | property "extra_field_compute_profile_response" is unsupported | property "group_id" is unsupported | property "group_name" is unsupported | property "initially_suspended" is unsupported | property "instance" is unsupported | property "instance_name" is unsupported | property "instance_type" is unsupported | property "max_instance_count" is unsupported | property "min_instance_count" is unsupported | property "name" is unsupported | property "profile_id" is unsupported | property "scaling_policy" is unsupported | property "start_time" is unsupported
```

With this change a reference to the schema is included in the headline error message thus allowing one to further customize the error message by omitting the schema and yet still have the error message make sense.

Not all schemas are referenced in the way my my project does, i.e. `#/components/schemas/ComputeProfileResponse`. Some schemas are inline. To that end I wrote a function to get an identifier for the schema. The function starts by trying to get the schema reference as shown above. If that doesn't exist the function then tries to get the schema's title. If that also doesn't exist the function then tries to get the schema's description. If there still is no value the function gives up and returns empty string.


openapi3filter/validate_request.go
```diff
@@ -273,10 +273,11 @@ func ValidateRequestBody(ctx context.Context, input *RequestValidationInput, req
 
 	// Validate JSON with the schema
 	if err := contentType.Schema.Value.VisitJSON(value, opts...); err != nil {
+		schemaId := getSchemaIdentifier(contentType.Schema)
 		return &RequestError{
 			Input:       input,
 			RequestBody: requestBody,
-			Reason:      "doesn't match the schema",
+			Reason:      fmt.Sprintf("doesn't match the schema %s", schemaId),
```
fenollp, line 280: Let's handle empty `schemaId` so that that space doesn't get produced.

Also let's drop the the.

openapi3filter/validate_response.go
```diff
@@ -159,11 +160,29 @@ func ValidateResponse(ctx context.Context, input *ResponseValidationInput) error
...
+	var id string
+
+	if schema != nil {
+		id = schema.Ref
+	}
+	if strings.TrimSpace(id) == "" && schema.Value != nil {
+		id = schema.Value.Title
+	}
+	if strings.TrimSpace(id) == "" && schema.Value != nil {
+		id = schema.Value.Description
```
fenollp, line 184: > description - [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.

This field can be many lines long. Aren't you overfitting to your problem space here? I don't think this is a good idea.

openapi3filter/validate_response.go
```diff
@@ -159,11 +160,29 @@ func ValidateResponse(ctx context.Context, input *ResponseValidationInput) error
 
 	// Validate data with the schema.
 	if err := contentType.Schema.Value.VisitJSON(value, append(opts, openapi3.VisitAsResponse())...); err != nil {
+		schemaId := getSchemaIdentifier(contentType.Schema)
 		return &ResponseError{
 			Input:  input,
-			Reason: "response body doesn't match the schema",
+			Reason: fmt.Sprintf("response body doesn't match the schema %s", schemaId),
```
fenollp, line 166: space + the the

openapi3filter/validate_response.go
```diff
@@ -159,11 +160,29 @@ func ValidateResponse(ctx context.Context, input *ResponseValidationInput) error
...
 			Err:    err,
+			// TODO: add an error message customizer function similar to SchemaError
```
fenollp, line 168: Open an issue instead of leaving that comment please.

fenollp (changes requested):

openapi3filter/validate_response.go
```diff
@@ -159,11 +160,29 @@ func ValidateResponse(ctx context.Context, input *ResponseValidationInput) error
...
 	return nil
 }
+
+func getSchemaIdentifier(schema *openapi3.SchemaRef) string {
+	var id string
+
+	if schema != nil {
+		id = schema.Ref
+	}
+	if strings.TrimSpace(id) == "" && schema.Value != nil {
```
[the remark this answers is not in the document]
slessard: I don't understand this comment. Can you clarify, please?

openapi3filter/validate_request.go
slessard: fixed

openapi3filter/validate_response.go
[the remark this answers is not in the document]
slessard: fixed in upcoming commit

openapi3filter/validate_response.go
slessard: I had that some concern but erred on the side of caution. Now that I see you share that concern I agree description should be removed. This is fixed in upcoming commit

openapi3filter/validate_response.go
slessard: I opened issue #701

fenollp (approved):

Issue #638, opened 2022-10-18
Reference resolve issue for 3GPP Rel.16 definitions


ShouheiNishi:
We are trying to generate code with the oapi-codegen from the 3GPP Rel.16 openapi definition .yaml file, but the reference resolution fails in the kin-openapi loader and the file cannot be loaded.

We use these files. All files are got from 3GPP webpage.
But encode of TS29544_Nspaf_SecuredPacket.yaml is fixed by hand.
The file from which the import is started is TS29571_CommonData.yaml.

TS28623_GenericNrm.yaml
TS28623_comDefs.yaml
TS29122_AsSessionWithQoS.yaml
TS29122_ChargeableParty.yaml
TS29122_CommonData.yaml
TS29122_CpProvisioning.yaml
TS29122_DeviceTriggering.yaml
TS29122_ECRControl.yaml
TS29122_GMDviaMBMSbyMB2.yaml
TS29122_GMDviaMBMSbyxMB.yaml
TS29122_MonitoringEvent.yaml
TS29122_MsisdnLessMoSms.yaml
TS29122_NIDD.yaml
TS29122_NpConfiguration.yaml
TS29122_PfdManagement.yaml
TS29122_RacsParameterProvisioning.yaml
TS29122_ReportingNetworkStatus.yaml
TS29122_ResourceManagementOfBdt.yaml
TS29502_Nsmf_PDUSession.yaml
TS29503_Nudm_EE.yaml
TS29503_Nudm_MT.yaml
TS29503_Nudm_NIDDAU.yaml
TS29503_Nudm_PP.yaml
TS29503_Nudm_SDM.yaml
TS29503_Nudm_UEAU.yaml
TS29503_Nudm_UECM.yaml
TS29505_Subscription_Data.yaml
TS29507_Npcf_AMPolicyControl.yaml
TS29508_Nsmf_EventExposure.yaml
TS29509_Nausf_SoRProtection.yaml
TS29509_Nausf_UEAuthentication.yaml
TS29509_Nausf_UPUProtection.yaml
TS29510_Nnrf_AccessToken.yaml
TS29510_Nnrf_Bootstrapping.yaml
TS29510_Nnrf_NFDiscovery.yaml
TS29510_Nnrf_NFManagement.yaml
TS29512_Npcf_SMPolicyControl.yaml
TS29514_Npcf_PolicyAuthorization.yaml
TS29517_Naf_EventExposure.yaml
TS29518_Namf_Communication.yaml
[...]

PR #700, opened 2022-12-13
Bugfix/issue638

ShouheiNishi:
Fix for https://github.com/getkin/kin-openapi/issues/638.

### DIFF
--- a/openapi3/issue638_test.go
+++ b/openapi3/issue638_test.go
@@ -1,0 +1,21 @@
+package openapi3
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue638(t *testing.T) {
+	for i := 0; i < 50; i++ {
+		loader := NewLoader()
+		loader.IsExternalRefsAllowed = true
+		// This path affects the occurrence of the issue #638.
+		// ../openapi3/testdata/issue638/test1.yaml : reproduce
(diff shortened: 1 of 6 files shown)


openapi3/issue638_test.go
```diff
@@ -0,0 +1,17 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue638(t *testing.T) {
+	for i := 0; i < 50; i++ {
+		loader := NewLoader()
+		loader.IsExternalRefsAllowed = true
+		doc, err := loader.LoadFromFile("../openapi3/testdata/issue638/test1.yaml")
```
fenollp, line 13: Please add a not on this line about the path expression being important for this regression test.

fenollp (changes requested):
Do you think `openapi3/issue570_test.go` can be updated as well or is that still triggering the circular ref error?

Amazing work here, thank you so much!
LGTM. Just a simple request then this gets merged!

cc @TristanSpeakEasy this should interest you. Can you take a look and see if that fixes some of the issues you've been having with this lib? I'll run tests on my side as well but I figure the orderedmap PR should take care of the rest.

openapi3/issue638_test.go
ShouheiNishi: I added a comment about path expression.

Issue #703, opened 2022-12-14
openapi3filter.ValidateRequest() does not update GetBody()


ShouheiNishi:
The openapi3filter.ValidateRequest() function sometimes change request body but does not update GetBody() member of request.

We try this test code.
```
diff --git a/openapi3filter/validate_request_test.go b/openapi3filter/validate_request_test.go
index 450ee59..8da550c 100644
--- a/openapi3filter/validate_request_test.go
+++ b/openapi3filter/validate_request_test.go
@@ -212,6 +212,12 @@ components:
                        assert.Equal(t, contentLen, bodySize, "expect ContentLength %d to equal body size %d", contentLen, bodySize)
                        bodyModified := originalBodySize != bodySize
                        assert.Equal(t, bodyModified, tc.expectedModification, "expect request body modification happened: %t, expected %t", bodyModified, tc.expectedModification)
+
+                       validationInput.Request.Body, err = validationInput.Request.GetBody()
+                       assert.NoError(t, err, "unable to re-generate body by GetBody(): %v", err)
+                       body2, err := io.ReadAll(validationInput.Request.Body)
+                       assert.NoError(t, err, "unable to read request body: %v", err)
+                       assert.Equal(t, body, body2, "body by GetBody() is not matched")
                })
        }
 }
```

Generate these error.
```
--- FAIL: TestValidateRequest (0.00s)
    --- FAIL: TestValidateRequest/Valid_request_without_certain_fields (0.00s)
        validate_request_test.go:220:
                Error Trace:    /home/nishi/free5gc-deb/free5gc/lib/kin-openapi/openapi3filter/validate_request_test.go:220
                Error:          Not equal:
                                expected: []byte{0x7b, 0x22, 0x63, 0x61, 0x74, 0x65, 0x67, 0x6f, 0x72, 0x79, 0x22, 0x3a, 0x22, 0x53, 0x77, 0x65, 0x65, 0x74, 0x73, 0x22, 0x2c, 0x22, 0x73, 0x75, 0x62, 0x43, 0x61, 0x74, 0x65, 0x67, 0x6f, 0x72, 0x79, 0x22, 0x3a, 0x22, 0x43, 0x68, 0x6f, 0x63, 0x6f, 0x6c, 0x61, 0x74, 0x65, 0x22, 0x7d}
[...]

PR #704, opened 2022-12-14
Use and update GetBody() member of request

ShouheiNishi:
Fix for https://github.com/getkin/kin-openapi/issues/703.


fenollp:
Thanks!

Issue #707, opened 2022-12-15
parameters with default values


orensolo:
Hi,

I am using kinopenapi version 0.110.0, 
If the schema contains parameters with a default value, and those parameters are missing in the request, then kinopenapi adds the
default values to the request after the validation.

This is not the desired behavior (in some cases) when the validation takes place on the client side.
This is because of the following reasons:
1. The payload is unnecessarily larger.
2. The server may respond differently if the default value exists.

So, the already existing `SkipSettingDefaults` option should be used for query parameters as well.
I will create a pull request to handle this issue.

Thanks,
Oren

PR #708, opened 2022-12-15
openapi3filter: use option to skip setting defaults on validation

orensolo:
this pull request solves https://github.com/getkin/kin-openapi/issues/707 issue

### DIFF
--- a/openapi3filter/validate_request.go
+++ b/openapi3filter/validate_request.go
@@ -66,7 +66,7 @@ func ValidateRequest(ctx context.Context, input *RequestValidationInput) (err er
 			}
 		}
 
-		if err = ValidateParameter(ctx, input, parameter); err != nil && !options.MultiError {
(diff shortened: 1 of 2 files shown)


openapi3filter/validate_request.go
```diff
@@ -137,7 +138,7 @@ func ValidateParameter(ctx context.Context, input *RequestValidationInput, param
 	}
 
 	// Set default value if needed
-	if value == nil && schema != nil && schema.Default != nil {
+	if !skipSettingDefaults && value == nil && schema != nil && schema.Default != nil {
```
fenollp, line 141: Use `input.Options.SkipSettingDefaults` directly :)

fenollp (changes requested):

openapi3filter/validate_request.go
orensolo: @fenollp fixed. Please merge if all is OK.

Issue #724, opened 2022-12-29
OAS 3 String format validation not working in path


zekth:
Seems like the string format validation does not seem to work in `path`  

```yaml
paths:
  /uuid-pattern/{param}:
    get:
      parameters:
        - in: path
          name: param
          required: true
          schema:
            type: string
            pattern: '^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
            minLength: 36
            maxLength: 36
      responses:
        '200':
          description: success
  /uuid-format/{param}:
    get:
      parameters:
        - in: path
          name: param
          required: true
          schema:
            type: string
            format: uuid
      responses:
        '200':
          description: success
```

In the previous example `/uuid-pattern/foobar` returns properly an error, which can be casted as multierror with 2 errors but the `uuid-format/foobar` passes the validation.

Seems like the format is handled in the codebase: https://github.com/getkin/kin-openapi/blob/46e0df8ae84cbb51b7c672a0bb8ebb6995d28c7a/openapi3/schema.go#L373-L378

Is there a miss somewhere?

fenollp:
Hi! Thanks for opening a pull request,
As you noted there, the solution is to `DefineStringFormat("uuid", FormatOfStringForUUIDOfRFC4122)` before validation.

The OpenAPI spec does not define the uuid format and so handling validation of it out of scope of this lib. Feel free to opt-in to the uuid validation mode though.

zekth:
> Hi! Thanks for opening a pull request, As you noted there, the solution is to `DefineStringFormat("uuid", FormatOfStringForUUIDOfRFC4122)` before validation.
> 
> The OpenAPI spec does not define the uuid format and so handling validation of it out of scope of this lib. Feel free to opt-in to the uuid validation mode though.

So why having email and not uuid then?

fenollp:
Indeed you're right, this is a mistake as old as the first commit! https://github.com/getkin/kin-openapi/commit/b8556159d2dcf88924f438bd364e3dc9d0e10380#diff-90e14c30c1c7403690a666a976d821851020aa362bf984c3d7a6e8e0b2abcca4R22

Thanks for spotting this. I'd accept a PR that removes that line and mentions this change in the README for the next minor version bump.

fenollp:
[...]

PR #725, opened 2022-12-29
fix(schema): define uuid

zekth:
Fix: https://github.com/getkin/kin-openapi/issues/724

Seems like the definition of `uuid` was missing. In the 3.0.0 spec it's written :

```
Formats such as "email", "uuid", and so on, MAY be used even though undefined by this specification
```

So because we already have `email`, we can add `uuid`


fenollp:
Working as intended :) See comment on linked issue.

PR #727, opened 2022-12-30
openapi3: remove email string format

zekth:
addresses https://github.com/getkin/kin-openapi/issues/724 follow-up


fenollp:
Thanks!

Issue #513, opened 2022-03-23
How to catch invalid OpenAPI v3 spec


JeanChristopheMorinPerso:
Hello! I'm trying to find how I can make this library to return an error when loading an invalid spec.

Basically, I have an invalid v3 spec that looks like this:
```yaml
openapi: "3.0.3"
info:
  title: 'My app'
  version: 1.0.0
  description: 'An API'

paths:
  /v1/operation:
    delete:
      summary: Delete something
      responses:
        200:
          description: Success
        default:
          description: '* **400** - Bad Request'
          # Notice here schema is invalid. It should instead be:
          # content:
          #   application/json:
          #     schema:
          #       $ref: '#/components/schemas/Error'
          schema:
            $ref: '#/components/schemas/Error'
components:
  schemas:
    Error:
      type: object
      description: An error response body.
      properties:
        message:
          description: A detailed message describing the error.
          type: string
```

I load that spec using the `openapi3.Loader`:

```go
loader := openapi3.NewLoader()
loader.IsExternalRefsAllowed = true

schema, err := loader.LoadFromFile("/path/to/spec.yaml")
if err != nil {
	fmt.Println("Failed to load swagger_template.yaml:", err.Error())
	os.Exit(1)
}

// Not used in my current example but that's what my code does
schema.InternalizeRefs(context.Background(), nil)

err = schema.Validate(context.Background())
if err != nil {
	fmt.Println("Validation failed:", err.Error())
}
```

I would expect that either `LoadFromFile` or `Validate` return an error saying that it's not a valid spec or something, but it's not.

Is there a way to tell the library to return an error in such cases?

Thanks!

timo-klarshift:
You could use something like this: https://github.com/xeipuuv/gojsonschema#meta-schema-validation

You can use any jsonschema validator to validate the openapi specs you have against the official openapi schema ...

JeanChristopheMorinPerso:
Thanks @fenollp !

PR #728, opened 2022-12-31
Disallow unexpected fields in validation and drop `jsoninfo` package

### LINKED ISSUE #37: Let's remove package 'jsoninfo'
Historical background: The package `jsoninfo` was created with the goal of having elegant serialization.
 
The fact that `jsoninfo` resides in this repository is just a historical accident. The code is just an internal implementation detail to this project. So I created [github.com/getkin/happyjson](https://github.com/getkin/happyjson) and refactored it a bit.

If there are no objections, I'll replace `jsoninfo` with the external package `happyjson`. Any feedback about how _kin-openapi_ implements serialization internally is welcome.

fenollp:
Fixes https://github.com/getkin/kin-openapi/issues/513
Fixes #37 : gets rid of ./jsoninfo package


lzap:
Hey, during update I am getting the following problem: https://github.com/getkin/kin-openapi/issues/771 Could this be related to this refactoring?

radwaretaltr:
Hi, after upgrading from v0.112.0 we aren't getting descriptive unmarshal errors anymore. It's probably related to the removal of jsoninfo.

For example, originally for invalid type errors we got the hierarchy of the field for which the unmarshalling failed:

image

After the upgrade we only get the last part: "json: cannot unmarshal bool into GO value of type string".

fenollp:
@radwaretaltr If you are looking to get that behavior back (which may be due to the removal of package `jsoninfo` or maybe something else) I encourage you to look into the unrolled `UnmarshalJSON` implementations. There one should easily be able to wrap errors as they happen. I'll review any PR :)

PR #730, opened 2023-01-03
openapi3filter: RegisterBodyDecoder for application/zip

k2tzumi:
Hi.

Added `application/zip` to body decoder.

I have verified that it is possible to read in zip format using [archive/zip](https://pkg.go.dev/archive/zip).


openapi3filter/req_resp_decoder.go
```diff
@@ -1013,6 +1015,7 @@ func init() {
 	RegisterBodyDecoder("application/x-www-form-urlencoded", urlencodedBodyDecoder)
 	RegisterBodyDecoder("multipart/form-data", multipartBodyDecoder)
 	RegisterBodyDecoder("application/octet-stream", FileBodyDecoder)
+	RegisterBodyDecoder("application/zip", ZipFileBodyDecoder)
```
fenollp, line 1018: Please sort this block of lines

openapi3filter/req_resp_decoder.go
```diff
@@ -1221,3 +1224,44 @@ func FileBodyDecoder(body io.Reader, header http.Header, schema *openapi3.Schema
...
+
+// ZipFileBodyDecoder is a body decoder that decodes a zip file body to a string.
+func ZipFileBodyDecoder(body io.Reader, header http.Header, schema *openapi3.SchemaRef, encFn EncodingFn) (interface{}, error) {
+	buff := bytes.NewBuffer([]byte{})
+	size, err := io.Copy(buff, body)
+	if err != nil {
+		return nil, err
+	}
+
+	zr, err := zip.NewReader(bytes.NewReader(buff.Bytes()), size)
```
fenollp, line 1236: Why copy? Isn't it possible to read from `body` directly?

openapi3filter/req_resp_decoder.go
```diff
@@ -1221,3 +1224,44 @@ func FileBodyDecoder(body io.Reader, header http.Header, schema *openapi3.Schema
...
+	const bufferSize = 256
+	var content []byte
+	buffer := make([]byte, bufferSize)
+
+	for _, f := range zr.File {
+		rc, err := f.Open()
+		if err != nil {
+			return nil, err
+		}
+		defer rc.Close()
```
fenollp, line 1250: Wrap the contents of this for loop inside a `func(){ ... ; return }()` as `defer` happens only at the next `return` statement.

openapi3filter/req_resp_decoder.go
```diff
@@ -1221,3 +1224,44 @@ func FileBodyDecoder(body io.Reader, header http.Header, schema *openapi3.Schema
...
+	}
+
+	zr, err := zip.NewReader(bytes.NewReader(buff.Bytes()), size)
+	if err != nil {
+		return nil, err
+	}
+
+	const bufferSize = 256
+	var content []byte
+	buffer := make([]byte, bufferSize)
```
fenollp, line 1243: How about 
```suggestion
	content := make([]byte, 0, bufferSize*len(zr.File))
	buffer := make([]byte, 0, bufferSize)
```

fenollp (changes requested):

openapi3filter/req_resp_decoder.go
k2tzumi: Yes, it is not possible to read directly from `body`.
It must be converted to `io.ReaderAt`.
https://pkg.go.dev/archive/zip#NewReader

openapi3filter/zip_file_upload_test.go
```diff
@@ -0,0 +1,125 @@
...
+      responses:
+        '200':
+          description: Created
+`
+
+	loader := openapi3.NewLoader()
+	doc, err := loader.LoadFromData([]byte(spec))
+	if err != nil {
+		require.NoError(t, err)
+	}
```
fenollp, line 49: Like so:
```suggestion
	require.NoError(t, err)
```

openapi3filter/req_resp_decoder.go
fenollp: Or don't use `defer`

k2tzumi: Does it look like this?
https://github.com/getkin/kin-openapi/pull/730/commits/5361b9b8b7cd549e34b13409f6dd919079e89b3a

openapi3filter/req_resp_decoder.go
```diff
@@ -1243,24 +1243,28 @@ func ZipFileBodyDecoder(body io.Reader, header http.Header, schema *openapi3.Sch
...
-				content = append(content, buffer...)
-			}
-			if err == io.EOF {
-				break
-			}
+		func() {
+			rc, err := f.Open()
 			if err != nil {
-				return nil, err
+				panic(err)
```
fenollp, line 1249: Just return an error, use a `func() error`.

fenollp (changes requested):

fenollp (approved):

PR #737, opened 2023-01-08
Remove value from SchemaError Reason field

### LINKED ISSUE #735: Make SchemaError.Reason field safe to log
### Background:

In regulatory environments, it is important to be mindful of the information that is logged and returned to users, as it may contain sensitive information such as private details, payment information, or secrets.

When a `SchemaError` occurs, the default error message produced by the `Error()` method includes the `Value` field. While it is possible to customize the error message using the `WithCustomSchemaErrorFunc` function, this requires the user to manually handle each possible schema validation error to produce a meaningful message.

### Suggested Solution:

I suggest making the `SchemaError.Reason` field "safe" by guaranteeing that it will not contain the original `Value`. This will allow users to simply return `SchemaError.Reason` in their custom error function, rather than having to craft the entire message for each possible scenario.

Upon review, most messages in the `Reason` field are already "safe" and do not include the original `Value`. However, there are a few messages that should be revised to remove the `Value`:

```go
"value \"%g\" must be an integer" => "value must be an integer"
`string %q doesn't match the regular expression "%s"` => `string doesn't match the regular expression %q`
"value %q is not one of the allowed values" => "value must be one of the allowed values %v"
"value of discriminator property %q is not a string: %v" => "value of discriminator property %q is not a string"
"discriminator property %q has invalid value: %q" => "discriminator property %q has invalid value"
```

ori-shalom:
Resolves #735


fenollp:
LGTM. What about values such as these though? https://github.com/getkin/kin-openapi/blob/e7d649f3f7d6ddbaaaed74a7d2f819a82118aab4/openapi3filter/unpack_errors_test.go

ori-shalom:
> LGTM. What about values such as these though? https://github.com/getkin/kin-openapi/blob/e7d649f3f7d6ddbaaaed74a7d2f819a82118aab4/openapi3filter/unpack_errors_test.go

Can you clarify what values you are referring to?
Are you referring to `RequestError`?
Looks like `RequestError.Reason` never gets user input:
image

fenollp:
I'm asking if you'd also like to hide the values that appear in the `// Output:` part of the example. They come from SchemaError if I'm not mistaken and they may contain sensible data.

ori-shalom:
IMO it is still reasonable to get the value in the error message for some use cases.
For example, APIs that do not involve sensitive inputs or for the dev environment when not handling real user inputs.

This is why I didn't propose changing completely the default output of `SchemaError.Error()` but instead add the guarantee that the `Reason` field won't include the original value.

This doesn't change the current behavior too much but provides a mechanism that will allow hiding value in the error message more easily with a `CustomSchemaErrorFunc` that utilizes `SchemaError.Reason`.

ori-shalom:
I guess a flag can be added to `Options` to decide whether to print the Value or not to control this printing in the default `SchemaError.Error()`:

https://github.com/getkin/kin-openapi/blob/e7d649f3f7d6ddbaaaed74a7d2f819a82118aab4/openapi3/schema.go#L2020-L2023

The name of the option field can be: `[Include/Exclude/Hide/Show][Value/UserInputs]InErrors`
I'm open to any suggestions you have.

If we should go in this direction, I'll prefer doing that in another PR as I don't see it coupled with current changes.

fenollp (changes requested):
Alright. I'm not sure this should require an option but I'd like you to document that the Reason field holds this property please.

ori-shalom:
> I guess a flag can be added to `Options` to decide whether to print the Value or not to control this printing in the default `SchemaError.Error()`:
> 
> https://github.com/getkin/kin-openapi/blob/e7d649f3f7d6ddbaaaed74a7d2f819a82118aab4/openapi3/schema.go#L2020-L2023
> 
> The name of the option field can be: `[Include/Exclude/Hide/Show][Value/UserInputs]InErrors` I'm open to any suggestions you have.
> 
> If we should go in this direction, I'll prefer doing that in another PR as I don't see it coupled with current changes.

OK, so I somehow missed there was already a `SchemaErrorDetailsDisabled` flag.

I added documentation for the `SchemaErrorDetailsDisabled` flag + `CustomSchemaErrorFunc` + `Reason` field guaranteed not to include the original value.

ori-shalom:
@fenollp added the documentation as requested.
If nothing else needs to be done, I'll be happy if this could be merged.

fenollp:
LGTM but do ask that you add a comment to the field https://github.com/getkin/kin-openapi/blob/1e5c86eeb045499f248a8393b8a28409d561b286/openapi3/schema.go#L1939 that reflects the requirement that you introduce.
Thanks

Issue #733, opened 2023-01-06
Integer validation issue


ShouheiNishi:
Since the numeric value is converted to float64 during JSON validation in body in function `openapi3filter.ValidateRequestBody’, the precision of integer value in updated request body is lost, and the range check is not performed correctly.

You can see this issue in the test code below.

```
package openapi3filter_test

import (
        "bytes"
        "context"
        "encoding/json"
        "math"
        "math/big"
        "net/http"
        "testing"

        "github.com/getkin/kin-openapi/openapi3"
        "github.com/getkin/kin-openapi/openapi3filter"
        "github.com/getkin/kin-openapi/routers/gorillamux"
        "github.com/stretchr/testify/assert"
        "github.com/stretchr/testify/require"
)

func TestIntMax(t *testing.T) {
        spec := `
openapi: 3.0.0
info:
  version: 1.0.0
  title: test large integer value
paths:
  /test:
    post:
      requestBody:
        content:
          application/json:
            schema:
              type: object
              properties:
                testInteger:
                  type: integer
                  format: int64
                testDefault:
                  type: boolean
                  default: false
      responses:
        '200':
          description: Successful response
`[1:]

        loader := openapi3.NewLoader()

        doc, err := loader.LoadFromData([]byte(spec))
        require.NoError(t, err)

        err = doc.Validate(loader.Context)
        require.NoError(t, err)

        router, err := gorillamux.NewRouter(doc)
        require.NoError(t, err)

        testOne := func(value *big.Int, pass bool) {
                valueString := value.String()

                req, err := http.NewRequest(http.MethodPost, "/test", bytes.NewReader([]byte(`{"testInteger":`+valueString+`}`)))
                require.NoError(t, err)
                req.Header.Set("Content-Type", "application/json")

                route, pathParams, err := router.FindRoute(req)
                require.NoError(t, err)

                err = openapi3filter.ValidateRequest(
                        context.Background(),
                        &openapi3filter.RequestValidationInput{
                                Request:    req,
                                PathParams: pathParams,
                                Route:      route,
                        })
                if pass {
                        require.NoError(t, err)

                        dec := json.NewDecoder(req.Body)
                        dec.UseNumber()
                        var jsonAfter map[string]interface{}
                        err = dec.Decode(&jsonAfter)
                        require.NoError(t, err)

                        valueAfter := jsonAfter["testInteger"]
                        require.IsType(t, json.Number(""), valueAfter)
[...]

PR #738, opened 2023-01-10
Specify UseNumber() in the JSON decoder during JSON validation

ShouheiNishi:
Partial fix for https://github.com/getkin/kin-openapi/issues/733.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1077,6 +1077,19 @@ func (schema *Schema) visitJSON(settings *schemaValidationSettings, value interf
 	switch value := value.(type) {
 	case bool:
 		return schema.visitJSONBoolean(settings, value)
+	case json.Number:
+		valueFloat64, err := value.Float64()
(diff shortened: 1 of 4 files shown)


fenollp:
Hey, thanks for your work!
Since this is only a partial fix, I'm wondering what the plan forward should be. 
From https://github.com/golang/go/issues/6384#issue-51285102:
> (There are other issues as you get to very large values, since float64 can't encode a
whole int64... not sure what or if anything can be done about that.)

Indeed!
Should we be turning all numbers `Schema` holds to `big.Int` or `Float` or `Rat`?

ShouheiNishi:
> Indeed! Should we be turning all numbers `Schema` holds to `big.Int` or `Float` or `Rat`?
Since I cannot be sure of the necessity of this change and there is a concern that DoS will be established when a long numeric string is entered into the parser in “math/big”, I think that the change is unnecessary for the time being.

PR #744, opened 2023-01-17
openapi3: support format conditional on `doc.openapi` version

cboitel:
Implements string format improvements in issue #582 with openapi minor version support.

* You can now define a string formats available for some minor version
* by default, string formats apply to all minor versions (>= 3.0)
* filters do enforce string formats according to document's OpenAPI version

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -100,8 +100,12 @@ type Schema struct{ ... }
     func NewArraySchema() *Schema
     func NewBoolSchema() *Schema
     func NewBytesSchema() *Schema
+    func NewDateSchema() *Schema
     func NewDateTimeSchema() *Schema
     func NewFloat64Schema() *Schema
+    func NewHostnameSchema() *Schema
+    func NewIPv4Schema() *Schema
+    func NewIPv6Schema() *Schema
     func NewInt32Schema() *Schema
     func NewInt64Schema() *Schema
     func NewIntegerSchema() *Schema
@@ -125,6 +129,7 @@ type SchemaValidationOption func(*schemaValidationSettings)
     func EnableFormatValidation() SchemaValidationOption
     func FailFast() SchemaValidationOption
     func MultiErrors() SchemaValidationOption
+    func SetOpenAPIMinorVersion(minorVersion uint64) SchemaValidationOption
     func SetSchemaErrorMessageCustomizer(f func(err *SchemaError) string) SchemaValidationOption
     func VisitAsRequest() SchemaValidationOption
     func VisitAsResponse() SchemaValidationOption
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -9,17 +9,20 @@ var CircularReferenceError = "kin-openapi bug found: circular schema reference n
 var DefaultReadFromURI = URIMapCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))
 var ErrURINotSupported = errors.New("unsupported URI")
 var IdentifierRegExp = regexp.MustCompile(identifierPattern)
-var SchemaStringFormats = make(map[string]Format, 4)
+var SchemaStringFormats = make(map[string]*Format, 4)
 func BoolPtr(value bool) *bool
 func DefaultRefNameResolver(ref string) string
 func DefineIPv4Format()
 func DefineIPv6Format()
-func DefineStringFormat(name string, pattern string)
-func DefineStringFormatCallback(name string, callback FormatCallback)
+func DefineStringFormat(name string, pattern string, options ...SchemaFormatOption)
+func DefineStringFormatCallback(name string, callback FormatCallback, options ...SchemaFormatOption)
 func Float64Ptr(value float64) *float64
 func Int64Ptr(value int64) *int64
 func ReadFromFile(loader *Loader, location *url.URL) ([]byte, error)
 func RegisterArrayUniqueItemsChecker(fn SliceUniqueItemsChecker)
+func RestoreDefaultStringFormats()
+func RestoreStringFormats(formatToRestore map[string]*Format)
+func SaveStringFormats(map[string]*Format) map[string]*Format
 func Uint64Ptr(value uint64) *uint64
 func ValidateIdentifier(value string) error
 func WithValidationOptions(ctx context.Context, opts ...ValidationOption) context.Context
@@ -108,6 +111,9 @@ type Schema struct{ ... }
     func NewStringSchema() *Schema
     func NewUUIDSchema() *Schema
 type SchemaError struct{ ... }
+type SchemaFormatOption func(options *SchemaFormatOptions)
+    func FromOpenAPIMinorVersion(fromMinorVersion uint64) SchemaFormatOption
+type SchemaFormatOptions struct{ ... }
 type SchemaRef struct{ ... }
     func NewSchemaRef(ref string, value *Schema) *SchemaRef
 type SchemaRefs []*SchemaRef
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -3,6 +3,7 @@ const TypeArray = "array" ...
 const FormatOfStringForUUIDOfRFC4122 = `^(?:[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}|00000000-0000-0000-0000-000000000000)$` ...
 const SerializationSimple = "simple" ...
 var SchemaErrorDetailsDisabled = false ...
+var ErrInvalidVersion ...
 var CircularReferenceCounter = 3
 var CircularReferenceError = "kin-openapi bug found: circular schema reference not handled"
 var DefaultReadFromURI = URIMapCache(ReadFromURIs(ReadFromHTTP(http.DefaultClient), ReadFromFile))
@@ -147,9 +148,12 @@ type ValidationOption func(options *ValidationOptions)
     func DisableSchemaDefaultsValidation() ValidationOption
     func DisableSchemaFormatValidation() ValidationOption
     func DisableSchemaPatternValidation() ValidationOption
+    func DisableVersionValidation() ValidationOption
     func EnableExamplesValidation() ValidationOption
     func EnableSchemaDefaultsValidation() ValidationOption
     func EnableSchemaFormatValidation() ValidationOption
     func EnableSchemaPatternValidation() ValidationOption
+    func EnableVersionValidation() ValidationOption
 type ValidationOptions struct{ ... }
+type Version string
 type XML struct{ ... }
--- a/openapi3/schema_formats.go
+++ b/openapi3/schema_formats.go
@@ -21,26 +21,99 @@ type FormatCallback func(value string) error
 
 // Format represents a format validator registered by either DefineStringFormat or DefineStringFormatCallback
 type Format struct {
+	versionedFormats []*versionedFormat
+}
+
+type versionedFormat struct {
 	regexp   *regexp.Regexp
 	callback FormatCallback
 }
 
+func (format *Format) add(minMinorVersion uint64, vFormat *versionedFormat) {
+	if format != nil {
+		if format.versionedFormats == nil {
+			format.versionedFormats = make([]*versionedFormat, minMinorVersion+1)
+			format.versionedFormats[minMinorVersion] = vFormat
+		} else {
+			numVersionedFormats := uint64(len(format.versionedFormats))
+			if minMinorVersion >= numVersionedFormats {
+				// grow array
+				lastValue := format.versionedFormats[numVersionedFormats-1]
+				additionalEntries := make([]*versionedFormat, minMinorVersion+1-numVersionedFormats)
+				if lastValue != nil {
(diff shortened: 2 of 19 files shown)


openapi3filter/validate_request.go
```diff
@@ -182,6 +182,9 @@
 	if options.customSchemaErrorFunc != nil {
 		opts = append(opts, openapi3.SetSchemaErrorMessageCustomizer(options.customSchemaErrorFunc))
 	}
+	if input.Route != nil {
+		opts = append(opts, openapi3.SetOpenAPIMinorVersion(uint(input.Route.Spec.OpenAPI.Minor())))
```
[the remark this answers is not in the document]
cboitel: fixed

openapi3filter/validate_request.go
```diff
@@ -283,6 +286,9 @@
...
```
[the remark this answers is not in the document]
cboitel: fixed

openapi3filter/validate_response.go
```diff
@@ -145,7 +148,7 @@
 	}
 
 	// Validate data with the schema.
-	if err := contentType.Schema.Value.VisitJSON(value, append(opts, openapi3.VisitAsResponse())...); err != nil {
+	if err := contentType.Schema.Value.VisitJSON(value, append(opts, openapi3.VisitAsResponse(), openapi3.SetOpenAPIMinorVersion(uint(route.Spec.OpenAPI.Minor())))...); err != nil {
```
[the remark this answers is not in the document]
cboitel: fixed

cboitel:
@fenollp finally made up with implementing a string format support based on OpenAPI minor version

fenollp (changes requested):

openapi3/version_test.go
```diff
@@ -0,0 +1,101 @@
...
+			"3.299.0",
+		},
+		ExpectedValid: true,
+	},
+	{
+		Title: "Invalid",
+		Values: []Version{
+			"3.0.1.2",
+			"3.0",
+			"3",
```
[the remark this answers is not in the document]
cboitel: 1. In OpenAPI 3.0.x, openapi field is a required string and must be the semantic version number of the specification version => see https://spec.openapis.org/oas/v3.0.3#oasVersion
2. In OpenAPI 3.1.x, field is still required and must match the x.y.z : see  https://spec.openapis.org/oas/v3.1.0#openapi-object

If you use swagger's editor (current and new one), you get an error if openapi field is '3' or '3.0'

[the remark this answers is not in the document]
cboitel: @fenollp any news on this ?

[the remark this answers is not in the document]
cboitel: @fenollp expecting some more info ?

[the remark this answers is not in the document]
fenollp: Hiya, sorry I was under the impression that `major` and `major.minor` were valid semver syntaxes but I was wrong => https://github.com/semver/semver/issues/237

openapi3/schema_formats.go
```diff
@@ -21,26 +21,111 @@ type FormatCallback func(value string) error
...
+		return
+	}
+	var newFormat Format
+	newFormat.add(minMinorVersion, vFormat, override)
+	SchemaStringFormats[name] = &newFormat
+}
+
+// DefineStringFormatStartingWithOpenAPIMinorVersion defines a new regexp pattern for a given format to apply for a minor version of OpenAPI 3
+// override allows to enforce regex usage to all subsequent minor versions of OpenAPI
+func DefineStringFormatStartingWithOpenAPIMinorVersion(name string, fromMinorVersion uint64, pattern string, override bool) {
```
fenollp, line 110: Why `uint64`? `uint8` or even `uint16` is large enough. 

I like the idea but I'd rather have a simpler API:
* a shorter method such as the existing `DefineStringFormat` (extend with variadic args)
* make `override=true` the only behavior. API users will simply make multiple calls to cut specific ranges
* turn `fromMinorVersion` into a `float` (`f32` if for all `Y`, `3.Y` can be expressed and I believe it's a yes) so users can see `DefineStringFormat(name, pattern, 3.1)`. The call should panic on non-nil non-floats and compare minors with a `0.4` epsilon.

openapi3/schema_formats_test.go
```diff
@@ -98,10 +91,319 @@ components:
...
+	assert.ErrorContains(t, err, `Error at "/ip"`)
 
-	delete(SchemaStringFormats, "ipv4")
 	SchemaErrorDetailsDisabled = false
 }
+
+func TestStringFormatsStartingWithOpenAPIMinorVersion(t *testing.T) {
+	DefineStringFormatStartingWithOpenAPIMinorVersion("test", 0, "test0", false)
+	for i := uint64(0); i < 10; i++ {
+		if assert.Contains(t, SchemaStringFormats, "test") &&
```
fenollp, line 105: Please prefer the `require` package over `assert`. This makes for fewer nesting.

openapi3/schema_validation_settings.go
```diff
@@ -18,6 +18,8 @@ type schemaValidationSettings struct {
 	defaultsSet         func()
 
 	customizeMessageError func(err *SchemaError) string
+
+	openapiMinorVersion uint64 // defaults to 0 (3.0.z)
```
fenollp, line 22: `uint8` is enough

openapi3/schema_validation_settings.go
```diff
@@ -65,3 +67,8 @@ func newSchemaValidationSettings(opts ...SchemaValidationOption) *schemaValidati
 	}
 	return settings
 }
+
+// SetOpenAPIMinorVersion setting allows to define minor OpenAPI version schema must comply with
+func SetOpenAPIMinorVersion(minorVersion uint64) SchemaValidationOption {
```
fenollp, line 72: This seems easier to read when used:
```suggestion
func SetOpenAPI(v float32) SchemaValidationOption {
```
And also panic if `NaN` or if not finite or actually just if not `3.0` or `3.1`.

openapi3/validation_options.go
```diff
@@ -92,6 +93,13 @@ func DisableExamplesValidation() ValidationOption {
 	}
 }
 
+// DisableVersionValidation disables openapi version validation.
+func DisableVersionValidation() ValidationOption {
```
fenollp, line 97: Please add the `Enable..` version as well.

openapi3/version.go
```diff
@@ -0,0 +1,42 @@
...
+func (oai Version) Validate(ctx context.Context) error {
+	if vo := getValidationOptions(ctx); !vo.versionValidationDisabled {
+		if !versionRegex.MatchString(string(oai)) {
+			return ErrInvalidVersion
+		}
+	}
+	return nil
+}
+
+// MinorVersion returns minor version from string assuming 0 is the default
```
fenollp, line 32: Mention that returned value is meaningful IFF `Validate` returned nil error.

openapi3/version.go
```diff
@@ -0,0 +1,42 @@
...
+func (oai Version) Minor() uint64 {
```
fenollp, line 33: `uint64` seems too big. Also why not return the two possible errors? (or edit func doc)

fenollp: I see you introduced this for use by the schema validation settings. Let's return a float then :)

openapi3/schema_formats.go
fenollp: (yes, I'm asking to extend the existing 2 methods with a variadic float)

.github/docs/openapi3.txt
```diff
@@ -3,18 +3,21 @@ const TypeArray = "array" ...
 const FormatOfStringForUUIDOfRFC4122 = `^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[1-5][0-9a-fA-F]{3}-[89abAB][0-9a-fA-F]{3}-[0-9a-fA-F]{12}$` ...
 const SerializationSimple = "simple" ...
 var SchemaErrorDetailsDisabled = false ...
+var ErrInvalidVersion ...
```
fenollp, line 6: What's your rationale to export this?

go.mod
```diff
@@ -3,12 +3,11 @@ module github.com/getkin/kin-openapi
...
 require (
-	github.com/go-openapi/jsonpointer v0.19.5
+	github.com/go-openapi/jsonpointer v0.19.6
 	github.com/gorilla/mux v1.8.0
-	github.com/invopop/yaml v0.1.0
+	github.com/invopop/yaml v0.2.0
 	github.com/mohae/deepcopy v0.0.0-20170929034955-c48cc78d4826
 	github.com/perimeterx/marshmallow v1.1.4
 	github.com/stretchr/testify v1.8.1
-	gopkg.in/yaml.v2 v2.4.0 // indirect
```
fenollp, line 12: What in this PR requires these upgrades?

If they are not needed by this PR I'd like these to be moved to another PR so we can move on them in parallel.

openapi3/schema_test.go
```diff
@@ -17,12 +18,11 @@ type schemaExample struct {
 	Title         string
 	Schema        *Schema
 	Serialization interface{}
-	AllValid      []interface{}
-	AllInvalid    []interface{}
+	AllValid      [][]interface{} // indexed by minor OpenAPI version supported
+	AllInvalid    [][]interface{} // also indexed by minor OpenAPI version supported
```
fenollp, line 22: I appreciate your work and these are quite a lot of changes but please try to keep changes to this file to a minimum and extract if possible the tests that rely on minor version to another part of this file or to another file dedicated to said changes. This file is quite dense and review is difficult already. If that's just too much work let's concentrate on the other comments. Thanks!

fenollp (changes requested):
Hey! Thanks for this. This is going to the right direction. I mostly have comments on the API design.

fenollp:
@cboitel Hey there. Do you need help moving this forward?

cboitel:
Nope. Just missed it while I was on vacations. Will get back to it next week.

cboitel:
@fenollp

I had a hard time to rebase. 

Finally found it was related to some new/existing tests defining string formats for uuid but not restoring string formats to original value.

I created a new Save/Restore/RestoreDefault functions for string formats so one can easily do that like we need in our tests.

I will now address your earlier comments in the upcoming days.

openapi3/schema_formats.go
```diff
@@ -21,26 +21,111 @@ type FormatCallback func(value string) error
...
+
+type versionedFormat struct {
 	regexp   *regexp.Regexp
 	callback FormatCallback
 }
 
+func (format *Format) add(minMinorVersion uint64, vFormat *versionedFormat, override bool) {
+	if format != nil {
+		if format.versionedFormats == nil {
+			format.versionedFormats = make([]*versionedFormat, minMinorVersion+1)
```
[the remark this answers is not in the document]
cboitel: Did that so it is very fast to determine format to use for a specific minor version when validating

openapi3/schema_formats.go
```diff
@@ -21,26 +21,111 @@ type FormatCallback func(value string) error
...
+	if format.versionedFormats != nil {
+		if minorVersion >= uint64(len(format.versionedFormats)) {
+			return format.versionedFormats[len(format.versionedFormats)-1]
+		}
+		return format.versionedFormats[minorVersion]
+	}
+	return nil
+}
+
+func (format Format) DefinedForMinorVersion(minorVersion uint64) bool {
```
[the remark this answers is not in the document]
cboitel: Will have to determine if we keep it so one can determine if a definition already exists for a specific minor version. Since format does not expose its content, this is the only way for one to know about it.

If we keep, I will need to add tests

openapi3/schema_formats.go
cboitel: uint64 comes from the string parsing function used when parsing OpenAPI document 
=> Go string parsing returns a uint64
=> Casting to uint8 would require minor version

Since most processors are 64 bits, using an uint8/uint64 does save memory nor it brings performance gain.

openapi3/schema_validation_settings.go
cboitel: see discussion earlier

openapi3/schema_validation_settings.go
cboitel: We are in the openapi3 package, so I assume this is for any OpenAPI 3.x definition. So having 3.0 is useless and we would have to ensure it is not <3.0 nor >= 4.0.

Using an uint also has the advantages of using 0 as default (not 3.0).

.github/docs/openapi3.txt
cboitel: So one can check at loading time he got an invalid error version, not some other kind of error.

go.mod
cboitel: removed if it is project way of handling decencies upgrade.

openapi3/schema_formats.go
cboitel: Gone with the variadic options. have a look

openapi3/schema_test.go
cboitel: I haven't gone through this one yet. Will keep you posted.

openapi3/version.go
cboitel: done

openapi3/validation_options.go
cboitel: done

fenollp:
Feel free to revisit this now that 3.1 and versioning API is here

PR #745, opened 2023-01-18
openapi3: fix error phrase in security scheme

Jefftree:
Encountered this while trying to validate a schema with "oauth2" type for security scheme.


openapi3/security_scheme.go
```diff
@@ -194,7 +194,7 @@ func (ss *SecurityScheme) Validate(ctx context.Context, opts ...ValidationOption
 	} else if len(ss.In) > 0 {
 		return fmt.Errorf("security scheme of type %q can't have 'in'", ss.Type)
 	} else if len(ss.Name) > 0 {
-		return errors.New("security scheme of type 'apiKey' can't have 'name'")
+		return errors.New("security scheme of type %q can't have 'name'", ss.Type)
```
fenollp, line 197:
```suggestion
		return fmt.Errorf("security scheme of type %q can't have 'name'", ss.Type)
```

PR #748, opened 2023-01-21
Refine schema error reason message

ori-shalom:
This PR adds tests to the behavior that was added in #735.
The purpose of the tests is to verify that the validated value doesn't appear in the `SchemaError.Reason` field.

Additionally, this PR refines the messages in `SchemaError.Reason` and makes sure `Reason` is always set.

### DIFF
--- a/openapi3/issue735_test.go
+++ b/openapi3/issue735_test.go
@@ -1,0 +1,279 @@
+package openapi3
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+type testCase struct {
+	name             string
+	schema           *Schema
+	value            interface{}
+	extraNotContains []interface{}
+	options          []SchemaValidationOption
+}
+
+func TestIssue735(t *testing.T) {
+	DefineStringFormat("uuid", FormatOfStringForUUIDOfRFC4122)
+	DefineStringFormat("email", FormatOfStringForEmail)
+	DefineIPv4Format()
+	DefineIPv6Format()
+
+	testCases := []testCase{
+		{
+			name:   "type string",
+			schema: NewStringSchema(),
+			value:  42,
+		},
+		{
+			name:   "type boolean",
+			schema: NewBoolSchema(),
+			value:  42,
+		},
+		{
+			name:   "type integer",
+			schema: NewIntegerSchema(),
+			value:  "foo",
+		},
+		{
+			name:   "type number",
+			schema: NewFloat64Schema(),
+			value:  "foo",
+		},
+		{
+			name:   "type array",
+			schema: NewArraySchema(),
+			value:  42,
+		},
+		{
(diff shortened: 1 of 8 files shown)


openapi3/issue735_test.go
```diff
@@ -0,0 +1,279 @@
...
+			value:            map[string]interface{}{"foo": 42},
+			extraNotContains: []interface{}{42},
+		},
+		{
+			name:             "additional properties other schema type",
+			schema:           NewSchema().WithAdditionalProperties(NewStringSchema()),
+			value:            map[string]interface{}{"foo": 42},
+			extraNotContains: []interface{}{42},
+		},
+		// TODO: uncomment when https://github.com/getkin/kin-openapi/pull/747 is merged
```
fenollp, line 142: It's merged

openapi3/issue735_test.go
```diff
@@ -0,0 +1,279 @@
...
+			}),
+			value:            map[string]interface{}{"foo": 42},
+			extraNotContains: []interface{}{42},
+		},
+		// TODO: uncomment when https://github.com/getkin/kin-openapi/issues/502 is fixed
+		//{
+		//	name: "read only properties",
+		//	schema: NewSchema().WithProperties(map[string]*Schema{
+		//		"foo": {ReadOnly: true},
+		//	}).WithNoAdditionalProperties(),
```
fenollp, line 164: We should rename this method to Without... in another PR

openapi3/schema.go
```diff
@@ -784,6 +784,11 @@ func (schema *Schema) WithAnyAdditionalProperties() *Schema {
 	return schema
 }
 
+func (schema *Schema) WithNoAdditionalProperties() *Schema {
```
fenollp, line 787: Well actually in this PR

openapi3/schema.go
```diff
@@ -1915,15 +1930,20 @@ func (schema *Schema) visitJSONObject(settings *schemaValidationSettings, value
...
 }
 
-func (schema *Schema) expectedType(settings *schemaValidationSettings, typ string) error {
+func (schema *Schema) expectedType(settings *schemaValidationSettings, value interface{}) error {
 	if settings.failfast {
 		return errSchema
 	}
+
+	a := "a"
+	if strings.Trim(schema.Type[:1], "aeiou") == "" {
```
fenollp, line 1939: Please just compare to the few possible values explicitly with a switch statement.

fenollp (changes requested):
LGTM. A few things to fix and we should look into that readOnly fields issue

ori-shalom:
Thanks for the review.
I fixed your comments + rebased and uncommented the `additionalProperties` test.
As for the `readOnly` & `writeOnly` issue, I think we can address in another PR.

fenollp (approved):

PR #750, opened 2023-01-24
openapi3: add support for UUID v6-8

alexg-axis:
The current UUID format regex matches UUID version 1-5 in accordance to
the current RFC 4122.

There's a draft to update the spec, which adds three new versions. The
intention of this PR is to add support for these versions.

See: https://datatracker.ietf.org/doc/html/draft-ietf-uuidrev-rfc4122bis-01.


alexg-axis:
I didn't find any reference stating that OpenAPI strictly supports UUID v1-5, hence I though it would be in scope to add support here directly.

fenollp:
Well this isn't the same RFC so I'm more encline to not update this var but instead to add a new one with the bis01 name. Is that fine with you?

zekth:
> > Well this isn't the same RFC so I'm more encline to not update this var but instead to add a new one with the bis01 name. Is that fine with you?
> 
> Do you still think that the `uuid` format should use the new bis01 name? If so, what would be the use of keeping the current var as well?

I think we might want to keep both esp is the consumer app have to support v1-5 and not v8. Eg: v8 has introduced `uuid.Max` also which is not supported in other implementations.

fenollp:
Please, don't modify current behavior. Add a second var with `bis01` suffix and test behavior (covered, not covered, covered by both).

fenollp:
@alexg-axis  Hey there. Do you need help moving this forward?

fenollp:
Done in https://github.com/getkin/kin-openapi/pull/1251

PR #752, opened 2023-01-25
openapi3: fix validation of non-empty interface slice value against array schema

### LINKED ISSUE #751: bug: validation of a slice of non-empty interface against an array schema should pass
### Context

An array schema validates only against a slice of empty interface: `[]interface{}{...}`. If a slice of non-empty interface (e.g. `[]string{...}`) is attempted to validate against that schema, this error occurs: `unhandled value of type []string`

### Reproduce

```go
package main

import (
	"fmt"

)

func main() {
	schema := &openapi3.Schema{
		Type:        "array",
		UniqueItems: true,
		Items:       openapi3.NewStringSchema().NewRef(),
	}
	data := []string{"foo", "bar"}
	err := schema.VisitJSON(data)
	fmt.Println(err)
}
```

### Expected

In this case, a nil error should be printed because `VisitJSON` would recognize that `data` is a slice.

There is a bug in `VisitJSON` where the `[]interface{}` case does not cover all slices: https://github.com/getkin/kin-openapi/blob/9f99fee5a643dd7522772022cda4eee8b7f3f614/openapi3/schema.go#L1108-L1109

andrewyang96:
Resolves #751 

In `VisitJSON` after the `switch` block, it checks `value`'s Kind. If it is a Slice, it transforms the `value` into a slice of empty interfaces and passes it to `visitJSONArray`

`TestSchemas` did not catch this because the data (`AllValid` and `AllInvalid`) are coerced to empty interface types.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1120,6 +1120,17 @@ func (schema *Schema) visitJSON(settings *schemaValidationSettings, value interf
 			return schema.visitJSONObject(settings, values)
 		}
 	}
+
+	// Catch slice of non-empty interface type
+	if reflect.TypeOf(value).Kind() == reflect.Slice {
+		valueR := reflect.ValueOf(value)
+		newValue := make([]interface{}, valueR.Len())
+		for i := 0; i < valueR.Len(); i++ {
+			newValue[i] = valueR.Index(i).Interface()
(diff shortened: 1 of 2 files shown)


fenollp:
Tanks!

PR #754, opened 2023-01-27
openapi3: empty scopes are valid

nodar963:
accoriding to v3.1.0 documentation the scopes field can be empty.
https://spec.openapis.org/oas/v3.1.0#fixed-fields-24


fenollp:
Thanks

PR #755, opened 2023-01-28
openapi3: fix integer enum schema validation after json.Number PR

k2tzumi:
The integer enum could not be validated correctly.
see.
https://github.com/k2tzumi/kin-openapi/actions/runs/4030301397/jobs/6928881789#step:18:19

Added appropriate conversion process since it was comparing `json.Number` with `float64`


fenollp:
Thanks!

PR #763, opened 2023-02-06
openapi3: add support for extensions on the few types left

fenollp:


openapi3/operation.go
```diff
@@ -162,24 +162,18 @@ func (operation Operation) JSONLookup(token string) (interface{}, error) {
...
-	if status != 0 {
+	if 0 < status && status < 1000 {
 		code = strconv.FormatInt(int64(status), 10)
 	}
-	responses[code] = &ResponseRef{
-		Value: response,
+	if operation.Responses == nil {
+		operation.Responses = &Responses{}
 	}
+	operation.Responses.Set(code, &ResponseRef{Value: response})
```
fenollp, line 176: `Responses.Set(..)` should fail loudly if responses is nil

openapi3/operation.go
```diff
@@ -162,24 +162,18 @@ func (operation Operation) JSONLookup(token string) (interface{}, error) {
...
 
 func (operation *Operation) AddResponse(status int, response *Response) {
-	responses := operation.Responses
-	if responses == nil {
-		responses = NewResponses()
-		operation.Responses = responses
-	}
 	code := "default"
-	if status != 0 {
+	if 0 < status && status < 1000 {
```
fenollp, line 170: move to another pr (as well as the other status changes)

openapi3/paths.go
```diff
@@ -2,35 +2,179 @@ package openapi3
...
+	if paths == nil || paths.m == nil {
+		return nil
+	}
+	return paths.m[key]
+}
+
+// Set adds or replaces the path value for key
+func (paths *Paths) Set(key string, value *PathItem) {
+	if paths == nil {
+		paths = &Paths{}
```
fenollp, line 43: test that this is effectual

openapi3/paths.go
```diff
@@ -2,35 +2,179 @@ package openapi3
...
+
+// Len returns the amount of paths
+func (paths *Paths) Len() int {
+	if paths == nil || paths.m == nil {
+		return 0
+	}
+	return len(paths.m)
+}
+
+// Map returns paths as an unordered map
```
fenollp, line 59: just mention that iteration isn't ordered

openapi3/response.go
```diff
@@ -7,63 +7,200 @@ import (
...
+func (responses *Responses) Default() *ResponseRef {
+	return responses.Value("default")
+}
+
+// Status returns the response that exactly matches the given status, or nil.
+func (responses *Responses) Status(status int) *ResponseRef {
+	return responses.Value(strconv.FormatInt(int64(status), 10))
+}
+
+// FIXME Any HTTP status code can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. A Reference Object can link to a response that is defined in the OpenAPI Object's components/responses section. This field MUST be enclosed in quotation marks (for example, "200") for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character X. For example, 2XX represents all response codes between [200-299]. Only the following range definitions are allowed: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.
```
fenollp, line 72: this -> PR++

openapi3/refs_test.go
```diff
@@ -230,48 +233,55 @@ components:
...
+	require.Equal(t, reflect.TypeOf(&Paths{}).Kind(), kind)
 
 	ptr, err = jsonpointer.New("/paths/~1pet")
 	require.NoError(t, err)
 	v, kind, err = ptr.Get(doc)
 	require.NoError(t, err)
-	require.IsType(t, &PathItem{}, v)
-	require.Equal(t, reflect.TypeOf(&PathItem{}).Kind(), kind)
+	require.NotNil(t, v)
+	require.IsType(t, PathItem{}, v)
```
fenollp, line 245: these (and below) should stay ptr types

openapi2conv/openapi2_conv.go
```diff
@@ -605,13 +603,16 @@ func FromV3(doc3 *openapi3.T) (*openapi2.T, error) {
...
+
 	if isHTTPS {
 		doc2.Schemes = append(doc2.Schemes, "https")
 	}
 	if isHTTP {
 		doc2.Schemes = append(doc2.Schemes, "http")
 	}
-	for path, pathItem := range doc3.Paths {
+
+	// TODO: add paths extensions to doc2
```
fenollp, line 614: todo

openapi2conv/openapi2_conv.go
```diff
@@ -1041,9 +1042,10 @@ func FromV3Parameter(ref *openapi3.ParameterRef, components *openapi3.Components
 	return result, nil
 }
 
-func FromV3Responses(responses map[string]*openapi3.ResponseRef, components *openapi3.Components) (map[string]*openapi2.Response, error) {
-	v2Responses := make(map[string]*openapi2.Response, len(responses))
-	for k, response := range responses {
+func FromV3Responses(responses *openapi3.Responses, components *openapi3.Components) (map[string]*openapi2.Response, error) {
+	// TODO: add responses extensions to doc2
```
fenollp, line 1046: todo

openapi3/loader.go
```diff
@@ -390,12 +390,21 @@ func readableType(x interface{}) string {
...
-	if s, ok := cursor.(*SchemaRef); ok && fieldName == "additionalProperties" {
-		if ap := s.Value.AdditionalProperties.Has; ap != nil {
-			return *ap, nil
+	switch c := cursor.(type) {
+	case *Responses:
+		cursor = c.m // m map[string]*ResponseRef
+	case *Callback:
+		cursor = c.m // m map[string]*PathItem
+	case *Paths:
+		cursor = c.m // m map[string]*PathItem
```
fenollp, line 399: Mutate `cursor` IFF `fieldName` in `c.m`?

openapi3/response.go
```diff
@@ -4,90 +4,102 @@ import (
...
 // See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#responses-object
-type Responses map[string]*ResponseRef
+type Responses struct {
+	Extensions map[string]interface{} `json:"-" yaml:"-"`
+
+	m map[string]*ResponseRef
+}
+
+// NewEmptyResponses creates an empty yet valid responses object
+func NewEmptyResponses() *Responses {
```
fenollp, line 20: use when capacity=0 in NewResponses and drop this pub impl

mcblair:
This shipped in `v0.122.0` and when we upgraded to it from `v0.121.0` resulted in a silent change in behaviors when marshalling to yaml; `paths: {}`. I have confirmed that prior to marshalling, the field has a pointer to expected value. I also have confirmed that `T.MarshalJSON()` behaves as expected, without any loss of `T.Paths`. We were still using `yaml.v2` but I did confirm that the problem is reproducible with `yaml.v3`.

Reproduce it by running the following code first with `v0.121.0` and then with `v0.122.0`. 
```golang
oas := openapi3.NewLoader()
existingDoc, err := oas.LoadFromFile("./spec/reference/example.yaml")
if err != nil {
	return fmt.Errorf("failed to load file: %s with err: %s", s.FileName, err.Error())
}
// `existingDoc.Paths` looks good when introspected with debugger.

err = existingDoc.Validate(ctx)
if err != nil {
	return fmt.Errorf("failed to validate file: %s with err: %s", s.FileName, err.Error())
}

marshalledJson, err := existingDoc.MarshalJSON() // Looks as expected.
if err != nil {
	return err
}

marshalledYaml, err := yaml.Marshal(&existingDoc) // Loss of `T.Paths`
if err != nil {
	return err
}
```

sonu27:
This was a big breaking change.
I can no longer loop on path items, which were made private.
Could `m map[string]*PathItem` be made public?

fenollp:
@sonu27 https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3#Paths.Map

sonu27:
> @sonu27 https://pkg.go.dev/github.com/getkin/kin-openapi/openapi3#Paths.Map

I'm using that. The problem is marshalling of yaml. Which I believe there is another bug for

PR #768, opened 2023-02-13
openapi3: fix an infinite loop that may have been introduced in #700

fenollp:


fenollp:
@TristanSpeakEasy This might be of interest to you :)

TristanSpeakEasy:
@fenollp ah perfect yep that looks like it could help 👍

ShouheiNishi:
Why did we make this fix?
In what cases does the #700 fix cause problems?

fenollp:
Hi @ShouheiNishi ! Feel like you wanna [open an issue?](https://github.com/getkin/kin-openapi/issues/new/choose)

ShouheiNishi:
I would like an explanation of the problem that was fixed by this fix.

PR #770, opened 2023-02-13
openapi3: fix default values count even when disabled (#767)

### LINKED ISSUE #767: Bug: Default values still count as properties even if setting them is disabled.
https://github.com/getkin/kin-openapi/blob/ecb06bc5c05b010b9c1322486ff74f2178f7927c/openapi3/schema.go#L1774-L1816

In `visitJSONObject`, if `value[propName]` is `nil`, it is assigned with the default value regardless of the setting (line 1794).

This counts towards `lenValue` (line 1813) and can falsely pass validations like `minProperties` that way.

I will add a PR to fix this bug.

orshlom:
fixes #767

### DIFF
--- a/openapi3/issue767_test.go
+++ b/openapi3/issue767_test.go
@@ -1,0 +1,90 @@
+package openapi3_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+func TestIssue767(t *testing.T) {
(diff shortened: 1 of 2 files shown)


fenollp:
Thanks for this! Could you add a regression test please?

gczobel:
@fenollp any idea when this will be released? I'm looking forward to this change...

fenollp:
@gczobel You got it => https://github.com/getkin/kin-openapi/releases/tag/v0.115.0

PR #774, opened 2023-03-06
feature: implement loading & interalizing examples references

DerkSchooltink:
When using the OpenAPI loader and using the interalizing references functionality, I noticed that examples are not picked up and added to components if they are set up using the following structure:

```
"content": {
  "application/json": {
    "examples": {
      "post-test-ref-response": {
        "summary": "Example POST test ref response",
        "value": {
          "$ref": "examples/response-post-test-ref.json"
        }
      }
    },
    "schema": {
      "$ref": "response-post-test-ref.json"
    }
  }
}
```

The loader expects the reference to be there at the top-level, right under `"examples"`, just like a singular `"example"`:

```
"example": {
  "$ref": "response-post-test-ref.json"
}
```

So in order to load nested value references, I added an extra functionality to the loader that checks if there's a `$ref` available inside the value. If there is one, then create a `Example` instance from scratch, using that same $ref value as value for the ref of the example, and also load the raw JSON/Yaml example and assign that to the value of the example. 

Every other test seems to succeed so I don't think there's regression from my change, but I would like to request someone to take a look at it to make sure the implementation matches expectations.


openapi3/internalize_refs.go
```diff
@@ -376,12 +379,12 @@ func (doc *T) derefPaths(paths map[string]*PathItem, refNameResolver RefNameReso
 // refNameResolver takes in references to returns a name to store the reference under locally.
 // It MUST return a unique name for each reference type.
 // A default implementation is provided that will suffice for most use cases. See the function
-// documention for more details.
+// documentation for more details.
 //
 // Example:
 //
-//	doc.InternalizeRefs(context.Background(), nil)
-func (doc *T) InternalizeRefs(ctx context.Context, refNameResolver func(ref string) string) {
```
DerkSchooltink, line 384: `Context` was not used anywhere in this function, so I removed it to avoid confusion when people try to use this function

fenollp: Please don't. That's an unrelated change involving API breakage. Let's leave this to a separate discussion.

fenollp:
Let's see what the open api spec says. I'm on my phone and have much time to look into this right now. Maybe you can paste the part of the documentation about such `examples` values? To make sure `$ref`s can be used there.

DerkSchooltink:
> Let's see what the open api spec says. I'm on my phone and have much time to look into this right now. Maybe you can paste the part of the documentation about such `examples` values? To make sure `$ref`s can be used there.

DD484799-E6D9-48F4-BF0C-6E2EB02558E3
AF629A1B-70DF-4BA0-876B-F55B76371E9D
D0A25056-FD6A-4744-AC47-CFE5A9FDF26B

This is what I could find. You can create "named" examples, which have a summary and value. The value can be an object holding a $ref. 

I know it works because I'm using it to define OpenAPI schemas with no inline schemas or examples whatsoever. And they validate just fine 😁 

When I have time tomorrow I can delve into the OpenAPI schema definition to find out the details but I remember it being quite liberal as to what is allowed to be into the examples block.

DerkSchooltink:
> Let's see what the open api spec says. I'm on my phone and have much time to look into this right now. Maybe you can paste the part of the documentation about such `examples` values? To make sure `$ref`s can be used there.

image

There's not a lot of documentation on it, this is what I could find. See https://swagger.io/docs/specification/adding-examples/ all the way at the bottom.

openapi3/internalize_refs.go
DerkSchooltink: Fair enough, will revert

fenollp:
From https://github.com/OAI/OpenAPI-Specification/blob/4bfca78fb67825214f65af404944ef1aff62190c/versions/3.0.3.md#media-type-object

(excerpt truncated for emphasis)

--- 

#### <a name="mediaTypeObject"></a>Media Type Object
Each Media Type Object provides schema and examples for the media type identified by its key.

##### Fixed Fields
Field Name | Type | Description
---|:---:|---
<a name="mediaTypeExamples"></a>examples | Map[ `string`, [Example Object](#exampleObject) \| [Reference Object](#referenceObject)] | Examples of the media type.  Each example object SHOULD  match the media type and specified schema if present.  The `examples` field is mutually exclusive of the `example` field.  Furthermore, if referencing a `schema` which contains an example, the `examples` value SHALL _override_ the example provided by the schema.

--- 

#### <a name="exampleObject"></a>Example Object

##### Fixed Fields
Field Name | Type | Description
---|:---:|---
<a name="exampleValue"></a>value | Any | Embedded literal example. The `value` field and `externalValue` field are mutually exclusive. To represent examples of media types that cannot naturally represented in JSON or YAML, use a string value to contain the example, escaping where necessary.
<a name="exampleExternalValue"></a>externalValue | `string` | A URL that points to the literal example. This provides the capability to reference examples that cannot easily be included in JSON or YAML documents.  The `value` field and `externalValue` field are mutually exclusive. 

--- 

To me, this reads as "examples can be refs" BUT the `value` field of an `examples` map, which by the way has the type `Any`, IS NOT made to include dereferenceable `$ref`s. So this lib's code as is is correct and your addition goes further than the OpenAPI Spec says.

I think you simply are looking for the `externalValue` field of the `examples` map. Don't you agree?

This is great work anyway. If you're so inclined I'll gladly accept PRs with other fixes, typos, documentation, ... Thanks!

Issue #775, opened 2023-03-09
Data Race when handling simultaneously requests


wrong-user:
I am facing an Data Race when I am sending simultaneously request against my service that is using https://github.com/getkin/kin-openapi/blob/e53fe386960a0f532637da538535f22c9a9249e6/openapi3filter/validate_request.go#L32 to validate the incoming requests.

The OpenApi specification is loaded by the service. Validation of the request against the spec is handle in a middleware before it is handle by the service itself, or reject if it does not adhere to the spec. 

The race is occurring because of a write at https://github.com/getkin/kin-openapi/blob/e53fe386960a0f532637da538535f22c9a9249e6/openapi3/schema.go#L1973 and a read at https://github.com/getkin/kin-openapi/blob/e53fe386960a0f532637da538535f22c9a9249e6/openapi3/schema.go#L1608

Out of curiosity I added a mutex in the schema struct and applied it in as below.

```go
// Schema.go:1599
schema.PatternMu.Lock()
if schema.Pattern != "" && schema.compiledPattern == nil && !settings.patternValidationDisabled {
  var err error
  if err = schema.compilePattern(); err != nil {
    if !settings.multiError {
      return err
    }
    me = append(me, err)
  }
}
schema.PatternMu.Unlock()
```

I am running it with v.0.114.0 at commit ecb06bc5, but the issue is also present in older versions.

fenollp:
Hello please show us: open a PR with a test that reproduces your issue. Thanks.

wrong-user:
Sure, I will do that.

fenollp:
Hey so I have a fix, two actually!

Short term: `doc.Validate(ctx)` before building your `router`. You should do that anyway. It pre compiles all patterns so in effect fixes this issue.

Longer term: I'll propose a patch that relies on a package-global `sync.Map`. 

But do validate your docs!

PR #776, opened 2023-03-17
Add test to reproduce issue 775

wrong-user:
Test to reproduce the issue with data race when service is hit with simultaneous request.

needs to be run with `-race` flag

Reproduces #775


fenollp:
Please do what's needed to have CI show your issue (the data race). Edit the ci file if needed. 
Right now ci fails for reasons not mentioned in your issue. Thanks.

wrong-user:
I tried running the github action locally with `github.com/nektos/act` , then I  got the test to fail as it did for  `1.x on macos-latest`. But then i used  `1.6 on ubuntu-latest`.    

I can troubleshoot it some more or is this good enough?

PR #778, opened 2023-03-23
feat: support nil uuid

zekth:
This makes the regex supporting `nil` UUID like defined in the RFC: https://www.rfc-editor.org/rfc/rfc4122#section-4.1.7


fenollp:
Please add a test. 
I'd also prefer you add the zeros to the third and fifth square brackets instead of a big group.

zekth:
> I'd also prefer you add the zeros to the third and fifth square brackets instead of a big group.

Wouldn't this allow to create invalid uuids then?

fenollp:
Ah you're right

PR #782, opened 2023-04-09
Expose the ConvertErrors func (#7)

QifanWuCFLT:
Need to expose the ConvertErrors func for our service.


codyaray:
why was this closed? :)

fenollp:
@codyaray No idea but if you can cherry pick that commit on top latest master and make a PR so both your attributions stick I'll review it. If you can't then send a PR with your commit :)

QifanWuCFLT:
I can't remember the context but this change is already there. https://github.com/getkin/kin-openapi/pull/783

PR #783, opened 2023-04-09
openapi3filter: introduce `func ConvertErrors(err error) error`

QifanWuCFLT:
- Need to expose the ConvertErrors func for our service.
- fix failed CI


QifanWuCFLT:
@fenollp Could I get a review on this pr? thx!

Issue #789, opened 2023-05-04
Validating OneOf in query params always matches each schema


ju-zp:
When use oneOf in the query schema validation will always fail due to the given value matching all of the schemas, even if it doesn't match them at all.

As an example, this schema:
```
paths:
  /items:
    get:
      description: Returns a list of stuff
      parameters:
      - description: test object
        explode: false
        in: query
        name: test
        required: true
        schema:
          type: string
          oneOf:
            - pattern: '\babc\b'
            - pattern: '\bfoo\b'
            - pattern: '\bbar\b'
``` 

Example request:
```
/items?test=abc
/items?test=def
/items?test=foo bar
``` 
Given the example requests above, all of them fail with the same error message "parameter \"test\" in query has an error: decoding oneOf failed: 3 schemas matched". It seems that they will always match each pattern regardless.

PR #790, opened 2023-05-04
openapi3filter: fix oneOF validation in query params

ju-zp:
Addresses issue raised here https://github.com/getkin/kin-openapi/issues/789. The fix differs validation to within the schema as the current implemention would always validate incorrectly. I've added test cases for the original issue, along with removing a test within `validation_test.go` which would always fail, so wasn't actually testing the intended behaviour.

### DIFF
--- a/openapi3filter/issue789_test.go
+++ b/openapi3filter/issue789_test.go
@@ -1,0 +1,128 @@
+package openapi3filter_test
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers/gorillamux"
+)
+
+func TestIssue789(t *testing.T) {
+
+	anyOfArraySpec := `
+openapi: 3.0.0
+info:
+  version: 1.0.0
+  title: Sample API
+paths:
+  /items:
+    get:
+      description: Returns a list of stuff
+      parameters:
+      - description: test object
+        explode: false
+        in: query
+        name: test
+        required: true
+        schema:
+          type: string
+          anyOf:
+            - pattern: '\babc\b'
+            - pattern: '\bfoo\b'
+            - pattern: '\bbar\b'
+      responses:
+        '200':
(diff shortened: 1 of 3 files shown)


openapi3filter/issue789_test.go
```diff
@@ -0,0 +1,128 @@
...
+        explode: false
+        in: query
+        name: test
+        required: true
+        schema:
+          type: string
+          anyOf:
+            - pattern: '\babc\b'
+            - pattern: '\bfoo\b'
+            - pattern: '\bbar\b'
```
fenollp, line 37: I understand `\b` is the backspace character. What do you think its use is here?

openapi3filter/issue789_test.go
```diff
@@ -0,0 +1,128 @@
...
+      responses:
+        '200':
+          description: Successful response
+`[1:]
+
+	oneOfArraySpec := strings.ReplaceAll(anyOfArraySpec, "anyOf", "oneOf")
+
+	allOfArraySpec := strings.ReplaceAll(strings.ReplaceAll(anyOfArraySpec, "anyOf", "allOf"),
+		"type: boolean", "type: number")
```
fenollp, line 46: The second `ReplaceAll` looks ineffectual.

openapi3filter/issue789_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers/gorillamux"
+)
+
+func TestIssue789(t *testing.T) {
+
```
fenollp, line 16:
```suggestion
```

fenollp (changes requested):
LGTM Please amend per review.

openapi3filter/issue789_test.go
ju-zp: yes, you are right. Have updated to only replace anyOf to allOf

openapi3filter/issue789_test.go
ju-zp: This is regex, and `\b` is being used here to indicate word boundary

PR #792, opened 2023-05-10
openapi3gen: fix panic when embedded type is not a struct

nrwiersma:
This PR fixes a panic when an embedded type is not a struct. 

To infer the actual type from the embedded type is specifically complex, it is simpler in these cases to use a schema customizer. It makes the most sense to ignore the type.


fenollp (approved):

fenollp:
Thanks!

PR #797, opened 2023-06-05
openapi3: fix #796 infinite loop during document validation

### LINKED ISSUE #796: Document validation infinite loop
Hi, 

We found out that in some cases, the Validate() method enters an infinite loop when validating a valid API document.

For example:
[https://github.com/APIs-guru/openapi-directory/blob/51fed63e3e6bce022cedba44ce28863bd712040c/APIs/presalytics.io/ooxml/0.1.0/openapi.yaml](url)

This happens because the "stack" slice isn't handled correctly in the validate receiver of "*Schema":
`func (schema *Schema) validate(ctx context.Context, stack []*Schema) error`

While the slice is being resliced during its lifecycle in the recursion, we do not store the resliced result and use the original slice instead. 

NOTE: In order to load this spec, we had to set the "CircularReferenceCounter" to 20.

radwaretaltr:
Fix for issue: #796


openapi3/issue796_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue796(t *testing.T) {
+	// Need to set CircularReferenceCounter to > 10
+	CircularReferenceCounter = 20
```
fenollp, line 11: To pass CI you'll need to edit the line in `go.yml` to be 
```shell
[[ $(git grep -InE 'return .*validateExtensions' -- openapi3 | wc -l) -eq $(git grep -InE '^\s+Extensions.+`' -- openapi3 | wc -l) ]]
```
Note the first regexp and flags is different from the one in https://github.com/getkin/kin-openapi/blob/fd52f34d334769c4cbcabba87272a2c2a5b7aa4f/.github/workflows/go.yml#L129

Here I feel you'll need to do a `defer` trick like https://github.com/getkin/kin-openapi/blob/fd52f34d334769c4cbcabba87272a2c2a5b7aa4f/openapi3/issue615_test.go#L23-L25

fenollp (changes requested):
LGTM just a nit to pass tests

openapi3/issue796_test.go
radwaretaltr: Thanks. Done.

radwaretaltr:
@fenollp Done. Can you please create a tag from it once it's merged?

Issue #819, opened 2023-07-06
Invalid handling of wildcard status codes


ryanslade:
It's possible to define "Patterned fields" in the [spec](https://swagger.io/specification/), for example `5XX`:

> Any [HTTP status code](https://swagger.io/specification/#http-codes) can be used as the property name, but only one property per code, to describe the expected response for that HTTP status code. This field MUST be enclosed in quotation marks (for example, "200") for compatibility between JSON and YAML. To define a range of response codes, this field MAY contain the uppercase wildcard character X. For example, 2XX represents all response codes between [200-299]. Only the following range definitions are allowed: 1XX, 2XX, 3XX, 4XX, and 5XX. If a response is defined using an explicit code, the explicit code definition takes precedence over the range definition for that code.

When validating a response, the current code does a simple [map lookup](https://github.com/getkin/kin-openapi/blob/4ba7452cbb41faf5a2fb5339919c6f3b3d685a23/openapi3filter/validate_response.go#L50) which will not match specific error codes with their wildcards. If no default clause is specified a `200` response will be returned instead.

PR #827, opened 2023-08-17
openapi3: support patterned fields in Responses (#819)

mgirouard:
Implements patterned fields in `openapi3.Responses.Get` 

This seemed to be the quickest fix for #819. Open to other ideas as well!

### DIFF
--- a/openapi3/issue_819_test.go
+++ b/openapi3/issue_819_test.go
@@ -1,0 +1,36 @@
+package openapi3
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue819ResponsesGetPatternedFields(t *testing.T) {
+	spec := `
+openapi: "3.0.3"
+info:
+  title: 'My app'
+  version: 1.0.0
+  description: 'An API'
+
+paths:
+  /v1/operation:
+    get:
+      summary: Fetch something
+      responses:
+        2XX:
+          description: Success
+          content:
+            application/json:
+              schema:
+                type: object
+                description: An error response body.
+`[1:]
+	sl := NewLoader()
+	doc, err := sl.LoadFromData([]byte(spec))
+	require.NoError(t, err)
+	err = doc.Validate(sl.Context)
+	require.NoError(t, err)
+	require.NotNil(t, doc.Paths["/v1/operation"].Get.Responses.Get(201))
+}
--- a/openapi3/response.go
+++ b/openapi3/response.go
@@ -30,17 +30,27 @@ func (responses Responses) Default() *ResponseRef {
 // Get returns a ResponseRef for the given status
 // If an exact match isn't initially found a patterned field is checked using
 // the first digit to determine the range (eg: 201 to 2XX)
+// See https://spec.openapis.org/oas/v3.0.3#patterned-fields-0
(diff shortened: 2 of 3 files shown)


openapi3/issue_819_test.go
```diff
@@ -0,0 +1,36 @@
...
+              schema:
+                type: object
+                description: An error response body.
+`[1:]
+	sl := NewLoader()
+	doc, err := sl.LoadFromData([]byte(spec))
+	require.NoError(t, err)
+	err = doc.Validate(sl.Context)
+	require.NoError(t, err)
+	require.NotNil(t, doc.Paths["/v1/operation"].Get.Responses.Get(201))
```
fenollp, line 35: * Please rename this file to `issue819_...` (drop first `_`) so it follows the others.
* Please add a test that `Get`s a `nil` response. e.g. `require.Nil(t, doc.Paths["/v1/operation"].Get.Responses.Get(404))`

openapi3/response.go
```diff
@@ -27,8 +27,21 @@ func (responses Responses) Default() *ResponseRef {
 	return responses["default"]
 }
 
+// Get returns a ResponseRef for the given status
+// If an exact match isn't initially found a patterned field is checked using
+// the first digit to determine the range (eg: 201 to 2XX)
 func (responses Responses) Get(status int) *ResponseRef {
-	return responses[strconv.FormatInt(int64(status), 10)]
+	st := strconv.FormatInt(int64(status), 10)
+	rref, ok := responses[strconv.FormatInt(int64(status), 10)]
```
fenollp, line 35:
```suggestion
	rref, ok := responses[st]
```

openapi3/response.go
```diff
@@ -27,8 +27,21 @@ func (responses Responses) Default() *ResponseRef {
...
+	if ok {
+		return rref
+	}
+	st = string(st[0]) + "XX"
```
fenollp, line 39: shouldn't this try `string(st[0]) + string(st[1]) + "X"` first?

Also, isn't this code missing `len(st)` checks?

mgirouard: It looks like there are only the specific ranges allowed in the spec.
> Only the following range definitions are allowed: 1XX, 2XX, 3XX, 4XX, and 5XX.

Perhaps a switch on `st` gives us the same guarantees all things considered?

fenollp: yes please let's stick to the spec and `case` each branch. You may panic on unexpected data. Just mention that in the func's docs.

mgirouard: Updated. Let me know if that works. I opted not to introduce a panic (even if the range is out of bounds) as the method didn't previously panic with invalid values -- perhaps this can change in a future release (although, I'd still probably prefer an error).

fenollp (approved):

ffkingkeshava-lgtm:
Screenshot_20250810_214447
1k

ffkingkeshava-lgtm:
Screenshot_20250810_214447

ffkingkeshava-lgtm:
ALLU ARJUN MASS HERO

ffkingkeshava-lgtm:
100k

ffkingkeshava-lgtm:
@ffkingkeshava-lgtm

PR #828, opened 2023-08-18
Merge allOf

reuvenharrison:
## Objective
This PR introduces a function that simplifies OpenAPI specifications by removing subschemas under `allOf` and replacing them with a single merged (or flattened) top-level schema.
The resulting schema remains equivalent to the original one in terms of validation.
The use case we had in mind for this PR is for detecting breaking-changes which is implemented in https://github.com/Tufin/oasdiff using kin-openapi3. There may be other use cases which we did not think about.

## The `Merge` function
* The `Merge` function accepts a schema as input and returns a new equivalent schema without `allOf` 
* If the specification can't be merged, `Merge` returns an error
* Source: openapi3/schema_merge.go

## Unit testing
* Test how `Merge` works for different kinds of subschemas under `allOf`
* Source: openapi3/schema_merge_test.go

## Validation testing
* In order to verify equivalence of the original schema and the merged schema, we apply `ValidateRequest` to various inputs and verify that either the input matches both schemas or the input doesn't match both schemas, if one matches and the other doesn't, then the schemas are not equivalent.
* Source: openapi3filter/validation_merge_allof_test.go

## Development Status
* This PR is still under active development, and not all schema fields are currently supported.
* At present, we support the following schema fields:
  * Properties
  * AdditionalProperties
  * MinProps
  * MaxProps
  * Type
  * Format
  * Required
  * Min
  * Max
  * ExclusiveMin
  * ExclusiveMax
  * Items
  * MinItems
  * MaxItems
  * UniqueItems
  * Enum
  * MinLength
  * MaxLength
  * MultipleOf
  * Pattern: not supported because combining regex expressions requires positive lookahead which isn't supported by kin-openapi's regex engine
  * Title
  * Description
  * Not
  * OneOf and AnyOf: supported, but there might be some tricky situations where the result doesn't quite match the original specification, we are working to resolve these cases.


openapi3/schema_merge.go
```diff
@@ -0,0 +1,789 @@
...
+	}
+	return findIntersectionOfArrays(nonEmptyEnum)
+}
+
+func resolvePattern(values []string) string {
+	var pattern strings.Builder
+	for _, p := range values {
+		if len(p) > 0 {
+			if !isPatternResolved(p) {
+				pattern.WriteString(fmt.Sprintf("(?=%s)", p))
```
reuvenharrison, line 439: @tcdsv can you explain how this combines the patterns?

tcdsv: we use positive lookahead (?=<pattern>) to specify that regex patterns must follow eachother, this enabled to combine multiple patterns.

The part you're talking about helps us avoid double-wrapping a pattern with "(?=..)"
This suggests a refactor for making that intent clearer.
Also, I've noticed that we don't have any validation tests for the pattern, I will add it.

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1033 @@
...
+				Value: &Schema{
+					Type:         "object",
+					ExclusiveMax: true,
+					Max:          Float64Ptr(1),
+				},
+			},
+			&SchemaRef{
+				Value: &Schema{
+					Type:         "object",
+					ExclusiveMin: false,
```
reuvenharrison, line 24: Note this is ExclusiveMin.
Perhaps you meant ExclusiveMax?

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1033 @@
...
+				Value: &Schema{
+					Type:         "object",
+					ExclusiveMax: false,
+					Max:          Float64Ptr(1),
+				},
+			},
+			&SchemaRef{
+				Value: &Schema{
+					Type:         "object",
+					ExclusiveMin: true,
```
reuvenharrison, line 48: Note this is ExclusiveMin.
Perhaps you meant ExclusiveMax?

openapi3/schema_merge_test.go
tcdsv: fixed, thanks.

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1273 @@
...
+								Required: []string{"boolean"},
+							},
+						},
+					},
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, []string{"string", "boolean"}, merged.OneOf[0].Value.Required)
```
reuvenharrison, line 310: I think it should be `boolean` only.

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1273 @@
...
+							},
+						},
+					},
+					Type: "object",
+				},
+			},
+		},
+	})
+
+	require.EqualError(t, err, TypeErrorMessage)
```
reuvenharrison, line 140: This error is used for two different scenarios.
I recommend to use a dedicated error here.

openapi3/schema_merge.go
```diff
@@ -0,0 +1,887 @@
...
+		return schema, nil
+	}
+
+	if areFormatsNumeric(formats) {
+		orderMap := make(map[string]int)
+		orderMap[formatInt32] = 1
+		orderMap[formatInt64] = 2
+		orderMap[formatDouble] = 3
+		orderMap[formatFloat] = 4
+		result := formatFloat
```
reuvenharrison, line 543: Shouldn't the order be reversed?
According to the OpenAPI Spec, Double is wider than Float:
| Type | Format | Description|
|-- | -- | -- |
|number | float | Floating-point numbers.|
|number | double | Floating-point numbers with double precision.|

openapi3/schema_merge.go
```diff
@@ -0,0 +1,887 @@
...
+	if len(formats) == 0 {
+		schema.Format = ""
+		return schema, nil
+	}
+
+	if areFormatsNumeric(formats) {
+		orderMap := make(map[string]int)
+		orderMap[formatInt32] = 1
+		orderMap[formatInt64] = 2
+		orderMap[formatDouble] = 3
```
reuvenharrison, line 541: Shouldn't the order be reversed?
According to the OpenAPI Spec, Double is wider than Float:
| Type | Format | Description|
|-- | -- | -- |
|number | float | Floating-point numbers.|
|number | double | Floating-point numbers with double precision.|

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1273 @@
...
+					Title: "second",
+				},
+			},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, "base schema", merged.Title)
+}
+
+func TestMerge_NumericFormat(t *testing.T) {
```
reuvenharrison, line 788: I recommend to split this into 3 tests

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1273 @@
...
+		},
+	})
+	require.EqualError(t, err, FormatErrorMessage)
+}
+
+func TestMerge_EmptySchema(t *testing.T) {
+	schema := Schema{}
+	merged, err := Merge(schema)
+	require.NoError(t, err)
+	require.Equal(t, &schema, merged) //todo &schema
```
reuvenharrison, line 940: I suggest to clarify the TODO message or remove it

openapi3/schema_merge_test.go
```diff
@@ -0,0 +1,1273 @@
...
+}
+
+func TestMerge_NoAllOf(t *testing.T) {
+	schema := Schema{
+		Title: "test",
+	}
+	merged, err := Merge(schema)
+	require.NoError(t, err)
+	require.Equal(t, &schema, merged) //todo &schema
```
reuvenharrison, line 949: I suggest to clarify the TODO message or remove it

reuvenharrison:
Hi @fenollp,
Can we do anything to help with this PR?

Please note that we expect frequent tweaks on this functionality during the coming months until it is stable, so
perhaps it would be better to implement it in https://github.com/Tufin/oasdiff to speed up the process?

Thank you,
Reuven

fenollp:
Hi @reuvenharrison !
I believe this should be implemented as a schema Walker. So first let's add a `Walk` method so a callback func can be called by the visitor. Then impl a `MergeAllOf` func that can be given to that visitor.
This should provide a simple and extensible API for you and other consumers of this lib (and I have other use cases for that schema walker). Also it helps cut this PR in two parts.

Does this seem reasonable to you?

I don't kind this func to change in the future. I agree its impl could leave in your repo but I'd rather the lib offers this kind of thing from upstream somewhat. But if you prefer let's have the visitor in this lib and you impl your merge in your repo and once it's stable enough for you then open a second PR with your merge impl. That's probably the easiest for everyone.

reuvenharrison:
Hi @fenollp,
Thanks for your quick reply.
We will proceed according to your suggestion above.
Regards,
Reuven

reuvenharrison:
We implemented this in [oasdiff](https://github.com/Tufin/oasdiff) in order to answer the primary requirement ASAP:
https://github.com/Tufin/oasdiff/blob/main/ALLOF.md
 
Happy to assist integration/migration to this repo if anyone wants to lead this effort.

PR #838, opened 2023-09-06
Exclude gopkg.in/yaml.v3 v3.0.0

natehart:
This version contains a nil pointer CVE and until the project dependencies (invopop/yaml and stretcher/testify) update their dependencies, we should exclude the vulnerable versions of the gopkg.in/yaml.v3 from the project.

This commit also upgrades the `marshmallow` and `testify` dependency versions because why not?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,13 +2,19 @@ module github.com/getkin/kin-openapi
 
 go 1.20
 
+exclude (
+	// these versions contain a nil pointer CVE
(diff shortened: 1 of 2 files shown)


natehart:
If (when) this is merged, would you please cut a new release? Due to the indirect inclusion of the vulnerable libraries, I can't get this approved as a third-party dependency by my employer.

Thanks!

Issue #843, opened 2023-09-20
Invalid content-type incorrectly recognised as "request body error"


slonka:
Hi all! This is my first issue here so please be patient 🙏  (I tried searching for a similar issue but couldn't find one so I'm opening a new one).

When sending a POST request with a different content type than defined, e.g. `text-html` to:

```
paths:
  /test:
    post:
      requestBody:
        required: true
        content:
          application/json:
```

the error that is produced is `request body has an error: header Content-Type has unexpected value ""` which is not correct. This should be a parameter error. I think the problem is that `content-type`, `accept` and `authorization` are not in `parameters` list but are [treated differently](https://swagger.io/docs/specification/describing-parameters/#header-parameters).

Please let me know if my reasoning is right.

PR to show the issue: https://github.com/getkin/kin-openapi/pull/844

fenollp:
Hi! Thanks for taking the time to raise an issue AND to bring a failing test :)

I don't see why this request body error should be a parameter error instead.

The bug I see here is `Content-Type has unexpected value ""` which should be mentioning `text/html`, not `""`. 

Wdyt?

slonka:
Hey! 

The way I understand this is that headers are not part of the body. The HTTP request as per HTTP spec:

```
Request       = Request-Line              ; [Section 5.1](https://www.rfc-editor.org/rfc/rfc2616#section-5.1)
                        *(( general-header        ; [Section 4.5](https://www.rfc-editor.org/rfc/rfc2616#section-4.5)
                         | request-header         ; [Section 5.3](https://www.rfc-editor.org/rfc/rfc2616#section-5.3)
                         | entity-header ) CRLF)  ; [Section 7.1](https://www.rfc-editor.org/rfc/rfc2616#section-7.1)
                        CRLF
                        [ message-body ]          ; [Section 4.3](https://www.rfc-editor.org/rfc/rfc2616#section-4.3)
```

So sending a request with an unexpected header should not mention an error in the body because it's not in the body. What is more [OpenAPI docs say](https://swagger.io/docs/specification/describing-parameters/#header-parameters):

```
Note: Header parameters named Accept, Content-Type and Authorization are not allowed. To describe these headers, use the corresponding OpenAPI keywords:
```

So these headers (and other custom headers) are called parameters and that's why I think the correct type of error here should be a `RequestError` with `Parameter` that has `.Name="content-type"`, `.In="header"` and `.Required=true`.

[...]

PR #844, opened 2023-09-20
fix: invalid content type recognised as request body error

slonka:
work in progress, just to better explain the issue [843](https://github.com/getkin/kin-openapi/issues/843)

add test case to show that invalid content type is recognised incorrectly as request body error

closes https://github.com/getkin/kin-openapi/issues/843


fenollp:
Closing as this isn't a bug.

Issue #847, opened 2023-10-17
openapi2conv: v3 formDataBody has extra required array


liankui:
swagger.json into v3.json
```json
{
  "swagger": "2.0",
  "info": {
    "description": "Hello Java Sec API",
    "version": "1.10",
    "title": "Swagger2 RESTful API"
  },
  "host": "localhost:8888",
  "basePath": "/",
  "paths": {
    "/UPLOAD/uploadSafe": {
      "post": {
        "operationId": "singleFileUploadSafeUsingPOST",
        "consumes": [
          "multipart/form-data"
        ],
        "parameters": [
          {
            "name": "file",
            "in": "formData",
            "description": "file",
            "required": true,
            "type": "file"
          }
        ],
        "responses": {
          "200": {
            "description": "OK",
            "schema": {
              "type": "string"
            }
          }
        }
      }
    }
  }
}
```
and the v3.json
```json
{
    "components": {},
    "info": {
        "description": "Hello Java Sec API",
        "title": "Swagger2 RESTful API",
        "version": "1.10"
    },
    "openapi": "3.0.3",
    "paths": {
        "/UPLOAD/uploadSafe": {
            "post": {
                "operationId": "singleFileUploadSafeUsingPOST",
                "requestBody": {
                    "content": {
                        "multipart/form-data": {
                            "schema": {
                                "properties": {
                                    "file": {
                                        "description": "file",
                                        "format": "binary",
                                        "required": [
                                            "file"
                                        ],
                                        "type": "string",
                                        "x-formData-name": "file"
                                    }
                                },
                                "required": [
                                    "file"
                                ],
                                "type": "object"
                            }
                        }
                    }
                },
                "responses": {
                    "200": {
                        "content": {
                            "application/json": {
                                "schema": {
                                    "type": "string"
                                }
                            }
                        },
                        "description": "OK"
                    }
                }
            }
        }
    },
    "servers": [
        {
            "url": "https://localhost:8888/"
        }
    ]
}
```
In v3.json, requestBody.schema.properties has a required array, and schema.properties.file has an extra required array again.

PR #848, opened 2023-10-17
openapi2conv: fix that removes an extra `required` array from `formDataBody`

liankui:
issue #847

### DIFF
--- a/openapi2conv/issue847_test.go
+++ b/openapi2conv/issue847_test.go
@@ -2,59 +2,42 @@ package openapi2conv
 
 import (
 	"context"
-	"encoding/json"
 	"testing"
 
 	"github.com/stretchr/testify/require"
 )
 
 func TestIssue847(t *testing.T) {
-	spec := []byte(`
-{
-  "swagger": "2.0",
-  "info": {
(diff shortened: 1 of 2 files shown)


openapi2conv/issue847_test.go
```diff
@@ -0,0 +1,60 @@
...
+    }
+  }
+}
+`)
+	doc3, err := v2v3JSON(spec)
+	require.NoError(t, err)
+
+	reqStr, err := json.MarshalIndent(doc3, "", "    ")
+	require.NotNil(t, reqStr)
+	require.NoError(t, err)
```
fenollp, line 56: This part seems irrelevant to the test.

openapi2conv/issue847_test.go
```diff
@@ -0,0 +1,60 @@
...
+
```
fenollp, line 57: Please add both checks to `doc3`:
* one looking for the presence of the `required` array where you expect it
* one looking for the absence of the `required` array, you fixed

liankui: I'm done, can you check it again?

PR #850, opened 2023-10-19
openapi3: fix schema validation for `anyOf`, `allOf`, `oneOf` operations

AmadeusK525:
Schemas with `XOF` children that allowed for nullable values had their validation broken, as the `null` check happened before the `XOF` check and would fail immediately, without checking if `XOF` allowed for such nullability

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1072,7 +1072,15 @@ func (schema *Schema) visitJSON(settings *schemaValidationSettings, value interf
 	if schema.IsEmpty() {
 		return
 	}
-	if err = schema.visitSetOperations(settings, value); err != nil {
+
+	if err = schema.visitNotOperation(settings, value); err != nil {
+		return
+	}
+	var run bool
+	if err, run = schema.visitXOFOperations(settings, value); err != nil || !run {
+		return
+	}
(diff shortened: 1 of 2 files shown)


fenollp:
Thanks for this!

micronull:
@AmadeusK525 Your edits broke our tests. I can't tell you exactly why yet.

micronull:
You have all the tests run so the execution process doesn't get to validation.
There are not enough negative test cases in the library.

AmadeusK525:
> You have all the tests run so the execution process doesn't get to validation.
> There are not enough negative test cases in the library. 

Could you provide an example of a broken test, please?

AmadeusK525:
@micronull could you provide an example of a broken test, please?

micronull:
@AmadeusK525 Yes. I have plans for a separate issue.

micronull:
@AmadeusK525 #890

AmadeusK525:
Thank you @micronull, this was a mistake on my part for misunderstanding OpenAPI, it's a super easy fix, I'll open up a PR right now

AmadeusK525:
@micronull Check it out #892
Thanks for the help with the issue :)

Issue #294, opened 2021-01-29
deepObject parameters that are not flat cause panic


jharting:
Consider the following schema:

```yaml
---
openapi: 3.0.3
info:
  title: Reproducer
  version: 0.0.1

paths:
  /example:
    get:
      operationId: Example
      parameters:
      - in: query
        name: filter
        required: false
        style: deepObject
        explode: true
        schema:
          type: object
          properties:
            labels:
              type: object
              additionalProperties:
                type: string
          required: []

      responses:
        '200':
          description: OK
          content:
            application/json:
              schema:
                type: object
```

This should let one make requests with multiple levels of nesting in the filter parameter, e.g.
`/example?filter[labels][foo]=bar`

However, kin-openapi only seems to support flat key/value deepObject parameters. The example above therefore yields

```
echo: http: panic serving 127.0.0.1:58216: schema has non primitive type "object"
goroutine 46 [running]:
net/http.(*conn).serve.func1(0xc0001f1cc0)
        /usr/lib/golang/src/net/http/server.go:1800 +0x139
panic(0x124f3e0, 0xc0003280e0)
        /usr/lib/golang/src/runtime/panic.go:975 +0x3e3
github.com/getkin/kin-openapi/openapi3filter.parsePrimitive(0xc000528039, 0x24, 0xc00035f7c0, 0x6, 0xc000144868, 0xc00042f0b0, 0xc00052802c)
        /home/jharting/go/pkg/mod/github.com/getkin/kin-openapi@v0.36.0/openapi3filter/req_resp_decoder.go:768 +0x86b
github.com/getkin/kin-openapi/openapi3filter.makeObject(0xc00042f0b0, 0xc00035f6e0, 0x0, 0x0, 0xc000328090)
        /home/jharting/go/pkg/mod/github.com/getkin/kin-openapi@v0.36.0/openapi3filter/req_resp_decoder.go:709 +0x127
github.com/getkin/kin-openapi/openapi3filter.(*urlValuesDecoder).DecodeObject(0xc000130030, 0xc0006a1a16, 0x6, 0xc00068e100, 0xc00035f6e0, 0xc000328090, 0xc000536160, 0xc000542df0)
        /home/jharting/go/pkg/mod/github.com/getkin/kin-openapi@v0.36.0/openapi3filter/req_resp_decoder.go:554 +0xd7
[...]

PR #855, opened 2023-11-21
fix(filters): allow oneOf inside deepObject

Koumbaya:
Hello,
We encountered the same problems as in (#294 ),
This fixes it, and doesn't break other tests, but... yeah it might not be the ideal place to make the fix, I think it should be done earlier in the callstack, but I'm not sure where ?

I know this is kind of a gray area of the spec but this schema is considered valid by our parser in Typescript, but panics in go so I figured it should be valid ?

edit: I haven't added allOf or anyOf because I think it may be the wrong place to put that in the first place, but I could add it there easily I think ?

### DIFF
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -922,6 +922,14 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
 	case "string":
 		return raw, nil
 	default:
+		if len(schema.Value.OneOf) > 0 {
+			for i := range schema.Value.OneOf {
+				res, err := parsePrimitive(raw, schema.Value.OneOf[i])
+				if err == nil {
+					return res, nil
+				}
+			}
+		}
 		panic(fmt.Sprintf("schema has non primitive type %q", schema.Value.Type))
 	}
 }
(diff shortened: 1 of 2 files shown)


Koumbaya:
:( yeah no it won't work now that I think about it because it doesn't enforce the rules of the oneOf anyway (if I put something else than "eq" or "eoq" it still passes the tests)

openapi3filter/req_resp_decoder.go
```diff
@@ -922,6 +922,14 @@ func parsePrimitive(raw string, schema *openapi3.SchemaRef) (interface{}, error)
 	case "string":
 		return raw, nil
 	default:
+		if len(schema.Value.OneOf) > 0 {
```
fenollp, line 925: I believe this line can be dropped.

fenollp:
> it still passes the tests
ah

fenollp:
The need for an intermediate internal representation of schemas comes up once again. Other endeavours require that as well so a comprehensive solution is on the way.

Now if you can patch this enough that it passes your and this lib 's tests I'm open to merge it :)

fenollp:
PTAL at https://github.com/getkin/kin-openapi/pull/911

PR #880, opened 2023-11-28
openapi3: upgrade Responses.Get

fenollp:


fenollp:
Actually kept in https://github.com/getkin/kin-openapi/pull/763

Issue #884, opened 2023-12-08
binding parameter default value with allOf schema


podhmo:
Default value binding does not work well when parameter schema has allOf.

For example, when I set a custom default value while referencing the schema with `$ref`, we can write it using allOf in OpenAPI 3.x.

```yaml
components:
  schemas:
    TaskSortEnum:
      enum:
        - createdAt
        - -createdAt
        - updatedAt
        - -updatedA
paths:
  /tasks:
    get:
      description: ""
      operationId: ListTask
      parameters:
        - in: query
          name: sort
          schema:
            allOf:
              - $ref: '#/components/schemas/TaskSortEnum'
              - default: -createdAt
```

But, this doesn't work in the current implementation.

Related oapi-codegen's issue is here.

- https://github.com/deepmap/oapi-codegen/issues/1379

PR #885, opened 2023-12-08
feat: support default value binding with allOfSchema

podhmo:
- fix https://github.com/getkin/kin-openapi/issues/884

📝 nested complex structure is not supported with this PR

### DIFF
--- a/openapi3filter/issue884_test.go
+++ b/openapi3filter/issue884_test.go
@@ -1,0 +1,112 @@
+package openapi3filter
+
+import (
+	"net/http"
+	"net/url"
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/routers/gorillamux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIssue884(t *testing.T) {
+	loader := openapi3.NewLoader()
+	ctx := loader.Context
+	spec := `
+  openapi: 3.0.0
+  info:
+    version: 1.0.0
+    title: Sample API
+  components:
+    schemas:
+      TaskSortEnum:
+        enum:
+          - createdAt
+          - -createdAt
+          - updatedAt
+          - -updatedAt
+  paths:
+    /tasks:
+      get:
+        operationId: ListTask
+        parameters:
+          - in: query
+            name: withDefault
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/TaskSortEnum'
+                - default: -createdAt
+          - in: query
+            name: withoutDefault
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/TaskSortEnum'
+          - in: query
+            name: withManyDefaults
+            schema:
+              allOf:
+                - default: -updatedAt
+                - $ref: '#/components/schemas/TaskSortEnum'
+                - default: -createdAt
+        responses:
+          '200':
+            description: Successful response
+  `[1:]
+
+	doc, err := loader.LoadFromData([]byte(spec))
+	require.NoError(t, err)
+
+	err = doc.Validate(ctx)
+	require.NoError(t, err)
+
+	router, err := gorillamux.NewRouter(doc)
+	require.NoError(t, err)
+
+	tests := []struct {
+		name          string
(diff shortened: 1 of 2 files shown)


podhmo:
need tests?

openapi3filter/validate_request.go
```diff
@@ -136,24 +136,35 @@ func ValidateParameter(ctx context.Context, input *RequestValidationInput, param
...
-			q.Add(parameter.Name, fmt.Sprintf("%v", value))
-			req.URL.RawQuery = q.Encode()
-		case openapi3.ParameterInHeader:
-			req.Header.Add(parameter.Name, fmt.Sprintf("%v", value))
-		case openapi3.ParameterInCookie:
-			req.AddCookie(&http.Cookie{
-				Name:  parameter.Name,
-				Value: fmt.Sprintf("%v", value),
-			})
+		if schema.AllOf != nil {
```
fenollp, line 141: superfluous `if` considering the following `for`

fenollp (changes requested):
> need tests?

Well, yeah. Please think proactively about this free software thing.
Here it's clear you should at the very least add a test for the behavior you're changing, so add a `func TestIssue884` in `openapi3filter/issue884_test.go` that makes sure a default can be extracted from the allof schema you mention elsewhere.

I'd also add a test that shows the `break` behavior, i.e.: the first non-nil default of allofs is the one picked.

Thanks for the patch :)

openapi3filter/validate_request.go
podhmo: ok

podhmo:
I added tests

openapi3filter/issue884_test.go
```diff
@@ -0,0 +1,113 @@
...
+                - default: -createdAt
+          - in: query
+            name: withoutDefault
+            schema:
+              allOf:
+                - $ref: '#/components/schemas/TaskSortEnum'
+          - in: query
+            name: withManyDefaults
+            schema:
+              allOf:
```
podhmo, line 50: this definition should be schema validation error ?

openapi3filter/issue884_test.go
```diff
@@ -0,0 +1,113 @@
...
+				SkipSettingDefaults: true,
+			},
+			expectedQuery: url.Values{},
+		},
+
+		{
+			name: "defaults are added to requests",
+			expectedQuery: url.Values{
+				"withDefault":      []string{"-createdAt"},
+				"withManyDefaults": []string{"-updatedAt"}, // first default is win
```
podhmo, line 85: Come to think of it, it would be more natural for the last value to be win when schema validation errors is not raised?

podhmo: To be honest, I don't know which choice is better.

- The first value is win (low cost)
- The last value is win (more natural)

podhmo: ( In oapi-codegen, the allOf has multiple default values ​​will cause an error when merging schema phase, but this is oapi-codegen's story. )

fenollp: AFAICT the spec is blurry here. 
So to me either choice is fine as long as it's mentioned in the field docs. 

My vote for using the earliest default for now (shorter code path). When the need arrives, then we can pick another way to read this, like a func for instance. 
You decide :)

openapi3filter/issue884_test.go
fenollp: I'm unable to find in the spec à justification to "forbid multiple schema defaults". But if you find one then yes let's impl this validation pass. 

Otherwise let's continue to not return a validation error here.

fenollp:
LGTM. If you feel this in a good state as is I'll merge.

fenollp (approved):

openapi3filter/issue884_test.go
podhmo: OK I also like "first default is win".

(As an extension of the "last default is win" specification, if deep nested structures were to be supported, a complete full scan would be required)

openapi3filter/issue884_test.go
podhmo: Yes, it is not required according to the specifications.

podhmo:
Yes, it's ready 🚀

PR #886, opened 2023-12-08
feat(openapi3filter): add ExludeRequestQuery

zekth:
Proposal of remediation of https://github.com/getkin/kin-openapi/pull/855

Until we support the `deepObject` validation; here's a proposal of adding the skip of query like it's done for `body`.

This would offload the validation to the application layer

If this is ok for you @fenollp i'll add a test suite to validate the behavior

### DIFF
--- a/openapi3filter/options.go
+++ b/openapi3filter/options.go
@@ -7,6 +7,9 @@ type Options struct {
 	// Set ExcludeRequestBody so ValidateRequest skips request body validation
 	ExcludeRequestBody bool
 
+	// Set ExcludeRequestQueryParams so ValidateRequest skips request query params validation
+	ExcludeRequestQueryParams bool
+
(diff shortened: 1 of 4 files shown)


openapi3filter/options.go
```diff
@@ -7,6 +7,9 @@ type Options struct {
 	// Set ExcludeRequestBody so ValidateRequest skips request body validation
 	ExcludeRequestBody bool
 
+	// Set ExcludeRequestQuery so ValidateRequest skips request query params validation
+	ExcludeRequestQuery bool
```
fenollp, line 11: LGTM. yup add a test that makes sure the flag works
I'd also prefer a more precise name such as `ExcludeRequestQueryParams`

zekth: Ok cool, i'll update the PR then

zekth:
@fenollp updated and ready to review

Issue #900, opened 2024-01-29
v3: allow extensions next to $ref in SchemaRef


paulmach:
I would like to extent `SchemaRef` to have a `Extensions` or `RefExtensions` to capture extensions next to $ref.

For example, if I have:
```
type: object
properties:
 id:
   $ref: "#/components/schemas/id"
   x-order: 1
 name:
   type: string
```
I would like to capture the `x-order: 1` within `SchemaRef.Extensions["x-order"]`

Looking at [the code](https://github.com/getkin/kin-openapi/blob/7d030b2b2bc170f1931e929b5401f785afe75a69/openapi3/refs.go#L590-L599) it looks like it's "if `$ref` only capture $ref"

I don't think the openapi v3 spec official supports this but it would be very helpful. 

My use-case is specifying the `x-order`, a new feature added to [oapi-codegen](https://github.com/deepmap/oapi-codegen). For example:
```
type: object
properties:
 start_port:
   $ref: "#/components/schemas/port"
   x-order: 0
 end_port:
   $ref: "#/components/schemas/port"
   x-order: 1
```

It works if you specify the `x-order` in the referenced object, but in the above it needs to be context dependent. A change would need to be made to oapi-codegen but it looks to be blocked by this issue.

If this change is okay, I can work on the PR.

paulmach:
I created a pr to add Extension to `SchemaRef`. If this approach is agreed I can expand to all *Refs.

https://github.com/getkin/kin-openapi/pull/901

paulmach:
I have verified that https://github.com/getkin/kin-openapi/pull/901 adds the support needed to fix my use-case in oapi-codegen

PR #901, opened 2024-01-30
openapi3: allow Extensions next to $ref in SchemaRef

paulmach:
See https://github.com/getkin/kin-openapi/issues/900 for full details and use-case. If this approach is agreed upon, I can extend to all *Refs (e.g. CallbackRef, ExampleRef, ResponseRef, etc.)

Captures `x-order` into `SchemaRef.Extensions`
```
type: object
properties:
 id:
   $ref: "#/components/schemas/id"
   x-order: 1
 name:
   type: string
```

Questions, should only extra attributes starting with 'x-' be captured?

I made it so yes, only extensions (those starting with 'x-') are captured. This is because extra attributes are technically not allowed by the spec, but also to maintain the way `JSONLookup` works. For example:
```
schemas:
  id:
    type: integer
    x-order: 2
  thing:
    type: object
    properties:
     thing_id:
       $ref: "#/components/schemas/id"
       type: string
       x-order: 1
     name:
       type: string
```
Doesn't change: `thingIDRef.JSONLookup("type")` will continue to return "integer" as overriding the `type` next to the ref is not allowed.

Does change: `thingIDRef.JSONLookup("x-order")`. Today, 2, as it goes back to the $ref definition. After this change it'll be `1` (value next to the $ref).

`Validate()` behavior doesn't change. `ref.Validate(ctx, AllowExtraSiblingFields("x-order"))` for it to be okay. I think this is correct as technically the spec doesn't allow stuff next to $ref.


paulmach:
Oh snap, looks like refs.go is auto generated. Please see if you like this change, then I make the change to `refs.sh`.

openapi3/refs.go
```diff
@@ -559,6 +560,10 @@ func (x *ResponseRef) JSONLookup(token string) (interface{}, error) {
 // SchemaRef represents either a Schema or a $ref to a Schema.
 // When serializing and both fields are set, Ref is preferred over Value.
 type SchemaRef struct {
+	// Extensions only captures fields starting with 'x-' as no other fields
+	// are allowed by the openapi spec.
+	Extensions map[string]interface{}
```
fenollp, line 565: I'm not sure this should live here: this field already exists in the `.Value`. Or maybe it'd make more sense to move `.Value.Extensions` here, alongside `.Ref` `.Value` and `.extra`. This makes sense to me. You?

fenollp: (this change alone would be its own PR)

openapi3/refs_test.go
```diff
@@ -332,3 +334,65 @@ components:
...
+
+	// captures extension
+	ref := doc.Components.Schemas["PetsXOrder"].Value.Items
+	assert.Equal(t, "petsXOrder", ref.Extensions["x-order"])
+	assert.Equal(t, "pet", ref.Value.Extensions["x-order"])
+
+	err = ref.Validate(context.Background())
+	require.EqualError(t, err, `extra sibling fields: [x-order]`)
+
+	err = ref.Validate(context.Background(), AllowExtraSiblingFields("x-order"))
```
fenollp, line 378: An other `func Allow/DisallowExtensionsInRefPosition() ValidationOption { .. }` would be nice (names suggestions welcomed)

openapi3/refs_test.go
```diff
@@ -332,3 +334,65 @@ components:
 	_, _, err = ptr.Get(doc)
 	require.Error(t, err)
 }
+
+func TestSchemaRef_Extensions(t *testing.T) {
```
fenollp, line 338: Yup please integrate with `ref.sh` and also have this test be generated per type as much as you can

fenollp (changes requested):

openapi3/refs.go
paulmach: I did a quick test and it looks like `ref.Value` objects are reused for all the places they are referenced. Thus updating them will change them everywhere.

I personally think we should keep it as proposed here. This allows users to know where the extension came from. They can use `schemaRef.JSONLookup("x-....")` to do the most common thing, check ref, then fallback to Value.

paulmach:
I made the following changes:
* moved tests in ref_test.go into refs_issue222_test.go and refs_issue247_test.go
* added validation options `AllowExtensionsWithRef` (default) and `ProhibitExtensionsWithRef`
* updated refs.sh to generate the code
* add refs_test.sh to generate tests
* reran docs.sh and added refs_test.sh as part of the github workflow.

I think the only "question" is if `ref.Extensions` should capture `^x-` fields or all extra fields. As mentioned above, there could be some impact for "weird" schemas if we capture all fields. I propose to only capture `^x-` fields for now as more backwards compatible. We can always change it based on feedback.

fenollp:
This is great work thank you!

WRT non x-.. in Extensions you're right: let's just keep the x-.. keys in here for now anyway

paulmach:
I rebased on the latest master. Hoping to get this into the next release.

paulmach:
Hi @fenollp, I rebased again on the latest master. Hoping to get this into the next release.

fenollp (approved):

PR #905, opened 2024-02-08
openapi3filter: add `Unwrap()` method to SecurityRequirementsError

nickajacks1:
Allows users to perform checks like `errors.Is` which may be returned from the given AuthenticationFunc.


fenollp:
> Allows users to perform checks like errors.Is which may be returned from the given AuthenticationFunc.

Please add a test showing this behavior (an Example would be best)

nickajacks1:
I've added an Example. Please let me know if it wasn't what you had in mind!

Issue #909, opened 2024-02-12
#/components/schemas across routes


EnriqueL8:
I have a set of routes that share a child schema such as one API returns an array of objects and another API returns one object. The object itself has the same schema but when I generate the OpenAPI I get duplicate values for that schema and do not get the schema stored under `#/components/schemas`. I might be doing something wrong here... so any insights appreciated 😄 

I'm using `NewSchemaRefForValue` for each return object from my API. It will get called either with the array of objects or the objects. With the schema ref it returns I will append it to the document I'm constructing. 

I see that for a new schema do be added to `#/components/schemas` it needs to be a cyclical dependency? But as far as I can see that only happens in the context of that specific schema as it checks for parents. How do I provide other schemas in the OpenAPI that I have parsed?

Thanks in advance!

sskserk:
Hi @EnriqueL8, @fenollp,

First of all thank you for a good project, it really helped a lot!

I also face the same obstacle (not a problem). Is it possible anyhow to generate a ref for a nested object? Right now it is only generated if there is recursion. Schemes for complex objects generated as nested sub-properties:

Example:
```json
User:
      properties:
        attributesMapping:
          properties:
            email:
              type: string
            login:
              type: string
            name:
              type: string
            secondName:
              type: string
          type: object
        name:
          type: string
        lastName:
          type: string
      type: object
```

@fenollp Is it possible instead to generate two schemes and refer one from another?

Thank you! 

P/S: I'm ready to bring a MR if you point where it can be fixed.

EnriqueL8:
@sskserk I raised a draft PR https://github.com/getkin/kin-openapi/pull/914 and will add more validation and see what the maintainers think!

fenollp:
Hi! I welcome such changes.
My only requirement is that it's tested and the new option behaves well with the others (and their combinations).

So yeah, please work together on that PR and once you both are satisfied I'll help you merge it.

PR #914, opened 2024-02-22
openapi3gen: add `CreateComponentSchemas` option to export object schemas to components

EnriqueL8:
PR to start the discussion on https://github.com/getkin/kin-openapi/issues/909


sskserk:
@EnriqueL8 let me try to test it and thus help you.

sskserk:
@EnriqueL8 Thanks!

EnriqueL8:
There is a problem with pointers where the name of struct is not defined properly - fixing

EnriqueL8:
Thanks - just making a few more change and will run it

sskserk:
@EnriqueL8,

I'm trying to compile my project having replaced original kin-openapi with your branch.

Observe the problem:

_../../../go/pkg/mod/github.com/davidebianchi/gswagger@v0.9.0/main.go:115:19: cannot use openapi3.Paths{} (value of type openapi3.Paths) as *openapi3.Paths value in assignment
../../../go/pkg/mod/github.com/davidebianchi/gswagger@v0.9.0/operation.go:31:22: invalid argument: cannot make openapi3.Responses; type must be slice, map, or channel
../../../go/pkg/mod/github.com/davidebianchi/gswagger@v0.9.0/route.go:101:29: invalid argument: cannot make openapi3.Responses; type must be slice, map, or channel_


I guess it because the gswagger@v0.9.0 refers to the incompatible version of kin-openapi

sskserk:
@EnriqueL8 there is one problem. If there are different types in different packages have the same name then the "NewSchemaRefForValue" produces one component, the same ref.


Example:
```golang
package root:

type AnotherStruct struct {
  Field1 string `json:"field1"`
  Field2 string `json:"field2"`
  Field3 string `json:"field3"`
}
type RecursiveType struct {
  Field1        string        `json:"field1"`
  Field2        string        `json:"field2"`
  Field3        string        `json:"field3"`
  AnotherStruct AnotherStruct `json:"children,omitempty"`
  **AnotherStruct2         subpkg.AnotherStruct    `json:"children,omitempty"`**
}
```

```golang
package subpkg:

type AnotherStruct struct {
  Field4 string `json:"field1"`
}
```

EnriqueL8:
Aah I see - is that something that we should document as a limitation as it seems quite improbable? The same thing would occur if that type is cyclic

sskserk:
@EnriqueL8, I think it makes to supply an optional type name generator and use it at openapi3gen.go:315 ... diving into the code

sskserk:
@EnriqueL8 please take a look at your forked repository. I have create a MR into your developed branch. I propose a new option making it possible to customize the generated type name

EnriqueL8:
thanks @sskserk I'll take a look

EnriqueL8:
Added your commit

sskserk:
@EnriqueL8 let me create a couple of additional tests. I'll bring an additional MR into the current branch.

EnriqueL8:
@sskserk awesome - I'm also testing this

.github/docs/openapi3gen_pack.txt
```diff
@@ -0,0 +1,9 @@
+package pack // import "github.com/getkin/kin-openapi/openapi3gen/pack"
```
EnriqueL8, line 1: Ah we don't want to include this in the docs but you can't export a struct from a test package

EnriqueL8:
I'm wondering if we should have a customiser for ignoring components instead of only top level? will leave to the reviewer to decide @fenollp fyi

EnriqueL8:
@sskserk I've found another issue here when you have a generic field, ignoring the top level is not enough if it's nested so you want to be able to not add generics to component schemas as you will get one per type used

sskserk:
@EnriqueL8 , hm, good idea. 

Let me create a test ....

EnriqueL8:
> I see that for generic attributes the components are not created if the "IgnoreGenerics" is set to True
Let me test a generic within a component and not at the top level

sskserk:
@EnriqueL8 ,
can it be that imports sequence is improperly formatted?

E.g:

import (
	"github.com/getkin/kin-openapi/openapi3gen"
	"github.com/getkin/kin-openapi/openapi3gen/subpkg"
	"testing"
	"time"
)

fenollp:
The import sequense was an issue, now it's `go vet` that's complaining:
> Error: openapi3gen/openapi3gen.go:398:77: invalid operation: schema.Type != "object" (mismatched types *openapi3.Types and untyped string)

EnriqueL8:
@fenollp the weird thing is that `go vet` passes locally, so I can't figure out why it's failing in the build

openapi3gen/openapi3gen.go
```diff
@@ -104,8 +129,13 @@ func (g *Generator) NewSchemaRefForValue(value interface{}, schemas openapi3.Sch
...
 	for ref := range g.SchemaRefs {
-		if _, ok := g.componentSchemaRefs[ref.Ref]; ok && schemas != nil {
-			schemas[ref.Ref] = &openapi3.SchemaRef{
+		refName := ref.Ref
+		if g.opts.exportComponentSchemas.ExportComponentSchemas && strings.HasPrefix(ref.Ref, "#/components/schemas/") {
+			refName = strings.TrimPrefix(ref.Ref, "#/components/schemas/")
+		}
+
+		if _, ok := g.componentSchemaRefs[refName]; ok && schemas != nil {
+			schemas[refName] = &openapi3.SchemaRef{
```
EnriqueL8, line 138: There is a problem here with the schema customizer... if we have the same schema called MyObject that we have already parsed for one path and then another path has that same object but it has a schema customizer that adds a field it overrides the existing object for the previous path... Is the solution here just that the user needs to have two different structs? As far as I know there is no way to extend a schema within an operation?

sskserk: @EnriqueL8 I'm not sure what you meant.

May I ask you please to provide a couple of samples so that I could help you?

openapi3gen/openapi3gen.go
```diff
@@ -355,9 +395,40 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 		}
 	}
 
+	if !g.opts.exportComponentSchemas.ExportComponentSchemas || schema.Type != "object" {
```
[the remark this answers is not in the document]
EnriqueL8: aah I see - let me fix that

openapi3gen/openapi3gen.go
EnriqueL8: So if you have two different paths in your api server:
`POST /object -> returns an object`
`PATCH /object -> returns an object with an extra field that contains an array of changes made by the patch`
If the object in Golang is declared as:
```
type Object struct {
     field1 string `json:"field1"`
     field2 string `json:"field2"`
     patchedFields []string `json:"patchedFields,omitempty"`
}
```

Then when the OpenAPI document gets generated the second `PATCH` operation will override `#/components/schemas/object` and add the `patchedFields` field and it will appear in the `POST /object` operation.

It would be the same case if a user supplies an schema customizer and adds a field at that point.

The solution could be that it's opinionated and they should be two structs:
```
type Object struct {
     field1 string `json:"field1"`
     field2 string `json:"field2"`
}
```
```
type ObjectWithPatchedFields struct {
     Object
     patchedFields []string `json:"patchedFields,omitempty"`
}
```


I don't if there is another way?

sskserk: @EnriqueL8 , I think it is not a technical problem. A developer must be aware and don't create different structs with the same name. Also, it is already possible to customize to component names with the names generator. I think it is already flexible enough.

What do you think on that?

openapi3gen/openapi3gen.go
```diff
@@ -355,9 +395,40 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 		}
 	}
 
+	if !g.opts.exportComponentSchemas.ExportComponentSchemas || schema.Type.Is("object") {
```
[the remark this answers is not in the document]
EnriqueL8: yes ! I just fixed it

sskserk:
@EnriqueL8 & @fenollp , let me just share the news. A related project has been fixed as well https://github.com/davidebianchi/gswagger/pull/148

EnriqueL8:
This is ready for review

EnriqueL8:
@fenollp I'm seeing a lot of issues with `found unresolved ref` , it seem that the package so far made an assumption that all ref are resolved? So no reference should exist in the final document?

EnriqueL8:
I've noticed another issue with `map[string]interface{}` as we are using the type `object` to export the schema. Ideally we shouldn't be exporting the top level maps to schemas.

@fenollp if you get a chance would appreciate a first review as well 😃

EnriqueL8:
Alright pushed up fix for maps, only perform the export for structs

sskserk:
@fenollp do you think this MR is ready to be merged?

openapi3gen/openapi3gen.go
EnriqueL8: I'm wondering if a better solution would be to merge schemas on the schema customiser if it exists as a component?

EnriqueL8: I'm thinking specifically for things such as `readOnly` and `writeOnly` or any other properties on the schema that need to be preserved

EnriqueL8:
@fenollp any bandwidth to take a look at this? Note would appreciate your input on https://github.com/getkin/kin-openapi/pull/914#discussion_r1524773005

openapi3gen/openapi3gen.go
fenollp: Maybe https://github.com/getkin/kin-openapi/pull/920 can help here ;)

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +591,413 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+
+	numberOfDetectedSchemas := 0
+	for schemaName, _ := range schemas {
+		switch schemaName {
+		case parentSchemaName,
+			childSchemaName:
+			numberOfDetectedSchemas++
+		default:
+			// unknown schema
+			t.Errorf("Unknown schema: %s", schemaName)
```
fenollp, line 674: `panic(..)` is fine in examples code

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +591,413 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+		}
+	}
+
+	if numberOfDetectedSchemas != 2 {
+		t.Errorf("Required number of expected schema names was not produced. Want 2, have %d", numberOfDetectedSchemas)
```
fenollp, line 679: why bother asserting count when full contents are matched above

openapi3gen/openapi3gen.go
```diff
@@ -291,6 +321,14 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 			schema.Type = &openapi3.Types{"string"}
 			schema.Format = "date-time"
 		} else {
+			typeName := g.generateTypeName(t)
+
+			if g.opts.exportComponentSchemas.ExportComponentSchemas && g.componentSchemaRefs[typeName] != struct{}{} {
```
fenollp, line 326: Please use `ok, _ := g.componentSchemaRefs[typeName]; ok {` syntax, this allows moving the `typeName :=` within the option check.

openapi3gen/openapi3gen.go
```diff
@@ -340,13 +378,17 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
...
 				}
+
 			}
 
 			// Object only if it has properties
 			if schema.Properties != nil {
 				schema.Type = &openapi3.Types{"object"}
 			}
+
+			// TODO move the export here!
```
fenollp, line 389: yes?

openapi3gen/openapi3gen.go
```diff
@@ -355,9 +397,40 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 		}
 	}
 
+	if !g.opts.exportComponentSchemas.ExportComponentSchemas || t.Kind() != reflect.Struct {
+		return openapi3.NewSchemaRef(t.Name(), schema), nil
+	}
+
+	// Best way I could find to check that
+	// this current type is a generic
+	isGeneric, err := regexp.Match(`^.*\[.*\]$`, []byte(t.Name()))
```
fenollp, line 406: How about just checking that the last rune is `]`?

openapi3gen/openapi3gen.go
```diff
@@ -47,6 +58,12 @@ func UseAllExportedFields() Option {
 	return func(x *generatorOpt) { x.useAllExportedFields = true }
 }
 
+func CreateTypeNameGenerator(tngnrt TypeNameGenerator) Option {
+	return func(x *generatorOpt) {
+		x.typeNameGenerator = tngnrt
+	}
```
fenollp, line 64: nit: please single-line this

openapi3gen/openapi3gen_test.go
fenollp, line 679: why bother asserting count when full contents are matched above

fenollp (changes requested):
Mostly nitpicks, LGTM!

openapi3gen/subpkg/sub_type.go
```diff
@@ -0,0 +1,5 @@
+package subpkg
```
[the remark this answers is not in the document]
EnriqueL8: sure will move it!

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +591,413 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+	var data []byte
+	if data, err = json.MarshalIndent(&schemas, "", "  "); err != nil {
+		panic(err)
+	}
+	fmt.Printf("schemas: %s\n", data)
+	if data, err = json.MarshalIndent(&schemaRef, "", "  "); err != nil {
+		panic(err)
+	}
+	fmt.Printf("schemaRef: %s\n", data)
+	// Output
```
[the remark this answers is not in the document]
EnriqueL8: awesome will do

openapi3gen/openapi3gen_test.go
EnriqueL8: fair enough

openapi3gen/openapi3gen.go
EnriqueL8: would be an array in that case?

EnriqueL8:
@fenollp thanks for the review! Will push changes up soon

openapi3gen/openapi3gen.go
sskserk: @EnriqueL8 comment fixed. There is a PR into your repo

openapi3gen/openapi3gen.go
EnriqueL8: not needed

openapi3gen/subpkg/sub_type.go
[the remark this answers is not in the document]
EnriqueL8: Solved it by modifying .docs.sh

openapi3gen/openapi3gen.go
EnriqueL8: Have modified it, not sure what you mean by _moving the `typeName :=` within the option check._

EnriqueL8:
@fenollp Ready for re-review 😃

openapi3gen/openapi3gen.go
fenollp: ah yes, then `runes[-1] == ']' && runes[-2] != '['`

fenollp (changes requested):
Looks like it's all down to that regex thing and we're good :)

fenollp:
Hey there! Please rebase & I'll make this move along.

fenollp (changes requested):
Rebase please

EnriqueL8:
@fenollp Thanks, sorry was away and not available to rebase! It's a shame that the commit was squashed and not awarded to the contributors of the PR

fenollp:
Hi @EnriqueL8, you're right. I messed up while rebasing your PR and the shared attribution went to someone that's not even taken part in that part of the lib. I should have checked, sorry.

I do not mean to steal anyone's work and would very much prefer to give attribution where it's deserved (e.g. here). So I suggest this: send a PR that both reverts then applies your commit (add some changes if you want but that's not needed). At least this way attribution is assured. If you have other ideas, I'm open :)

EnriqueL8:
Sounds good - I'll do that!

EnriqueL8:
@fenollp I think this solves it https://github.com/getkin/kin-openapi/pull/937

PR #918, opened 2024-03-04
fix: update regex to avoid replacing literals without `\u`

thiagownt:


thiagownt:
@fenollp this causes errors on validation of OpenAPI like 
```
validating OpenAPI failed: invalid components: schema "SomeRequest": error parsing regexp: invalid repeat count: `{6789}`
```

fenollp:
Thnks!

Issue #919, opened 2024-03-06
openapi3gen: Allow client with custom type to supply their own type(s)


taufik-rama:
Hello :wave: 

I'm currently trying stuff out using `openapi3gen` package & I noticed a functionality to generates OpenAPI's types based on the struct fields

https://github.com/getkin/kin-openapi/blob/7aa9f7ee3d530160cd6479d28578a40b290d008f/openapi3gen/openapi3gen.go#L198-L206

I've an issue where the resulting object type is "unknown" because I uses newtype/custom type for my type

```golang
package mypkg

type ID uuid.UUID

type T struct {
    ID ID `json:"id"`
}
```

So when I generates `T`, it will result in something like
```yaml
# ...
paths:
    /my-api:
        delete:
            requestBody:
                content:
                    application/json:
                        schema:
                            properties:
                                id: {}
# ...
```

The expectation is that the type should instead refer to the "actual" type, or other possibility would be from the user to also generate custom schema definitions

```yaml
# ...
paths:
    /my-api:
        delete:
            requestBody:
                content:
                    application/json:
                        schema:
                            properties:
                                id:
                                    type: string
# ...
```

## My current solution

I made some changes to above type mapping & provides a `SetSchemar` interface, which would supply the `openapi3.Schema` to the user, which they then can set themselves based on their requirement

```patch
diff --git a/openapi3gen/openapi3gen.go b/openapi3gen/openapi3gen.go
index 95805ee..a301337 100644
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -35,6 +35,13 @@ type Option func(*generatorOpt)
 // the final output
 type SchemaCustomizerFn func(name string, t reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error
 
+// SetSchemar allows client to set their own schema definition according to
+// their specification. Useful when some custom datatype is needed and/or some custom logic
+// is needed on how the schema values would be generated
+type SetSchemar interface {
+	SetSchema(*openapi3.Schema)
+}
+
 type generatorOpt struct {
 	useAllExportedFields bool
 	throwErrorOnCycle    bool
@@ -347,6 +354,14 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 				schema.Type = &openapi3.Types{"object"}
 			}
 		}
+
+	default:
+		// Object has their own schema's implementation, so we'll use those
[...]

PR #920, opened 2024-03-06
openapi3gen: allow overriding how a Schema is generated

taufik-rama:
Implementation of #919

### DIFF
--- a/openapi3gen/openapi3gen_test.go
+++ b/openapi3gen/openapi3gen_test.go
@@ -590,3 +590,32 @@ func ExampleNewSchemaRefForValue_recursive() {
 	//   "type": "object"
 	// }
 }
+
+type ID [16]byte
+
+// T implements SetSchemar, allowing it to set an OpenAPI schema.
+type T struct {
+	ID ID `json:"id"`
+}
+
+func (_ *ID) SetSchema(schema *openapi3.Schema) {
+	schema.Type = &openapi3.Types{"string"} // Assuming this matches your custom implementation
+	schema.Format = "uuid"
+}
+func TestNewSchemaRefForValueWithSetSchemar(t *testing.T) {
+	schemas := make(openapi3.Schemas)
+	instance := &T{
+		ID: ID{},
+	}
+
+	// Generate the schema for the instance
+	schemaRef, err := openapi3gen.NewSchemaRefForValue(instance, schemas)
+	require.NoError(t, err, "should not error when generating schema ref")
+
+	// Optional: Marshal and inspect the schemas and schemaRef if needed for further verification
+	_, err = json.MarshalIndent(schemas, "", "  ")
+	require.NoError(t, err, "error marshaling schemas")
+
+	_, err = json.MarshalIndent(schemaRef, "", "  ")
+	require.NoError(t, err, "error marshaling schemaRef")
+}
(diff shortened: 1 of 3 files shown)


fenollp (changes requested):
Hey! LGTM, just: commit `./docs.sh` and add a test or better: an example that shows your use case.

taufik-rama:
Thanks for the input, tagging @blackhun11 for his implementation, feel free to re-review those anytime @fenollp

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +590,32 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+// T implements SetSchemar, allowing it to set an OpenAPI schema.
+type T struct {
+	ID ID `json:"id"`
+}
+
+func (_ *ID) SetSchema(schema *openapi3.Schema) {
+	schema.Type = &openapi3.Types{"string"} // Assuming this matches your custom implementation
+	schema.Format = "uuid"
+}
+func TestNewSchemaRefForValueWithSetSchemar(t *testing.T) {
```
fenollp, line 605: Please turn this into an example so usage of `SetSchema` gets examples in the docs

This name example's name should include the name of the interface you introduce.

https://go.dev/blog/examples

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +590,32 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+	// Generate the schema for the instance
+	schemaRef, err := openapi3gen.NewSchemaRefForValue(instance, schemas)
+	require.NoError(t, err, "should not error when generating schema ref")
+
+	// Optional: Marshal and inspect the schemas and schemaRef if needed for further verification
+	_, err = json.MarshalIndent(schemas, "", "  ")
+	require.NoError(t, err, "error marshaling schemas")
+
+	_, err = json.MarshalIndent(schemaRef, "", "  ")
+	require.NoError(t, err, "error marshaling schemaRef")
```
fenollp, line 620: There should be at least one call to `require.JSONEq(t, .., ..)` with the serialized json data, showing that `format` is indeed set, possibly that it isn't set any wrong place, ...

Feel free to copy from https://github.com/getkin/kin-openapi/issues/919

fenollp (changes requested):

openapi3gen/openapi3gen_test.go
blackhun11: hi @fenollp thank you for the review! i just change my `Test` to use `Example` and already put the output also. kindly re-review it again. thank you

openapi3gen/openapi3gen_test.go
```diff
@@ -590,3 +590,43 @@ func ExampleNewSchemaRefForValue_recursive() {
...
+type T struct {
+	ID ID `json:"id"`
+}
+
+func (_ *ID) SetSchema(schema *openapi3.Schema) {
+	schema.Type = &openapi3.Types{"string"} // Assuming this matches your custom implementation
+	schema.Format = "uuid"
+}
+
+func ExampleID_SetSchema() {
```
fenollp, line 606: Almost! See here you define an `Example` for the `SetSchema` method of the `ID` type. It doesn't help because these type and method live in test files so don't end up in the generated docs. Instead this should be named `ExampleSetSchemar`.

I'll amend this myself, no worries.

fenollp (approved):

PR #923, opened 2024-03-12
openapi3filter: deepObject array of objects and array of arrays support

danicc097:


fenollp (changes requested):
LGTM just one request

danicc097:
I don't see any changes? @fenollp

openapi3filter/req_resp_decoder.go
```diff
@@ -896,127 +895,193 @@ func deepSet(m map[string]interface{}, keys []string, value interface{}) {
...
+	for k := range m {
+		key, err := strconv.Atoi(k)
+		if err != nil {
+			return nil, fmt.Errorf("array indexes must be integers: %w", err)
 		}
+		keys = append(keys, key)
 	}
 
-	return obj, nil
+	for i := 0; i <= slices.Max(keys); i++ {
```
fenollp, line 953: Ah sorry. I probably misstepped and lost the message I was leaving here.

LGTM, just: could you inline this call here? So as to drop the dependency on `x/exp`

fenollp (approved):

PR #944, opened 2024-04-19
fix yaml marshaling of AdditionalProperties.Schema

safareli:
note that this issue doesn't exists for json marshaling.


fenollp:
Can you kindly add a test please?

fenollp (changes requested):
Needs a regression test plz

PR #945, opened 2024-04-22
openapi3: introduce `ReferencesComponentInRootDocument(doc *T, ref componentRef) (string, bool)`

percivalalb:
This MR adds a function (`ReferencesComponentInRootDocument`) to check if a component reference references the same document or element as another component reference in the root document's `#/components/<type>`. If it does, it returns the name of it in the form `#/components/<type>/NameXXX`.

It does this by keeping track of the source file paths of schemas in multi-file'd specs.

This enables the amalgamation of schemas which reference the same underlying model which removes *very annoying* to use anonymous structs in generated models when using codegen tools like oapi-codegen.

TODO: ~~add MR in oapi-codegen displaying improved model generation.~~ https://github.com/deepmap/oapi-codegen/pull/1573


percivalalb:
> Hello! 
> 
> How about a func that, given a `doc` and a `Schema`, returns a `*SchemaRef` (a ptr under `doc`) or `nil`? Would this not cover your use case?
> 
> 
> 
> Note: the `refs.go` file is generated.

Yes, if it also returned the name in the #components/schemas/...

Input: doc, Schema
Output: name, SchemaRef

I'll have a look at moving it out of the gen'ed file

percivalalb:
I've moved the alternations out of the code gen'ed files, sorry I'm still getting familar with the project setup and just thrown everything in the helper file. I'd like to ad some unit tests of couse but hopefully this is more to your liking.

fenollp:
This is looking pretty good!
Could you re-push the docs? It fails CI.
Also please send your tests away :) And an `Example` that shows usage of the function.

openapi3/helpers.go
```diff
@@ -39,3 +42,140 @@ func Int64Ptr(value int64) *int64 {
...
+}
+
+// MatchesComponentInRootDocument returns if the given component is identical
+// to a component defined in the root document's '#/components/<type>'.
+// It returns a reference to the schema in the form
+// '#/components/<type>/NameXXX'
+//
+// Of course given it a component from the root document will always match.
+//
+// https://swagger.io/docs/specification/using-ref/#syntax
```
fenollp, line 101: Also replace your links to swagger.io with ones to the actual spec at https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md

percivalalb:
Things to do:
- I think these snippets https://github.com/getkin/kin-openapi/pull/945/files#diff-035b395ae81006bbe6cd7144c686238113762fccdb051ddaef5d507560039ee6R540 which construct the refPath should be moved to a single function in `resolveComponent`. I believe componentPath should be the value we need already.
- I'm not sure how this plays well with schema's loaded from memory (i.e. no paths)
- Also not sure if it supports external referencing other spec files or URL references
- Need to add unit tests
- Add example usage of function

percivalalb:
Address most of the bullet points. WHere is the best place for the example?

openapi3/helpers.go
```diff
@@ -39,3 +42,219 @@ func Int64Ptr(value int64) *int64 {
...
+	}
+
+	// If there are no schemas defined in the root document return early.
+	if doc.Components == nil {
+		return "", false
+	}
+
+	var components map[string]componentRef
+
+	switch ref.CollectionName() {
```
percivalalb, line 169: Could I use https://pkg.go.dev/github.com/go-openapi/jsonpointer#GetForToken here?

percivalalb: Given this a go in https://github.com/getkin/kin-openapi/pull/945/commits/da3f155ef53e845b6ca750654b88c9c116d25b85

Save on having another JSON name to component mapping but uses reflection which isn't ideal. Works anyway

fenollp (approved):

PR #950, opened 2024-05-23
openapi{2,3}: surface both json/yaml unmarshal errors

timothympace:
If an error is encountered when attempting to unmarshal an openapi document formatted as JSON, callers will only see the error message from the YAML unmarshal. This adds an error message that combines both.

### DIFF
--- a/openapi2/marsh.go
+++ b/openapi2/marsh.go
@@ -17,10 +17,18 @@ func unmarshalError(jsonUnmarshalErr error) error {
 }
 
 func unmarshal(data []byte, v interface{}) error {
+	var jsonErr, yamlErr error
+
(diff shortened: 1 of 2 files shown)


fenollp:
Do you have a use case where both errors are required?

Or would it be fine to simply return the json error instead of the yml one?

timothympace:
If json unmarshal fails (due to invalid openapi schema), the current code will try to unmarshal as yaml. Most of the times this is ok because `yaml.Unmarshal` will surface the same schema issue, but it can also return other errors beforehand like [this one](https://github.com/invopop/yaml/blob/main/yaml.go#L52). In that case, it's very unclear what the actual issue was. This PR may not be the right fix for that use case though.

fenollp:
Thank you!

Issue #472, opened 2022-01-06
Consider adding context to Validator Middleware's ErrFunc and LogFunc functions


nsp:
This would be a breaking change, but for us, a good thing if these openapi3filter/middleware.go's types were instead

```
// ErrFunc handles errors that may occur during validation.
type ErrFunc func(ctx context.Context, w http.ResponseWriter, status int, code ErrCode, err error)

// LogFunc handles log messages that may occur during validation.
type LogFunc func(ctx context.Context, message string, err error)
```

And invoked by the middleware with `r.Context()`. This way, the logging and error handling functions can access important info from the request context, like tracingIDs, requestIDs, deadline information, etc.

It might not be a breaking change if additional handler functions are added (`ErrWithContextFunc` and `LogWithContextFunc`?) for additional complexity.

If one of these approaches is acceptable, I can contribute a MR.

crissi98:
Hi @fenollp,

I wanted to use the LogFunc as described in this issue, so i went ahead with a first proposition, but without breaking changes

PR #953, opened 2024-06-01
openapi3filter: add context to Validator Middleware's ErrFunc and LogFunc functions

crissi98:
This feature modifies the ErrFunc and LogFunc of the validator, just as described in #472 
In contrast to the change described in the issue, this is not a breaking changes, since it adds additional functions to set the LogFunc and the ErrFunc instead of overwriting the existing ones.

### DIFF
--- a/.github/docs/openapi3filter.txt
+++ b/.github/docs/openapi3filter.txt
@@ -182,7 +182,7 @@ type ErrCode int
     occur during validation. These may be used to write an appropriate response
     in ErrFunc.
 
-type ErrFunc func(w http.ResponseWriter, status int, code ErrCode, err error)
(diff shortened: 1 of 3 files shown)


fenollp:
Hi! Thanks for taking interest in this!

Really, let's perpetrate the tiny API breakage this patch entails, please. As a solo maintainer I'm always happy to keep the API surface low.

Also be sure to push the diffs from `.github/docs` once you run `./docs.sh`. It'll help with CI.

crissi98:
Thanks for your feedback!

I implemented the changes as you suggested by updating the existing functions instead of creating new ones.

Also, the `.github/docs` are fixed.

Issue #952, opened 2024-05-29
`internalizeRefs` behaviour may not be consistent across multiple calls


jamietanna:
Hey folks,

I just wondered if you were aware of / had any insights into something we've had reported in oapi-codegen:

- https://github.com/deepmap/oapi-codegen/issues/1572
- https://github.com/deepmap/oapi-codegen/issues/1623

This leads to cases where we have two different resulting JSON representations of the specification (i.e. https://github.com/deepmap/oapi-codegen/issues/1572#issuecomment-2093692468) which in our case can cause noise for folks, but may be an actual bug we want to look at here.

This only appears to have been reported with oapi-codegen v2.1.0 onwards, in which we bumped from `v0.118.0` to `v0.120.0`.

jamietanna:
Changes between the two releases: https://github.com/getkin/kin-openapi/compare/v0.118.0...v0.120.0

percivalalb:
The minimal-spec in https://github.com/percivalalb/kin-openapi/commit/04d5174c70427898aad888140ecb25fc244e215b shows the issue:

- Firstly note there is only one schema component `record` which both of the endpoint reference even though they have a different schema (one has the property `tracks`, the other `pages`). At a minium it would be good to warn when names collide as it's generated an incorrect schema which will mostly cause issues at a later point.
- Secondly when running the unit tests the schema which wins out and is included as the record component changes `record`. Could this logic be made deterministic.
```
Running tool: /usr/local/go/bin/go test -timeout 30s -run ^TestInternalizeRefs$ github.com/getkin/kin-openapi/openapi3 -v

=== RUN   TestInternalizeRefs
=== RUN   TestInternalizeRefs/testdata/testref.openapi.yml
=== RUN   TestInternalizeRefs/testdata/recursiveRef/openapi.yml
=== RUN   TestInternalizeRefs/testdata/spec.yaml
=== RUN   TestInternalizeRefs/testdata/callbacks.yml
=== RUN   TestInternalizeRefs/testdata/issue831/testref.internalizepath.openapi.yml
=== RUN   TestInternalizeRefs/testdata/interalizationNameCollision/api.yml
--- PASS: TestInternalizeRefs (0.04s)
    --- PASS: TestInternalizeRefs/testdata/testref.openapi.yml (0.01s)
    --- PASS: TestInternalizeRefs/testdata/recursiveRef/openapi.yml (0.01s)
[...]

PR #955, opened 2024-06-02
openapi3: improve internalization ref naming to avoid collisions

percivalalb:
#952

Internalizating references is the way in which a spec which references external files can be shipped as one combined document. The current implementation in [DefaultRefNameResolver](https://github.com/getkin/kin-openapi/blob/42a2d80973ff0d2b1254f327e1ddac9d6cff623c/openapi3/internalize_refs.go#L17) uses just the final part of the `$ref`. That either being the name of the fixed field component or file name of a whole document.

This works in most specs if they maintain documents in the same directory but as soon as you bring child directories the chance of a comflicting file name increasing meaning that two distinct schemas might resolve to the same internalized name. This then goes unnoticed and you unwittingly ship an incorrect spec. Also due to some indeterminisitc logic internalisting the references either of the distinct schemas can be internalized first which leads to an inconistent incorrect spec. (I've refered to schema's here but this appies to all the component fixed field types).

This MR:
- Keeps track of the document paths relative to the root document.
- These paths are then used to generate the internalized name using the entire path to avoid collisions.
- Makes the parsing and interlization deterministic by iterating over maps (which have no defined order) in sorted string key order.
- The `DefaultRefNameResolver` has been switched to the new logic, which therefore includes an API breaking change & internalised spec changes.

Like before the interlization code doesn't warn if you provide a "name internalizer" that produces colision - this could be added in a follow-up PR. Work would need think out about that could be done.

### DIFF
--- a/openapi3/internalize_refs.go
+++ b/openapi3/internalize_refs.go
@@ -3,7 +3,6 @@ package openapi3
 import (
 	"context"
 	"path"
-	"path/filepath"
 	"strings"
 )
 
@@ -62,13 +61,13 @@ func DefaultRefNameResolver(doc *T, ref componentRef) string {
 		}
 
 		// Remove the path extentions to make this JSON/YAML agnostic.
-		for ext := filepath.Ext(filePath); len(ext) > 0; ext = filepath.Ext(filePath) {
+		for ext := path.Ext(filePath); len(ext) > 0; ext = path.Ext(filePath) {
 			filePath = strings.TrimSuffix(filePath, ext)
 		}
 
 		// Trim the common prefix with the root doc path.
 		if doc.url != nil {
-			commonDir := filepath.Dir(doc.url.Path)
+			commonDir := path.Dir(doc.url.Path)
 			for {
 				if commonDir == "." { // no common prefix
 					break
@@ -79,7 +78,7 @@ func DefaultRefNameResolver(doc *T, ref componentRef) string {
 					break
 				}
 
-				commonDir = filepath.Dir(commonDir)
+				commonDir = path.Dir(commonDir)
 			}
 		}
 	}
@@ -88,7 +87,7 @@ func DefaultRefNameResolver(doc *T, ref componentRef) string {
 
 	// Trim .'s & slashes from start e.g. otherwise ./doc.yaml would end up as __doc
 	if filePath != "" {
-		internalisedName = strings.TrimLeft(filePath, "./"+string(filepath.Separator))
+		internalisedName = strings.TrimLeft(filePath, "./")
 	}
 
 	if componentPath != "" {
@@ -96,7 +95,7 @@ func DefaultRefNameResolver(doc *T, ref componentRef) string {
 			internalisedName += "_"
 		}
 
-		internalisedName += strings.TrimLeft(componentPath, "./"+string(filepath.Separator))
+		internalisedName += strings.TrimLeft(componentPath, "./")
 	}
 
 	// Replace invalid characters in component fixed field names.
@@ -112,19 +111,19 @@ func cutDirectories(p, dirs string) (string, bool) {
 		return p, false
 	}
 
-	p = strings.TrimRight(p, "/"+string(filepath.Separator))
-	dirs = strings.TrimRight(dirs, "/"+string(filepath.Separator))
+	p = strings.TrimRight(p, "/")
+	dirs = strings.TrimRight(dirs, "/")
 
 	var sb strings.Builder
 	sb.Grow(len(ParameterInHeader))
-	for _, segments := range filepath.SplitList(p) {
+	for _, segments := range strings.Split(p, "/") {
 		sb.WriteString(segments)
 
 		if sb.String() == p {
 			return strings.TrimPrefix(p, dirs), true
 		}
 
-		sb.WriteRune(filepath.Separator)
+		sb.WriteRune('/')
 	}
 
 	return p, false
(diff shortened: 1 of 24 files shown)


openapi3/internalize_refs.go
```diff
@@ -29,6 +33,70 @@ func DefaultRefNameResolver(ref string) string {
...
+	if componentPath != "" {
+		componentPath = strings.TrimLeft(componentPath, "/")
+		if refPath != "" {
+			refPath += "_"
+		}
+		refPath += componentPath
+	}
+
+	// XXX(alb): These should really be escaped to ~1
+	// But kin-openapi doesn't support refs with ~ at the moment.
```
fenollp, line 95: Hm in what way? It's possible I missed some corners

openapi3/internalize_refs.go
```diff
@@ -2,18 +2,22 @@ package openapi3
...
+type RefNameResolver func(*T, componentRef) string
 
 // DefaultRefResolver is a default implementation of refNameResolver for the
 // InternalizeRefs function.
 //
 // If a reference points to an element inside a document, it returns the last
 // element in the reference using filepath.Base. Otherwise if the reference points
 // to a file, it returns the file name trimmed of all extensions.
+//
+// Deprecated: Use UpdatedDefaultRefNameResolver which is less likely to generate
```
fenollp, line 19: No, let's go with the API breakage. Just add an entry to the changelog section of the README

openapi3/internalize_refs.go
percivalalb: If I didn't replace slashes with https://github.com/getkin/kin-openapi/pull/955/files#diff-ae5414b666f1cdb100809e5937dc1a92e1b0010861ec51943ec70faad26019eeR97 I'd get invalid character error and if I encoded the slash to `~1` I'd also get an invalid character error.

I'll expand on this and make another issue if there is one.

openapi3/internalize_refs.go
percivalalb: Sure,

openapi3/internalize_refs.go
percivalalb: With either
```
	return strings.ReplaceAll(refPath, "/", "~1") # encode / to ~1
	return refPath # leave slashes as they are
```
I get:
```
 	invalid components: schema "components.openapi.yml/CustomTestSchema": identifier "components.openapi.yml/CustomTestSchema" is not supported by OpenAPIv3 standard (regexp: "^[a-zA-Z0-9._-]+$")
```

ccoVeille:
Great changes!

One small remark: unless if the project is totally non-US based (so using colour, behaviour, and company), and that's commonly and widely accepted, I would have expected the whole code, issues and PR to mention : `internalize` (and all variations) instead of `internalise`

Please note, I'm French. It's the kind of error I used to do, so I spotted it

percivalalb:
> One small remark: unless if the project is totally non-US based (so using colour, behaviour, and company), and that's commonly and widely accepted, I would have expected the whole code, issues and PR to mention : `internalize` (and all variations) instead of `internalise`

Thanks for pointing out, I had it in the back of my mind but wasn't actively making sure I used the American spelling. I'm English so do this automatically too :D

openapi3/internalize_refs.go
percivalalb: Done

percivalalb: https://github.com/getkin/kin-openapi/pull/955/commits/23f97c6de03b179de8df6cc55aaa51681d172758

openapi3/internalize_refs.go
percivalalb: `/` or `~` are not valid characters in fixed field names which is why I was running into issues in both cases: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#components-object

Opened this issue only to release this shortly after: https://github.com/getkin/kin-openapi/issues/963#issuecomment-2169387253

Therefore swapping to a supported character like `_` might be the solution. This function at least needs to make sure the name it generates only has the following characters: `^[a-zA-Z0-9\.\-_]+$`

openapi3/refs.tmpl
```diff
@@ -40,6 +40,14 @@ func (x *{{ $type.Name }}Ref) CollectionName() string { return "{{ $type.Collect
 // RefPath returns the path of the $ref relative to the root document.
 func (x *{{ $type.Name }}Ref) RefPath() *url.URL { return &x.refPath }
 
+func (x *{{ $type.Name }}Ref) setRefPath(u *url.URL) {
+	if x.refPath != (url.URL{}) || u == nil {
```
fenollp, line 44: nitpick: let's switch the sides of `||` so the simpler check probably gets called first :)

percivalalb: Sure. I'm not actually a fan of this logic anyway but haven't had the time to invest in making it better.

percivalalb:
I think this is very close now, if you'd like to do a full review \@fenollp then please go ahead (you may have already as I've been trundling along). I'll do another pass through later this evening too.

openapi3/internalize_refs.go
percivalalb: Just to say I did implement that solution in this PR - any non-supported character is replaced by a `_`.

fenollp (changes requested):
LGTM! Please rebase and use `v0.126.0` as the release version in README and we'll be good to go :)

fenollp (approved):

PR #966, opened 2024-06-15
Replace interface{} with any

percivalalb:
Go 1.18 introduced any which is prefered over interface{}, they are equivalent (type alias).

https://github.com/golang/go/commit/2580d0e08d5e9f979b943758d3c49877fb2324cb

The go.mod currently targets Go 1.20+

### DIFF
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -1184,7 +1184,7 @@ func parsePrimitiveCase(raw string, schema *openapi3.SchemaRef, typ string) (any
 type EncodingFn func(partName string) *openapi3.Encoding
 
 // BodyDecoder is an interface to decode a body of a request or response.
-// An implementation must return a value that is a primitive, []interface{}, or map[string]interface{}.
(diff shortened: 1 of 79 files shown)


openapi3filter/req_resp_decoder.go
```diff
@@ -991,7 +991,7 @@ func buildResObj(params map[string]interface{}, parentKeys []string, key string,
 		if err != nil {
 			return nil, &ParseError{path: pathFromKeys(mapKeys), Kind: KindInvalidFormat, Reason: fmt.Sprintf("could not convert value map to array: %v", err)}
 		}
-		resultArr := make([]interface{} /*not 0,*/, len(arr))
+		resultArr := make([]any, len(arr))
```
fenollp, line 994:
```suggestion
		resultArr := make([]any /*not 0,*/, len(arr))
```

fenollp (changes requested):
Thanks for this (as well!) I only have 2 requests and this needs rebasing. LGTM!

fenollp (approved):

PR #970, opened 2024-06-25
openapi3: implement circular reference backtracking

AnatolyRugalev:
This PR adds a test scenario that leads to failure of resolving recursive references.

After troubleshooting for a while, I realized that current reference tracking algorithm in not perfect and will always have some edge cases, so I decided to try fixing that for good.

During reference resolution, we keep track of all previous references, and if we find out that the reference is already being resolved, we register a backtracking callback that will assign a pointer to the resolved schema once it's done.

This guarantees that schema will always be resolved regardless of the depth.

### DIFF
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -912,12 +912,12 @@ func (loader *Loader) resolveSecuritySchemeRef(doc *T, component *SecurityScheme
 			component.Value = &scheme
 			component.refPath = *documentPath
 		} else {
-			var resolved SecuritySchemeRef
 			if !loader.shouldVisitRef(ref, func(value any) {
 				component.Value = value.(*SecurityScheme)
 			}) {
 				return nil
 			}
+			var resolved SecuritySchemeRef
 			loader.visitRef(ref)
 			doc, componentPath, err := loader.resolveComponent(doc, ref, documentPath, &resolved)
 			defer loader.unvisitRef(ref, resolved.Value)
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -16,7 +16,14 @@ import (
 	"strings"
 )
 
+// CircularReferenceError is deprecated.
+// kin-openapi will never throw this error anymore, and it's kept for compatibility reasons
+// Deprecated: CircularReferenceError is deprecated.
 var CircularReferenceError = "kin-openapi bug found: circular schema reference not handled"
+
+// CircularReferenceCounter is deprecated.
+// kin-openapi does not use this counter anymore, and it's kept for compatibility reasons
+// Deprecated: CircularReferenceCounter is deprecated.
 var CircularReferenceCounter = 3
 
 func foundUnresolvedRef(ref string) error {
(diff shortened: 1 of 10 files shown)


AnatolyRugalev:
@fenollp I updated the PR description and its contents :)

This should solve circular references resolution for good. Let me know if you have any questions or concerns

openapi3/loader.go
```diff
@@ -16,7 +16,14 @@ import (
...
 
+// CircularReferenceError is deprecated.
+// kin-openapi will never throw this error anymore, and it's kept for compatibility reasons
+// Deprecated: CircularReferenceError is deprecated.
 var CircularReferenceError = "kin-openapi bug found: circular schema reference not handled"
+
+// CircularReferenceCounter is deprecated.
+// kin-openapi does not use this counter anymore, and it's kept for compatibility reasons
+// Deprecated: CircularReferenceCounter is deprecated.
 var CircularReferenceCounter = 3
```
fenollp, line 27: Actually just drop these. We're `v0`, meaning we advertise as potentially breaking on next release.
Simply just add a mention on the changelog section of the README, under a new `v0.126.0` section.

openapi3/loader.go
```diff
@@ -299,6 +300,35 @@ func isSingleRefElement(ref string) bool {
...
+		}
+	}
+	delete(loader.visitedRefs, ref)
+	delete(loader.backtrack, ref)
+	loader.visitedPath = loader.visitedPath[:len(loader.visitedPath)-1]
+}
+
+func (loader *Loader) shouldVisitRef(ref string, fn func(value any)) bool {
+	_, ok := loader.visitedRefs[ref]
+	if ok {
```
fenollp, line 325:
```suggestion
	if _, ok := loader.visitedRefs[ref]; ok {
```

openapi3/loader.go
```diff
@@ -494,6 +524,8 @@ func drillIntoField(cursor any, fieldName string) (any, error) {
 	}
 }
 
+var errAlreadyResolved = errors.New("already resolved")
```
fenollp, line 527: dead code: drop

openapi3/loader.go
```diff
@@ -817,14 +842,15 @@ func (loader *Loader) resolveSchemaRef(doc *T, component *SchemaRef, documentPat
...
-				return fmt.Errorf("%s with length %d - %s", CircularReferenceError, len(visited), strings.Join(visited, " -> "))
+			if !loader.shouldVisitRef(ref, func(value any) {
+				component.Value = value.(*Schema)
+			}) {
+				return nil
 			}
-			visited = append(visited, ref)
-
+			loader.visitRef(ref)
 			var resolved SchemaRef
```
fenollp, line 851:
```suggestion
			var resolved SchemaRef
			loader.visitRef(ref)
```

openapi3/loader.go
```diff
@@ -911,7 +926,14 @@ func (loader *Loader) resolveSecuritySchemeRef(doc *T, component *SecurityScheme
 			component.refPath = *documentPath
 		} else {
 			var resolved SecuritySchemeRef
+			if !loader.shouldVisitRef(ref, func(value any) {
+				component.Value = value.(*SecurityScheme)
+			}) {
+				return nil
+			}
```
fenollp, line 933:
```suggestion
			if !loader.shouldVisitRef(ref, func(value any) {
				component.Value = value.(*SecurityScheme)
			}) {
				return nil
			}
			var resolved SecuritySchemeRef
```

fenollp (changes requested):
Hi & thanks for taking the time to look into this!

Mostly LGTM, just: I see you're not using the visitref utils you're introducing in these two places and it feels like this was fogotten?
* `func (loader *Loader) resolveLinkRef(doc *T, component *LinkRef, documentPath *url.URL) (err error)`
* `func (loader *Loader) resolvePathItemRef(doc *T, pathItem *PathItem, documentPath *url.URL) (err error)`
Could you comment on or fix that please?

fenollp (approved):
LGTM!

PR #973, opened 2024-07-01
replicate circular refs issue

reuvenharrison:
This PR contains a unit test that demonstrates the circular refs problem that was reported [here](https://github.com/Tufin/oasdiff/issues/442).
Unfortunately https://github.com/getkin/kin-openapi/pull/970 doesn't seem to resolve this.

### DIFF
--- a/openapi3/testdata/circularRef2/AwsEnvironmentSettings.yaml
+++ b/openapi3/testdata/circularRef2/AwsEnvironmentSettings.yaml
@@ -1,0 +1,7 @@
+type: object
+properties:
+    children:
+        type: array
+        items:
+            $ref: './AWSEnvironmentSettings.yaml'
+description: test
--- a/openapi3/loader_circular_test.go
+++ b/openapi3/loader_circular_test.go
@@ -1,0 +1,14 @@
+package openapi3
+
+import (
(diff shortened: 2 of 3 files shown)


openapi3/testdata/circularRef2/AwsEnvironmentSettings.yaml
```diff
@@ -0,0 +1,7 @@
+type: object
+properties:
+    children:
+        type: array
+        items:
+            $ref: './AWSEnvironmentSettings.yaml'
```
fenollp, line 6: Note: replacing this with `'#/'` passes tests

AnatolyRugalev:
This PR https://github.com/getkin/kin-openapi/pull/974 handles this issue.

BTW, `./AwsEnvironmentSettings.yaml` and `./AWSEnviromentSettings.yaml` are technically to different references, so this will produce two identical types when resolving them on MacOS and Windows due to case insensitive file systems. You can actually see that the test failure in this PR complaints about missing file.

In my PR I fixed this inconsistency

fenollp:
Addressed in https://github.com/getkin/kin-openapi/pull/974

PR #976, opened 2024-07-03
openapi3: add support for number and integer format validators

### LINKED ISSUE #107: Add DefineNumberFormat function
Hey, 
We use a custom format with type "number" e.g.:
"eo:epsg": {
    "type": "number",
    "format": "epsg-code",
    ...
 }

There is the function DefineStringFormat implemented into openapi3 (related issues #103, #79), but it only handles elements with type "string". So there is a need for a DefineNumberFormat function.

AnatolyRugalev:
This PR adds support for Number and Integer format validators.

Fixes #107


fenollp (approved):

fenollp:
Thanks for your great work!

PR #978, opened 2024-07-07
openapi3: include local reference parts in refPath saved

percivalalb:
A bug fix for code added in https://github.com/getkin/kin-openapi/pull/955 for a seldom run code path with recursive refs

* The local part was not included when visiting components during recursion
* Also handles the case when base path is null when loading files data in memory
* Copy uri.URL elements to avoid manipulation from outside source

### DIFF
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -288,6 +288,31 @@ func resolvePathWithRef(ref string, rootPath *url.URL) (*url.URL, error) {
 	return resolvedPath, nil
 }
 
+func (loader *Loader) resolveRefPath(ref string, path *url.URL) (*url.URL, error) {
+	if ref != "" && ref[0] == '#' {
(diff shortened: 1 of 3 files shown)


fenollp (approved):
LGTM. Any regression tests to include before we merge this?

percivalalb:
> LGTM. Any regression tests to include before we merge this?

I thought you might ask :P, I will look at adding at least one direct test of the refPath of component references. There are a good amount of indirect tests (like internalise reference naming) but not for this particular case. I want to do that later today, we'll see if that happens

percivalalb:
Struggling to come up with test case, ideally we'd like to walk through the referenced components in a document. That almost involves re-implementing the resolving in loader.go. I think there are sufficent test cases for most the logic, up to you if you merge prior to additional tests for this specific case. I'll have to come back to this another time

Issue #968, opened 2024-06-17
openapi3gen does not generate `nullable: true` for pointers


josephbuchma:
This should include `"nullable": true`
https://github.com/getkin/kin-openapi/blob/master/openapi3gen/openapi3gen_test.go#L115-L117

This issue is not possible to fix using `openapi3gen.SchemaCustomizer` because `reflect.Type` passed into `openapi3.SchemaCustomizerFn` is never a pointer.

endertunc:
You can still work around it using SchemaCustomizer and custom StructTag.

```
type StructWithPointer struct {
	A *string `json:"a" nullable`
}
	
customizer := openapi3gen.SchemaCustomizer(func(name string, ft reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error {
	if _, ok := tag.Lookup("nullable"); ok {
		schema.Nullable = true
	}
	return nil
})
```

For the solution to the problem, this is where we _lose_ the "pointer" information https://github.com/getkin/kin-openapi/blob/master/openapi3gen/openapi3gen.go#L204-L206

I think something like following should solve the problem:
```
isNullable := false
for t.Kind() == reflect.Ptr {
	t = t.Elem()
	isNullable = true
}

...

schema := &openapi3.Schema{} // this already exists
schema.Nullable = isPointer // new change
```

I did run the tests you pointed and it seems to work. I am not sure if there are edge cases for but I am happy to raise a PR if this is acceptable.

On the other hand, openapi3gen seems to be trying to be minimalisti. For example, you also don't get `required` in anyone your `required` fields. I think that's they idea with `SchemaCustomizer` which allow you to customize the schema the way you want so the workaround I mentioned might be the suggest way to solve these kind of issues :)

fenollp:
@endertunc your patch proposal seems very reasonable. Please open a PR :)

PR #987, opened 2024-07-19
openapi3gen: generate "nullable: true" for pointers

endertunc:
Fixes https://github.com/getkin/kin-openapi/issues/968

### DIFF
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -195,14 +195,16 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 			return nil, &CycleError{}
 		}
 	}
-
-	if cap(parents) == 0 {
+	isRoot := cap(parents) == 0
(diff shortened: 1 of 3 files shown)


openapi3gen/openapi3gen.go
```diff
@@ -195,14 +195,16 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
...
+	isRoot := cap(parents) == 0
+	if isRoot {
 		parents = make([]*theTypeInfo, 0, 4)
 	}
 	parents = append(parents, typeInfo)
 
+	isNullable := false
 	for t.Kind() == reflect.Ptr {
 		t = t.Elem()
+		isNullable = !isRoot
```
endertunc, line 207: I assume we don't want to mark root structs `nullable` so I we have this extra check here.

endertunc:
Hi @fenollp, this should be ready for review. Let me know if you have any feedback!

PR #1002, opened 2024-08-12
Add openapi3.Schema.OrderedPropertyKeys

diamondburned:
This commit adds the `OrderedPropertyKeys` method to the
`openapi3.Schema`:

    OrderedPropertyKeys returns the keys of the properties in the order
    they were defined. This is useful for generating code that needs to
    iterate over the properties in a consistent order. If the keys could
    not be extracted for some reason, then this method automatically
    sorts the keys to be deterministic.

This is done via a temporary fork of the YAML-to-JSON transformation library.
It will not be ready until these PRs are merged in this order:

    - https://github.com/silasdavis/yaml/pull/1
    - https://github.com/ghodss/yaml/pull/62

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,10 +2,14 @@ module github.com/getkin/kin-openapi
 
 go 1.14
 
+replace github.com/ghodss/yaml/v2 => github.com/diamondburned/yaml/v2 v2.0.0-20240812065612-baf990d70122
+
 require (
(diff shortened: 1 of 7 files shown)


diamondburned:
Replaced by https://github.com/getkin/kin-openapi/pull/1003.

PR #1003, opened 2024-08-12
Add openapi3.Schema.OrderedPropertyKeys

diamondburned:
This commit adds the `OrderedPropertyKeys` method to the
`openapi3.Schema`:

    OrderedPropertyKeys returns the keys of the properties in the order
    they were defined. This is useful for generating code that needs to
    iterate over the properties in a consistent order. If the keys could
    not be extracted for some reason, then this method automatically
    sorts the keys to be deterministic.

This is done via a temporary fork of the YAML-to-JSON transformation library.
It will not be ready until invopop/yaml#13 is merged.


lzap:
I like this, the random order is driving me crazy.

No response so far in the `yaml` PR, I pinged them but this is already a second issue/pr without any response. It looks like to me that those folks were only interested in patching the CVE with no further development. Let's see if we get more information.

PR #1008, opened 2024-08-29
make form required field order deterministic

jlsherrill:
The required fields list of a form was generated by iterating over a map, but in Go maps do not seem to have a natural order.  This causes the list of fields to have some randomness to it.  

Everything else so far is deterministic, and we rely on this to determine if the openapi spec needs to be updated

### DIFF
--- a/openapi2conv/issue1008_test.go
+++ b/openapi2conv/issue1008_test.go
@@ -1,0 +1,49 @@
+package openapi2conv
+
+import (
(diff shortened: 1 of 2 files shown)


fenollp:
LGTM but please add a (regression) test

jlsherrill:
added a test that should reliably fail on master but passes with this.  Thanks!

Issue #1010, opened 2024-09-17
OpenAPI 2.0: Unable to Unmarshal Spec with Discriminator Field Set


reversearrow:
OpenAPI 2.0 supports the [discriminator field](https://swagger.io/specification/v2/#schema-object), but when the discriminator is included in an OpenAPI 2.0 document, unmarshaling the document fails. 

OpenAPI 2.0 the discriminator field is set using `string`, whereas in OpenAPI 3.0 and later, the discriminator is represented as an object.

The unmarshaling process fails because the [schema struct](https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L132) used for parsing expects the discriminator field to be an object, which is valid in OpenAPI 3.0 but not in OpenAPI 2.0.

This [schema struct](https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L132) is shared between the OpenAPI v2.0 and v3.0+. 

The solution is to update the type of the discriminator to interface{} (or any) and cast it based on the OpenAPI version, ensuring compatibility between versions.

**Example Payload with `discriminator` field set**
```
{
  "basePath": "/v2",
  "host": "test.example.com",
  "info": {
    "title": "MyAPI",
    "version": "0.1",
    "x-info": "info extension"
  },
  "paths": {
    "/foo": {
      "get": {
        "operationId": "getFoo",
        "responses": {
          "200": {
            "description": "returns all information",
            "schema": {
              "$ref": "#/definitions/Pet"
            }
          },
          "default": {
            "description": "OK"
          }
        },
        "summary": "get foo"
      }
    }
  },
  "schemes": [
    "http"
  ],
  "swagger": "2.0",
  "definitions": {
    "Pet": {
      "type": "object",
      "required": [
        "petType"
      ],
      "properties": {
        "petType": {
          "type": "string"
        },
        "name": {
          "type": "string"
        },
        "age": {
          "type": "integer"
        }
      },
      "discriminator": "petType"
    },
    "Dog": {
      "allOf": [
        {
          "$ref": "#/definitions/Pet"
        },
        {
          "type": "object",
          "properties": {
            "breed": {
              "type": "string"
            }
          }
        }
      ]
    },
    "Cat": {
      "allOf": [
        {
          "$ref": "#/definitions/Pet"
        },
        {
          "type": "object",
          "properties": {
            "color": {
              "type": "string"
            }
          }
        }
      ]
    }
  }
}
```

**Response**

```
[...]

PR #1011, opened 2024-09-17
openapi2: fix un/marshalling discriminator field

reversearrow:
This PR addresses issue https://github.com/getkin/kin-openapi/issues/1010. It allows `discriminator` field to be of any type and case it based on the OpenAPI version.

### DIFF
--- a/openapi2conv/openapi2_conv.go
+++ b/openapi2conv/openapi2_conv.go
@@ -272,7 +272,6 @@ func ToV3Parameter(components *openapi3.Components, parameter *openapi2.Paramete
 			MinLength:       parameter.MinLength,
 			MaxLength:       parameter.MaxLength,
 			Default:         parameter.Default,
-			Items:           parameter.Items,
 			MinItems:        parameter.MinItems,
 			MaxItems:        parameter.MaxItems,
 			Pattern:         parameter.Pattern,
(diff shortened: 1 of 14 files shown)


reversearrow:
@fenollp can please review this PR? Thanks

fenollp:
Hi! Thanks for tackling this. 

The ideal solution to me is to define a different Discriminator type in v2 and leave the current definition as is in v3. I. E. to have two different types and avoid casting. 
Do you think you'd be able to update your PR with this in mind?
Thanks

reversearrow:
> Hi! Thanks for tackling this.
> 
> The ideal solution to me is to define a different Discriminator type in v2 and leave the current definition as is in v3. I. E. to have two different types and avoid casting. Do you think you'd be able to update your PR with this in mind? Thanks

Hi @fenollp, thank you for your feedback. I agree with your suggestion. After briefly reviewing the proposed approach, it seemed like it would involve significant effort. In the meantime, I’ve updated the PR with a simpler approach than before, while not perfect, addresses the bug. Do you think we can proceed with merging this PR? If so, I can work on implementing the original suggested approach without the pressure of a strict timeline, as this is currently impacting one of our customers in production.

openapi2conv/openapi2_conv.go
```diff
@@ -789,19 +844,19 @@ func FromV3SchemaRef(schema *openapi3.SchemaRef, components *openapi3.Components
...
+				Type:         paramType,
+				Enum:         schema.Value.Enum,
+				Minimum:      schema.Value.Min,
+				Maximum:      schema.Value.Max,
+				ExclusiveMin: schema.Value.ExclusiveMin,
+				ExclusiveMax: schema.Value.ExclusiveMax,
+				MinLength:    schema.Value.MinLength,
+				MaxLength:    schema.Value.MaxLength,
+				Default:      schema.Value.Default,
+				// Items:           schema.Value.Items,
```
reversearrow, line 859: I have commented this out as Items is only required when type is "array". Since this is a binary/file format type. Array is not required. 

https://spec.openapis.org/oas/v3.0.3#schema-object

reversearrow:
@fenollp I have create a new Schema Object for the OpenAPI V2. Please review and let me know.

PR #1007, opened 2024-08-25
openapi3: reference originating locations in YAML specs - step 1

reuvenharrison:
This PR adds the ability to correlate OpenAPI components with the location in the originating YAML spec.
Here's a summary of the changes required to achieve this:
1. Modify [yaml.v3 ](https://github.com/go-yaml/yaml/tree/v3) to add line and column numbers of each YAML element to the unmarshaled interfaces (see https://github.com/oasdiff/yaml3)
2. Modify [invopop/yaml](https://github.com/invopop/yaml) with a new function: `UnmarshalWithOrigin` (see https://github.com/oasdiff/yaml)
3. Modify this repo to add an `Origin` struct to OpenAPI components. This requires changes to the UnmarshalJSON functions and, in some cases, replacing native types, like `map[string]string`, by dedicated types that include the original native type and an Origin member.

This PR addresses issue https://github.com/getkin/kin-openapi/issues/986

Follow up tasks to be addressed by future PRs:
- Add `origin` to more elements, namely `Components` and some types that are unmarshalled with `unmarshalStringMapP` or `unmarshalStringMap`
- Merge changes in yaml.v3 and invopop/yaml back to original repos and remove `replace`  directives from go.mod
- Enhance `origin` to include end line and column and YAML comments as proposed by @fenollp

PR #1018, opened 2024-10-14
openapi3: introduce StringMap type to enable unmarshalling of maps with Origin

reuvenharrison:
This PR replaces two fields in openapi3 from `map[string]string` by a new type `StringMap`:
- `Discriminator.Mapping`
- `OAuthFlow.Scopes`

This new `StringMap`type will enable unmarshalling of maps with origin in the subsequent PR: https://github.com/getkin/kin-openapi/pull/1007.


fenollp:
Please rebase on top of https://github.com/getkin/kin-openapi/pull/1019

Issue #1021, opened 2024-10-16
discuss: refs in discriminator object


jgresty:
I have noticed that while the upstream openapi 3 spec for discriminator object specifies the data type is `Map[string]string`, in their written examples they use refs: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#discriminator-object

While this library sticks to the literal spec type definition, we have noticed many real world specs which do accept refs in this field. Since even the upstream spec is not well defined here perhaps this library should be more lax?

Curious to hear opinions about this. I have also raised an issue upstream about this: https://github.com/OAI/OpenAPI-Specification/issues/4138

jgresty:
So from upstream the typing is correct as [reference objects](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#reference-object) also use type string for their referenced strings.

In this case I think the discriminator mapping object needs to be considered in functions such as `InternalizeRefs` and `Validate`

PR #1022, opened 2024-10-16
openapi3: process discriminator mapping values as refs

jgresty:
While the type of the discriminator mapping values is a string in the upstream specs, it contains a jsonschema reference to a schema object. It is surprising behaviour that these refs are not handled when calling functions such as InternalizeRefs.

This patch adds the data structures to store the ref internally, and updates the Loader and InternalizeRefs to handle this case. There may be several more functions that need to be updated that I am not aware of.

Since it is not a full Ref object we have to do some fudging to make it work with all the existing ref handling code.

This is a proposal for #1021 , I do not expect it to be the actual implementation.

### DIFF
--- a/openapi3/discriminator.go
+++ b/openapi3/discriminator.go
@@ -11,8 +11,8 @@ type Discriminator struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
 	Origin     *Origin        `json:"origin,omitempty" yaml:"origin,omitempty"`
 
-	PropertyName string    `json:"propertyName" yaml:"propertyName"` // required
-	Mapping      StringMap `json:"mapping,omitempty" yaml:"mapping,omitempty"`
+	PropertyName string            `json:"propertyName" yaml:"propertyName"` // required
+	Mapping      StringMap[string] `json:"mapping,omitempty" yaml:"mapping,omitempty"`
 }
 
 // MarshalJSON returns the JSON encoding of Discriminator.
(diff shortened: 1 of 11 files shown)


openapi3/discriminator.go
```diff
@@ -10,8 +10,22 @@ import (
 type Discriminator struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
 
-	PropertyName string    `json:"propertyName" yaml:"propertyName"` // required
-	Mapping      StringMap `json:"mapping,omitempty" yaml:"mapping,omitempty"`
+	PropertyName string                `json:"propertyName" yaml:"propertyName"` // required
+	Mapping      map[string]MappingRef `json:"mapping,omitempty" yaml:"mapping,omitempty"`
```
fenollp, line 14: Please rebase. Sorry about that, I noticed your issue and PR after merging the other one.

Also please keep the newtype `StringMap` functionality (see recent changes).

fenollp (changes requested):

openapi3/discriminator.go
jgresty: Sorry I did write this before the `StringMap` change came in and naïvely rebased, I saw this patch as a proposal for discussion but if you believe this is the right approach then I will fix it up

jgresty: To build upon the existing `StringMap` I have introduced a change to make it accept a generic value. It is only used in 2 places so the diff is tiny

fenollp:
Hi @jgresty 
I had to revert this change due to https://github.com/getkin/kin-openapi/issues/1028
Please feel free to give this another try when you have the time. Thanks!

PR #1050, opened 2025-01-28
feat(openapi3gen): Customize json.RawMessage

kyleconroy:
Using a SchemaCustomizer, I can't configure schemas for `json.RawMessage` fields. If I have the following struct

```go
type Foo struct {
    Bar string `json:"bar"`
    Output json.RawMessage `json:"output"`
}
```

The schema customization function will only be called for the Bar field, not the Output field.


kyleconroy:
@fenollp please let me know if this is something you'd consider merging. Thanks!

fenollp:
Please add tests.

kyleconroy:
Tests added

PR #1052, opened 2025-02-09
openapi3gen: Fix issue with separate component generated for time.Time

d1vbyz3r0:
openapi3gen generates separate schema component for time.Time properies, when ExportComponentSchemas opt provided.
Here is a small example:
```go
type Some struct {
	Name      string
	CreatedAt time.Time
}

func main() {
	schemas := make(openapi3.Schemas)
	g := openapi3gen.NewGenerator(
		openapi3gen.UseAllExportedFields(),
		openapi3gen.CreateComponentSchemas(openapi3gen.ExportComponentSchemasOptions{
			ExportComponentSchemas: true,
		}),
	)

	ref, err := g.NewSchemaRefForValue(&Some{}, schemas)
	if err != nil {
		panic(err)
	}

	schema, _ := json.MarshalIndent(ref, "", "  ")
	fmt.Println(string(schema))
}
```

Expected output is:
```json
{
  "properties": {
    "CreatedAt": {
      "format": "date-time",
      "type": "string"
    },
    "Name": {
      "type": "string"
    }
  },
  "type": "object"
}
```

But for now generator produces following:
```json
{
  "properties": {
    "CreatedAt": {
      "$ref": "#/components/schemas/Time"
    },
    "Name": {
      "type": "string"
    }
  },
  "type": "object"
}
```

### DIFF
--- a/openapi3gen/openapi3gen.go
+++ b/openapi3gen/openapi3gen.go
@@ -429,6 +429,11 @@ func (g *Generator) generateWithoutSaving(parents []*theTypeInfo, t reflect.Type
 
 	// For structs we add the schemas to the component schemas
 	if len(parents) > 1 || g.opts.exportComponentSchemas.ExportTopLevelSchema {
+		// If struct is a time.Time instance, separate component shouldn't be generated
+		if t == timeType {
+			return openapi3.NewSchemaRef(t.Name(), schema), nil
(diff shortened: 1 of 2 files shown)


fenollp:
Please add your example as a test and we should be good to go.

Issue #1053, opened 2025-03-06
Bug with explode=false and default on a query parameter array


nhochstr:
When a `query` parameter is defined with `explode: false` and a `default` value as an array, the generator applies the `default` as if `explode` were `true`, causing a validation failure.

The issue occurs with the following OpenAPI definition:

````
openapi: 3.0.0
info:
  title: 'Validator'
  version: 0.0.1
paths:
  /category:
    get:
      parameters:
        - $ref: "#/components/parameters/Type"
      responses:
        '200':
          description: Ok
components:
  parameters:
    Type:
      in: query
      name: type
      required: false
      explode: false
      description: Type parameter
      schema:
        type: array
        default:
          - A
          - B
          - C
        items:
          type: string
          enum:
            - A
            - B
            - C
````

PR #1054, opened 2025-03-06
openapi3filter: apply default values of an array in a query param with exploded = false

nhochstr:
This is a fix for https://github.com/getkin/kin-openapi/issues/1053

Fix to apply default values of an array in query parameters when exploded is set to false


openapi3filter/issue991_test.go
```diff
@@ -60,7 +60,7 @@ components:
 			name: "Valid request without type parameters set",
 			args: args{
 				url:      "/category",
-				expected: []string{"A", "B", "C"},
+				expected: []string{"A,B,C"},
```
fenollp, line 63: LGTM. Please duplicate this test case with `explode:true` as well.

fenollp (changes requested):

openapi3filter/issue991_test.go
nhochstr: I added a new test case with explode set to true

Issue #1051, opened 2025-02-03
Field named "origin" is not loaded


lwalen:
As of [v0.129.0](https://github.com/getkin/kin-openapi/releases/tag/v0.129.0), a field named `origin` is no longer loaded.

v0.128.0 example: https://go.dev/play/p/nWestYSO-CX

v0.129.0 example: https://go.dev/play/p/L1oax7Vwd-M

```
package main

import (
	"fmt"
	"log"

	"github.com/getkin/kin-openapi/openapi3"
)

var data = `
paths:
  /foo:
    get:
      responses:
        "200":
          description: OK
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Foo"
components:
  schemas:
    Foo:
      type: object
      properties:
        origin:
          type: string
`

func main() {
	loader := openapi3.NewLoader()
	doc, err := loader.LoadFromData([]byte(data))
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("%#v", doc.Paths.Find("/foo").Get.Responses.Status(200).Value.Content.Get("application/json").Schema.Value.Properties["origin"])
}
```

jonathansharman:
I suspect this is caused by https://github.com/getkin/kin-openapi/pull/1007 in particular, which added `Origin` fields to a bunch of schema types.

julienflyr:
This makes this package unusable for usage in any project related to travel where you have `origin` and `destination` fields in the API.

fenollp:
@reuvenharrison PTAL

reuvenharrison:
The problem is due to such lines which were added to the `UnmarshalJSON` functions:
`delete(x.Extensions, originKey)`

I intend to fix this by making this delete conditional on loader.IncludeOrigin.

PR #1055, opened 2025-03-07
openapi3: delete origin keys only when IncludeOrigin=true

reuvenharrison:
This PR fixes https://github.com/getkin/kin-openapi/issues/1051.
The problem was caused because openapi3 removed `origin` keys from `Schema.Properties` and other map-based types including all `Components` types.
To fix this I changed the behavior to only delete `origin` keys when `IncludeOrigin` is set to true.
Note that this required moving `Loader.IncludeOrigin` to the global scope so that map unmarshalling functions can access it.
For example: `Schemas.UnmarshalJSON` -> `unmarshalStringMapP` -> `popOrigin` which accesses `IncludeOrigin`

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -90,6 +90,11 @@ var ErrURINotSupported = errors.New("unsupported URI")
     ErrURINotSupported indicates the ReadFromURIFunc does not know how to handle
     a given URI.
 
+var IncludeOrigin = false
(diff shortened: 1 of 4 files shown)


fenollp:
LGTM but please rebase now that I merged https://github.com/getkin/kin-openapi/pull/1057
Also please rename the PR to something that makes standalone sense.

reuvenharrison:
@fenollp the PR is ready to merge

PR #1061, opened 2025-03-25
style: Use fmt.Sprint without formating

gaiaz-iusipov:
`fmt.Sprintf("%v", somthing)` can be safely simplified by `fmt.Sprint(something)`

### DIFF
--- a/openapi3filter/req_resp_encoder_test.go
+++ b/openapi3filter/req_resp_encoder_test.go
@@ -2,7 +2,6 @@ package openapi3filter
 
 import (
 	"fmt"
-	"net/http"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


openapi3filter/req_resp_encoder_test.go
```diff
@@ -14,15 +13,13 @@ func TestRegisterAndUnregisterBodyEncoder(t *testing.T) {
 	encoder = func(body any) (data []byte, err error) {
 		return []byte(strings.Join(body.([]string), ",")), nil
 	}
-	contentType := "text/csv"
-	h := make(http.Header)
-	h.Set(headerCT, contentType)
```
fenollp, line 19: Why drop this?

gaiaz-iusipov: The variable `h` is not used in the test and does not affect the result in any way

PR #1064, opened 2025-04-07
openapi3filter: don't consume request body in `AuthenticatorFunc`

### LINKED ISSUE #743: http: invalid Read on closed Body error comes in AuthenticationFunc with request body validation
I have a POST request with the request body parameter. I am using `OapiRequestValidatorWithOptions` to validate and authenticate each and every request. However, it gave `http: invalid Read on closed Body` error and fails to bind the JSON to the request object. 

I was able to avoid the error by setting ExcludeRequestBody so ValidateRequest skips request body validation as shown in the example below. 

```
validator := middleware.OapiRequestValidatorWithOptions(swagger,
		&middleware.Options{
			Options: openapi3filter.Options{
				ExcludeRequestBody: true,
				AuthenticationFunc: NewAuthenticator(middleware.AuthConfig{}),
			},
		})
```
However, I would like to continue to use ValidateRequest. 
Is there any solution to this?

jamietanna:
As noted in #743, some users of the `openapi3filter`, largely via
`oapi-codegen` and `openapi.tanna.dev/go/validator`[0] are seeing that
when using an `AuthenticationFunc`, this can lead to errors such as:

    request body has an error: value is required but missing

Or:

    request body has an error: reading failed: http: invalid Read on closed Body

This turns out to be down to the fact that a call to the
`AuthenticationFunc` may read the request body, but due to the way that
Go's `io.Reader`s can only be read once, this then means the body cannot
be read by anyone else further in the request chain.

To resolve this, we can make sure that each iteration of the
`AuthenticationFunc` gets a fresh copy of the request body (in its own
`io.Reader`).

This follows logic that exists within the `ValidateRequestBody` method,
with an additional layer of caching the raw bytes from the body.

We also add tests to validate:

- calling the `ValidateRequest` body function with an
  `AuthenticationFunc` that reads the request body
- using an integration test with a test HTTP server also works,
  indicating that the HTTP handler can then read the request body itself

Closes #743.

[0]: https://gitlab.com/jamietanna/httptest-openapi/-/issues/5

### DIFF
--- a/openapi3filter/issue743_test.go
+++ b/openapi3filter/issue743_test.go
@@ -1,0 +1,207 @@
+package openapi3filter_test
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers/gorillamux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestValidateRequestWithAnAuthenticatorFunc_CanConsumeTheRequestBody(t *testing.T) {
+	const spec = `
+openapi: 3.0.0
+info:
+  title: 'Validator'
+  version: 0.0.1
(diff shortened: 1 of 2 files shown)


fenollp:
Thanks!

PR #1066, opened 2025-04-28
openapi3: resolve Snyk security warning with path traversal

seborama:
I am using kin-openapi on a project and Snyk blocked me with an error:

> **Path Traversal**
> Unsanitized input from the request URL flows into os.ReadFile, where it is used as a path. This may result in a Path Traversal vulnerability and allow an attacker to read arbitrary files.

This PR proposes a remediation.

Thank for this really useful library!

### DIFF
--- a/openapi3/loader_uri_reader.go
+++ b/openapi3/loader_uri_reader.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


fenollp:
Does path.Clean resolve the Snyk warning?

seborama:
As far as I can tell, using this branch with a `replace` statement in the `go.mod`, seems to have put Snyk at peace.

Issue #817, opened 2023-06-28
property described with multipleOf does not pass the validation


jpaliwoda13:
In my openapi file, I have one property that is described as follows:
```
Quantity:
    type: number
    maximum: 999999999.99
    minimum: -999999999.99
    multipleOf: 0.01
```
The problem is that not every value provided as Quantity passes validation. 

This part of lib code seems to be responsible for random outcome: [kin-openapi/openapi3/schema.go#L1508](https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go#L1508) as the division value/0.01 sometimes gives a rounded value (so it’s Int and it passes validation), but sometimes it gives decimal outcome and the validation returns an error. 

Here are some examples of Quantity for which division by 0.01 returns Int, so the validation passes:
8.11
68.15
2130.46
5895083

And here are some examples of Quantity for which division by 0.01 returns non-Int, so the validation fails:
8.1
2.07
19628.87
323.39
40428.2

I believe some adjustments to this validation should be provided.

fenollp:
Hello! Indeed IEEE floats seem to be getting in the way. https://pkg.go.dev/math/big#Rat should be used instead.
If you're down to open a PR with this fix as well as with regression tests, I'll gladly review it! Thanks

maxxant:
Good discussion and proposal for golang "decimal float types (IEEE 754-2008)" :  
https://github.com/golang/go/issues/19787

Good explanation "Why don't you just use big.Rat" for money and trading or financial application:
https://github.com/shopspring/decimal#why-dont-you-just-use-bigrat

fenollp:
Hey @maxxant thanks for doing this!

Indeed big.Rat usage is full of footguns. `decimal` is too short on precision for JSON (float64). `decimal128` looks good enough to go with it!
We should keep that choice of lib hidden (to the API) so as to be able to switch to a future golang/std impl.
This means tweaking `openapi3.Schema` de-serialization of `Min` `Max` and `MultipleOf`, also `json.Number`.
If we can't manage this well we'll that'd just be a breaking change and new version.

Feel free to publish your attempts at this! I'll help :)

PR #1068, opened 2025-05-08
openapi3: replace bigfloat with decimal128 to fix rounding errors during validation

Revolyssup:
Fixes https://github.com/getkin/kin-openapi/issues/817

Following the suggestion here https://github.com/getkin/kin-openapi/issues/817#issuecomment-1823392667

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/go-openapi/jsonpointer"
 	"github.com/mohae/deepcopy"
+	"github.com/woodsbury/decimal128"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


openapi3/schema.go
```diff
@@ -1643,7 +1644,10 @@ func (schema *Schema) visitJSONNumber(settings *schemaValidationSettings, value
 	if v := schema.MultipleOf; v != nil {
 		// "A numeric instance is valid only if division by this keyword's
 		//    value results in an integer."
-		if bigFloat := big.NewFloat(value / *v); !bigFloat.IsInt() {
+		numParsed, _ := decimal128.Parse(fmt.Sprintf("%.10f", value))
```
Revolyssup, line 1647: Going higher, example %.15f here will break test case like `19628.87`
As using %.15f turns this to 19628.869999999989.... causing non zero remainder later. I think it's okay to lose a little precision here to have precision in the remainder.

Issue #1072, opened 2025-06-02
`OneOf` schema results in an error message that is hard to parse and customize


mayurm88:
A schema like the following results into a validation error message that is not customizable as it does not return a `MultiError` type but instead returns a string of concatenated error messages to the application.

```yaml
Policy:
  description: The policy instance to be created
  required: true
  content:
    application/json:
      schema:
        oneOf:
          - $ref: '#/components/schemas/Reference'
          - $ref: '#/components/schemas/Instance'
        discriminator:
          propertyName: type
          mapping:
            reference: '#/components/schemas/Reference'
            instance:  '#/components/schemas/Instance'
CreateGatewayRoutePolicyInstance:
  type: object
  oneOf:
    - $ref: '#/components/schemas/HeaderPolicy'
    - $ref: '#/components/schemas/WasmPolicy'
    - $ref: '#/components/schemas/TokenPolicy'
    - $ref: '#/components/schemas/LimitPolicy'
  discriminator:
    propertyName: name
    mapping:
      header: '#/components/schemas/FingerprintDeviceHeaderPolicy'
      wasm: '#/components/schemas/HashHeaderWithWasmPolicy'
      token: '#/components/schemas/IntrospectTokenPolicy'
      limit: '#/components/schemas/LimitPolicy'
BasePolicy:
  type: object
  properties:
    id:
      type: string
      format: uuid
      readOnly: true
    name:
      type: string
      description: The name identifier of the policy
    config:
      type: object
      description: The configuration of the policy
LimitPolicy:
  allOf:
    - $ref: '#/components/schemas/BasePolicy'
    - type: object
      properties:
        name:
          type: string
          enum: ["limit"]
        config:
          type: object
          properties:
            limit:
              type: integer
            size:
              type: integer
            key:
              type: string
          required: ["limit", "size"]
      required: ["name", "config"]
```

```sh
http POST :80/policies 'Content-Type: application/json' type=instance name="limit"  'config[ceiling]':=10

{
[...]

PR #1073, opened 2025-06-02
fix: surface multi-errors in Origin field in cases of oneOf schema errors

mayurm88:
Resolves https://github.com/getkin/kin-openapi/issues/1072


mayurm88:
Realized that we could always unwrap the error using `errors.As` or `errors.Unwrap` to get the underlying `MultiError` since we're using `fmt.Errorf`. 🤦🏼‍♂️

PR #1087, opened 2025-08-03
Preserve all validation errors for allOf

alexbakker:
Previously, only the first error would be reported


alexbakker:
@fenollp Done!

PR #1090, opened 2025-08-17
openapi3filter: support primitive parsing for individual text like parts in multipart/form-data

nmeheus:
All parts of a multipart/form-data request that are decoded using the PlainBodyDecoder (which is the default for omitted or text/plain parts) are considered to be of type string without these parts being parsed according to their specified schema.
This has been amended and tested in this PR.
This implementation relies on the fact that either an omitted or specified text/plain Content-Type part header will result in a string typed value.
Only parsing of primitive values is supported, as there seems no specifications available for array of object typed parts, as is the case for the different types of supported parameters according to OpenAPI. If this were to change, it should be relatively straightforward to amend the code to make use of an additional implementation of the valueDecoder interface specifically for this purpose.

### DIFF
--- a/openapi3filter/issue949_test.go
+++ b/openapi3filter/issue949_test.go
@@ -1,0 +1,115 @@
+package openapi3filter_test
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"mime/multipart"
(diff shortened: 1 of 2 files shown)


fenollp:
Thanks!

PR #1101, opened 2025-09-25
Reject requests with body when non expected

CynanX:
When a the specification does not specify a request body but one is included the validation currently passes.

This change will reject the request.

An example use case is when validation a DELETE request where you do not expect a body.


CynanX:
@fenollp Hi. I believe this solves issue 1100. I'm not sure of the protocol to get this reviewed? I couldn't see a contributing guide so forgive me if I've not done this correctly. Thank you for you time.

PR #1154, opened 2026-04-20
openapi3: add JoinFunc for custom $ref path resolution

reuvenharrison:
## Problem

`LoadFromDataWithPath` + `ReadFromURIFunc` is the supported extension point for loading specs from non-filesystem sources. The caller provides a virtual base path and a custom reader. However, when the spec contains relative `$ref`s, the internal `join()` function resolves them using `path.Dir()` before `ReadFromURIFunc` ever sees the result. There is no way for the caller to control how relative paths are resolved.

This breaks callers whose virtual paths use a `prefix:filepath` convention (e.g. git objects via `git show origin/main:openapi.yaml`). `path.Dir("origin/main:openapi.yaml")` returns `"origin"`, so `$ref: "./schemas/pet.yaml"` resolves to `origin/schemas/pet.yaml` instead of `origin/main:schemas/pet.yaml`. The mangled path is then passed to `ReadFromURIFunc`, which cannot recover the correct prefix.

## Fix

Add a `JoinFunc` callback on `Loader` — similar to `ReadFromURIFunc` — that lets callers override how relative `$ref` paths are resolved against the base path. When `JoinFunc` is nil (the default), behavior is unchanged.

```go
loader := openapi3.NewLoader()
loader.JoinFunc = func(basePath *url.URL, relativePath *url.URL) *url.URL {
    // custom resolution logic
}
```

## Test plan

- [x] `TestDefaultJoin` — verifies the default resolution logic (relative, dot-slash, parent directory, nil base)
- [x] `TestJoinFunc` — end-to-end: loads a multi-file spec through a prefixed virtual path using `JoinFunc` + `ReadFromURIFunc` + `LoadFromDataWithPath`
- [x] `ExampleLoader_JoinFunc` — godoc Example showing the full pattern
- [x] All existing tests pass (`go test ./...`)
- [x] Regenerated `openapi3.txt`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -820,6 +820,13 @@ type Loader struct {
 	// ReadFromURIFunc allows overriding the any file/URL reading func
 	ReadFromURIFunc ReadFromURIFunc
 
+	// JoinFunc allows overriding how relative $ref paths are resolved against
+	// a base path. When set, it is called instead of the default join logic
+	// that uses path.Dir and path.Join. This is useful when loading specs from
+	// non-filesystem sources (e.g. git objects, remote archives) where the base
+	// path follows a different convention than filesystem paths.
+	JoinFunc func(basePath *url.URL, relativePath *url.URL) *url.URL
+
 	Context context.Context
 
 	// Has unexported fields.
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -278,7 +278,16 @@ func join(basePath *url.URL, relativePath *url.URL) *url.URL {
 		return relativePath
 	}
 	newPath := *basePath
-	newPath.Path = path.Join(path.Dir(newPath.Path), relativePath.Path)
+	// Handle git ref paths like "origin/main:openapi.yaml" where ":"
+	// separates the ref from the file path. path.Dir does not understand
+	// this syntax and would treat the colon as a regular character.
+	if i := strings.IndexByte(newPath.Path, ':'); i >= 0 {
+		prefix := newPath.Path[:i+1] // e.g. "origin/main:"
+		filePath := newPath.Path[i+1:]
+		newPath.Path = prefix + path.Join(path.Dir(filePath), relativePath.Path)
+	} else {
+		newPath.Path = path.Join(path.Dir(newPath.Path), relativePath.Path)
+	}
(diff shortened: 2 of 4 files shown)


fenollp (changes requested):
name:     "git ref with relative path",
			base:     "origin/main:openapi.yaml",
			rel:      "schemas/pet.yaml",
			expected: "origin/main:schemas/pet.yaml",

This isn't how I understand paths to work. The colon has no special meaning in paths.

If you'd like your own `ReadFromURIFunc` to do some special handling for git paths so be it but (1) I am very unfamiliar with these "git paths" and (2) I fail to see how that behavior would ever be useful in this lib?

oasdiff-test:
You're right that `:` has no special meaning in filesystem paths. But I think the issue is broader than git paths specifically.

`LoadFromDataWithPath` + `ReadFromURIFunc` is the supported extension point for loading specs from non-filesystem sources (remote archives, databases, git objects, etc.).
The caller provides a virtual base path and a reader that knows how to fetch from that path. This works well, except that `join()` resolves relative `$refs` against the base path using `path.Dir()` before the `ReadFromURIFunc` ever sees the result.

The problem is that `join()` assumes the path follows filesystem directory semantics, but `LoadFromDataWithPath` accepts arbitrary paths. There's no way for the caller to intercept or customize how relative refs are resolved: `join()` mangles the path and then hands the mangled result to `ReadFromURIFunc`.

Two possible solutions:

1. This PR (minimal: detect a prefix separator)
2. A `JoinFunc` callback (similar to `ReadFromURIFunc`) that lets callers override path resolution for their virtual paths

Would either of these be acceptable? Happy to implement option 2 if you prefer. It keeps kin-openapi agnostic about the path format while giving callers the control they need.

fenollp:
Yes to me option (1) is too specific to support without opting in. I'd like (2) along with an Example test please.

reuvenharrison:
Updated per your feedback — replaced the colon-specific logic with a `JoinFunc` callback on `Loader`. Added `ExampleLoader_JoinFunc` as the godoc Example. Ready for another look.

fenollp (approved):
LGTM
Will merge after the storm!

PR #1161, opened 2026-04-28
revert to go 1.25 and revert cc4f8d99

### LINKED ISSUE #1160: Please revert the Go 1.26 upgrade in go.mod
It makes this library unusable in projects that work with a 2 version policy (e.g. Hugo):

https://github.com/getkin/kin-openapi/pull/1152

fenollp:
Closes https://github.com/getkin/kin-openapi/issues/1160

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -121,6 +121,11 @@ var IncludeOrigin = false
 
 FUNCTIONS
 
+func BoolPtr(value bool) *bool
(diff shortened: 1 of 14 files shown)


bep:
Thanks! Note that I wouldn't argue against you going for Go 1.26 only if there were some "must have" feature in that release you needed, but I don't think that's the case here.

fenollp:
I understood as much. No worries

PR #1162, opened 2026-04-29
feat(openapi3conv): canonicalization pass for 3.0 -> 3.x

reuvenharrison:
> **Status: draft.** Opening early to gauge whether this fits in kin-openapi proper, and to settle a couple of small open questions before marking ready for review.

## Motivation

kin-openapi parses both OpenAPI 3.0 and 3.1, sharing one `openapi3.T` type. Schema-level constructs serialize differently between the two versions but represent the same semantics:

- `nullable: true` (3.0) ↔ `type: ["string", "null"]` (3.1)
- `minimum: 5, exclusiveMinimum: true` (3.0) ↔ `exclusiveMinimum: 5` (3.1)
- `example: v` (3.0) ↔ `examples: [v]` (3.1)

Downstream consumers that need a *single canonical form* (diff tools, code generators, validators) currently have to special-case both representations or accept inconsistent output. In the JS/Python ecosystems this is a solved problem with multiple tools — [apiture/openapi-down-convert](https://github.com/apiture/openapi-down-convert) (3.1 → 3.0), [openapi-format](https://www.npmjs.com/package/openapi-format) (Node, upgrades 3.0 → 3.2), [Scalar's openapi-upgrader](https://github.com/scalar/scalar), [openapi-downgrade](https://pypi.org/project/openapi-downgrade/) (Python), [openapi-spec-converter](https://denseanalysis.org/blog/post/2025-03-27-openapi-spec-converter/). In Go (kin-openapi), the equivalent slot — paralleling the existing `openapi2conv` — is empty.

This PR fills that slot.

## Scope

- **In scope**: 3.0 ↔ 3.1 ↔ 3.2 ↔ any future 3.x where representational differences can be rewritten in place. OpenAPI 3.2 is purely additive over 3.1 (structured tags, streaming, `additionalOperations`, OAuth device flow, `defaultMapping`), so 3.1 → 3.2 is a version-string change with no further rewrites.
- **Out of scope**: 3.x → 3.0 (lossy by nature).
- **Out of scope**: cross-major upgrades (3 → 4 if/when v4 ships). Those belong in a dedicated package mirroring `openapi2conv`. The package errors out explicitly with a message pointing at that pattern.

## What it does

The 3.0 → 3.1 direction is **mechanical and lossless**: every 3.0 construct has a direct 3.1 form per the [OAI upgrade guide](https://learn.openapis.org/upgrading/v3.0-to-v3.1.html). The package walks the schema graph and rewrites in place.

| 3.0 | 3.1 |
|---|---|
| `openapi: 3.0.x` | `openapi: 3.1.1` (or any 3.x via `Target`) |
| `nullable: true` + `type: T` | `type: [T, "null"]` |
| `nullable: true` + `type: [T, U]` | `type: [T, U, "null"]` (deduped) |
| `nullable: true` (no `type`) | drop `nullable` |
| `minimum: x` + `exclusiveMinimum: true` | `exclusiveMinimum: x`; `minimum` cleared |
| `maximum: x` + `exclusiveMaximum: true` | `exclusiveMaximum: x`; `maximum` cleared |
| `exclusiveMinimum: false` / `exclusiveMaximum: false` | dropped (default) |
| Schema `example: v` | `examples: [..., v]` |

## API

```go
package openapi3conv

// Upgrade canonicalizes doc into the representation of opts.Target.
// Cross-major and downgrade attempts return an error.
func Upgrade(doc *openapi3.T, opts UpgradeOptions) error

// UpgradeTo31 is a convenience wrapper for Upgrade with Target = "3.1.1".
func UpgradeTo31(doc *openapi3.T) error

// UpgradeSchema canonicalizes a single schema (and its descendants) in place.
func UpgradeSchema(s *openapi3.Schema)

type UpgradeOptions struct {
    Target          string    // e.g. "3.1.1", "3.2.0"; defaults to "3.1.1"
    SkipVersionBump bool      // leave doc.OpenAPI alone
    Verbose         io.Writer // log each rewrite
}
```

In-place mutation (vs returning a new doc à la `openapi2conv.ToV3`) is intentional: kin-openapi already shares one `openapi3.T` type between 3.0, 3.1, and 3.2, so no type conversion is happening — only field rewrites. Deep-copying the entire tree just to leave the input untouched would be wasteful and error-prone for refs and components.

The walk is cycle-safe via a `map[*openapi3.Schema]struct{}` visited set.

## Tests

24 tests covering all transformations, idempotence, custom targets including 3.2, the cross-major / downgrade / bad-version error paths, operation-tree walking (parameters, request bodies, responses, callbacks), cycle safety, verbose logging, and nil safety.

```
$ go test ./openapi3conv/
ok  github.com/getkin/kin-openapi/openapi3conv  0.562s
```

Full `go test ./...` is green; no changes outside the new package.

## Open questions

1. **Default `Target`** — currently `3.1.1` (the OAI guide uses this). Open to `3.1.0` if you prefer the floor.
2. **Naming** — `openapi3conv` parallels `openapi2conv`. The package's purpose is upgrading *within* 3.x rather than between major versions, so the name is slightly off-key. Open to alternatives.

## Out of scope

Lower-priority items I deliberately did not include:

- **File-upload normalization** (`format: byte` → `contentEncoding: base64`, multipart `format: binary` → `contentMediaType`). Context-dependent on media type and rarely the bottleneck.
- **`$schema` dialect declaration** on Schema Objects. Opt-in enhancement, not a representational difference.
- **Direct 2.0 → 3.1**. Composes cleanly: `openapi2conv.ToV3` then `openapi3conv.Upgrade`.

Happy to address any of these in follow-ups if there's interest.

### DIFF
--- a/openapi3conv/openapi3_conv.go
+++ b/openapi3conv/openapi3_conv.go
@@ -3,64 +3,93 @@ package openapi3conv
 import (
 	"fmt"
 	"io"
+	"strconv"
+	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
 )
 
 // DefaultTargetVersion is the OpenAPI version string written when bumping the
-// document version. Set to the current 3.1 patch release; the OAI upgrade
-// guide uses the same.
+// document version. The OAI upgrade guide uses the current 3.1 patch release.
 const DefaultTargetVersion = "3.1.1"
 
 // UpgradeOptions controls per-pass behaviour.
 type UpgradeOptions struct {
+	// Target is the version string written when SkipVersionBump is false.
+	// Defaults to DefaultTargetVersion. Currently any 3.x version is accepted;
+	// representational rewrites only exist for 3.0 → 3.1, since 3.1 → 3.2 is
+	// purely additive (no breaking changes). Future minor versions that
+	// introduce representational changes can extend the dispatch in Upgrade.
+	Target string
+
 	// SkipVersionBump leaves doc.OpenAPI unchanged. Useful for consumers
 	// that want representations canonicalized while preserving the stated
 	// version (e.g., a pre-diff normalization step).
 	SkipVersionBump bool
 
-	// TargetVersion is the version string written when SkipVersionBump is
-	// false. Defaults to DefaultTargetVersion.
-	TargetVersion string
-
 	// Verbose, if non-nil, receives one line per rewrite for debugging.
 	Verbose io.Writer
 }
 
-// UpgradeTo31 rewrites every 3.0-form construct in doc into its 3.1 form
-// in place: bumps the version, replaces nullable: true with type arrays,
-// replaces boolean exclusiveMinimum/exclusiveMaximum with numeric form,
-// replaces example with examples. Idempotent on already-3.1 documents.
+// Upgrade canonicalizes doc into the representation of opts.Target in place.
 //
-// Returns an error only if the document is structurally invalid (e.g., nil
-// document).
-func UpgradeTo31(doc *openapi3.T) error {
-	return UpgradeTo31WithOptions(doc, UpgradeOptions{})
-}
-
-// UpgradeTo31WithOptions is the variant with explicit options.
-func UpgradeTo31WithOptions(doc *openapi3.T, opts UpgradeOptions) error {
+// The schema-level rewrites the walker applies (nullable → type array, boolean
+// exclusive bounds → numeric, example → examples) are idempotent and
+// convergent on the 3.1+ form. Calling Upgrade on an already-3.1 (or 3.2)
+// document is a no-op aside from the version bump.
+//
+// Cross-major upgrades are not supported. OpenAPI 4 (or any future major
+// version) will require a separate package mirroring the openapi2conv
+// pattern (which converts Swagger 2.0 documents to OpenAPI 3.0). Returning
+// an error here keeps that boundary explicit.
+func Upgrade(doc *openapi3.T, opts UpgradeOptions) error {
 	if doc == nil {
 		return fmt.Errorf("openapi3conv: doc is nil")
 	}
 
-	target := opts.TargetVersion
+	target := opts.Target
 	if target == "" {
 		target = DefaultTargetVersion
+	}
+
+	srcMajor, srcMinor, err := parseVersion(doc.OpenAPI)
+	if err != nil {
+		return fmt.Errorf("openapi3conv: invalid doc.OpenAPI %q: %w", doc.OpenAPI, err)
+	}
+	tgtMajor, tgtMinor, err := parseVersion(target)
+	if err != nil {
+		return fmt.Errorf("openapi3conv: invalid Target %q: %w", target, err)
+	}
+
+	if srcMajor != tgtMajor {
+		return fmt.Errorf(
+			"openapi3conv: cross-major upgrade not supported (%s -> %s); "+
+				"a separate package is the right home for cross-major conversions "+
+				"(see openapi2conv for the existing 2 -> 3 pattern)",
+			doc.OpenAPI, target,
+		)
+	}
+	if tgtMinor < srcMinor {
+		return fmt.Errorf("openapi3conv: cannot downgrade %s to %s", doc.OpenAPI, target)
 	}
 
 	w := &walker{
 		visited: map[*openapi3.Schema]struct{}{},
 		opts:    opts,
 	}
+	w.walkDoc(doc)
 
 	if !opts.SkipVersionBump && doc.OpenAPI != target {
 		w.logf("openapi: %s -> %s", doc.OpenAPI, target)
 		doc.OpenAPI = target
 	}
-
-	w.walkDoc(doc)
 	return nil
+}
+
+// UpgradeTo31 is a convenience wrapper for Upgrade with Target = "3.1.1".
+// Idempotent on already-3.1 documents.
+func UpgradeTo31(doc *openapi3.T) error {
+	return Upgrade(doc, UpgradeOptions{})
 }
 
 // UpgradeSchema canonicalizes a single schema (and its descendants) in place.
@@ -72,6 +101,24 @@ func UpgradeSchema(s *openapi3.Schema) {
 	}
 	w := &walker{visited: map[*openapi3.Schema]struct{}{}}
 	w.walkSchema(s)
+}
+
+// parseVersion splits an OpenAPI version string ("3.0.3", "3.1.1", "3.2.0")
+// into major and minor integers. Patch and pre-release suffixes are ignored.
+func parseVersion(v string) (major, minor int, err error) {
+	parts := strings.SplitN(v, ".", 3)
+	if len(parts) < 2 {
+		return 0, 0, fmt.Errorf("expected MAJOR.MINOR[.PATCH], got %q", v)
+	}
+	major, err = strconv.Atoi(parts[0])
+	if err != nil {
+		return 0, 0, fmt.Errorf("invalid major %q: %w", parts[0], err)
+	}
+	minor, err = strconv.Atoi(parts[1])
+	if err != nil {
+		return 0, 0, fmt.Errorf("invalid minor %q: %w", parts[1], err)
+	}
+	return major, minor, nil
 }
 
 // walker carries cycle-tracking state and verbose output across the schema
--- a/openapi3conv/openapi3_conv.go
+++ b/openapi3conv/openapi3_conv.go
@@ -3,6 +3,7 @@ package openapi3conv
 import (
 	"fmt"
 	"io"
+	"slices"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+		doc.OpenAPI = target
+	}
+	return nil
+}
+
+// UpgradeTo31 is a convenience wrapper for Upgrade with Target = "3.1.1".
+// Idempotent on already-3.1 documents.
+func UpgradeTo31(doc *openapi3.T) error {
+	return Upgrade(doc, UpgradeOptions{})
+}
```
fenollp: I don't see a need to specify the "to which version" to upgrade to, nor the "skipversionbump" options.

To me the `const DefaultTargetVersion` should be dropped, the UpgradeOptions should be private and be passed using a `WithWriter(w)` func. The target should always be latest (3.2.0 today), thus UpgradeTo31 needn't exist.

openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+	if !s.Nullable {
+		return
+	}
+	if s.Type != nil && len(*s.Type) > 0 {
+		alreadyHasNull := false
+		for _, t := range *s.Type {
+			if t == openapi3.TypeNull {
+				alreadyHasNull = true
+				break
+			}
```
fenollp: Please use slices.Contains

openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+		}
+		if !alreadyHasNull {
+			newTypes := append(openapi3.Types(nil), *s.Type...)
+			newTypes = append(newTypes, openapi3.TypeNull)
```
fenollp: You mean simply this?
```go
			newTypes := append(*s.Type, openapi3.TypeNull)
```

openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+type walker struct {
+	visited map[*openapi3.Schema]struct{}
+	opts    UpgradeOptions
+}
+
+func (w *walker) logf(format string, args ...any) {
+	if w.opts.Verbose == nil {
+		return
+	}
+	fmt.Fprintf(w.opts.Verbose, format+"\n", args...)
```
fenollp: Minor nit: call fmt.Fprintln(w, "") instead of editing the format string

openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+// The schema-level rewrites the walker applies (nullable → type array, boolean
+// exclusive bounds → numeric, example → examples) are idempotent and
+// convergent on the 3.1+ form. Calling Upgrade on an already-3.1 (or 3.2)
+// document is a no-op aside from the version bump.
+//
+// Cross-major upgrades are not supported. OpenAPI 4 (or any future major
+// version) will require a separate package mirroring the openapi2conv
+// pattern (which converts Swagger 2.0 documents to OpenAPI 3.0). Returning
+// an error here keeps that boundary explicit.
+func Upgrade(doc *openapi3.T, opts UpgradeOptions) error {
```
fenollp: I don't think this needs to be able to return an error: document should first be Validate()'d, then there should not be a way to upgrade to a version anterior to oldest v3 (3.2.0 today).

openapi3conv/openapi3_conv.go
```diff
@@ -0,0 +1,381 @@
...
+					}
+				}
+			}
+		}
+	}
+
+	if doc.Paths != nil {
+		for _, pathItem := range doc.Paths.Map() {
+			w.walkPathItem(pathItem)
+		}
```
fenollp:
```go
func (paths *Paths) Map() (m map[string]*PathItem) {
	if paths == nil || len(paths.m) == 0 {
		return make(map[string]*PathItem)
	}
```
please avoid superfluous `if .. != nil`

fenollp (changes requested):

reuvenharrison:
Thanks for the careful review, Pierre. Quick summary of how I'm planning to handle each:

- **Style nits** (lines 135, 184, 316, 320): all four agreed, will push in next iteration.
- **No-error return** (line 45): agreed in spirit; one open detail noted in my reply there.
- **Drop `Target` + always upgrade to latest** (line 93): this is the one I'd like to discuss before changing — full argument in my inline reply. Headline: I think `Target` serves a real case (teams whose CI tooling doesn't yet understand newer OAS versions). Happy to drop it if the case doesn't land.

Will reply inline on each.

openapi3conv/openapi3_conv.go
reuvenharrison: Coming back to this with a stronger case for keeping `Target` than I had in the top-level reply.

The real concern isn't speculative downstream tooling — it's **any direct consumer of `openapi3conv.Upgrade()` that has version-conditional logic of its own**.

Consider what happens when OAS 3.3 ships and kin-openapi releases an update bumping the internal "latest" const:

- Every consumer that calls `Upgrade(doc)` in their own code now gets back a 3.3.0-stamped document, with no code change on their side. Just a `go get -u`.
- Consumers whose own logic checks `doc.OpenAPI` to dispatch behavior (validators, generators, diff tools, doc renderers, anything with version-conditional code paths) suddenly receive a version they may not yet have implementations for.
- Their workarounds — pin kin's version (defensive but means falling behind on bug fixes), or post-process `doc.OpenAPI = "3.2.0"` (hacky), or fork the rewrite logic — all hurt.

`WithTarget(version)` keeps the simple case simple — default is still "latest" — while giving direct consumers an escape hatch that doesn't require them to fight the library:

```go
// Most users — same simplicity you proposed:
openapi3conv.Upgrade(doc)

// Consumers with their own version-conditional logic:
openapi3conv.Upgrade(doc, openapi3conv.WithTarget("3.1.1"))
```

The cost of keeping `Target` is one optional `WithTarget(string) Option` function and a single field in the private options struct. The cost of dropping it is paid by every direct consumer of `openapi3conv` whenever a new OAS major-minor lands — silently, without compile-time warning.

Happy to drop it if you still feel the simplification wins, but wanted you to weigh that consumer-side cost explicitly. Your call either way.

fenollp: I mean the upgrade doesn't happen "suddenly", it's the result of bumping the lockfile.
But anyway my actual point is v3.x versions are all strictly compatible (see milestones 3.2.1 and 3.3.0 (as of today) at https://github.com/OAI/OpenAPI-Specification/milestones ). Knowing this I expect (and maybe I am mistaken) that tools stemming from handling v3.x documents should also be strictly compatible.
In other words, I expect v3.x changes to always be minor (except v3.0 -> v3.1, which we've started to handle anyway).

However in your usage if you do have a case for WithTarget I would love to discuss this.

reuvenharrison:
Pushed five of your six items in 7c84a48 (the design ones still pending the [Target discussion](https://github.com/getkin/kin-openapi/pull/1162#discussion_r3195083522)):

- `logf` uses `Fprintln(w)` for the trailing newline (line 135)
- dropped the superfluous nil-before-`Map()` check (line 184)
- `slices.Contains` replaces the hand-rolled "already has null" loop (line 316)
- simpler in-place append in `rewriteNullable` (line 320)
- removed `SkipVersionBump` from `UpgradeOptions` plus its test — callers can `original := doc.OpenAPI; ...; doc.OpenAPI = original` if they want the escape hatch

`Target` retention, error return, private options + functional `WithX` are all held pending the design call on the inline thread.

openapi3conv/openapi3_conv.go
reuvenharrison: Fair points. The 3.x strict-compat commitment (per the 3.2.1 and 3.3.0 milestones) means a tool that handles 3.1 handles 3.2+ too — so once everything's canonicalized to ≥3.1, the version string is cosmetic. No concrete oasdiff case for `Target`; dropping it.

Pushing the rest of your line-93 asks now: drop `Target`, drop error return, private options, `WithWriter`, drop `DefaultTargetVersion`. `UpgradeTo31` goes too.

reuvenharrison:
Pushed the rest in 32160b0. New API:

```go
openapi3conv.Upgrade(doc, openapi3conv.WithWriter(w))   // optional
```

Always upgrades to the latest 3.x version the package knows about (`3.2.0` today; the const moves when a new minor lands). No error return, no `Target`, no public options struct. `UpgradeTo31`, `DefaultTargetVersion`, and `parseVersion` are gone.

Tests rewritten to match: dropped the `Target` / cross-major / downgrade / invalid-version cases (no longer reachable), `TestUpgradeTo31_*` renamed to `TestUpgrade_*`, cycle test uses an actual `time.After` timeout, verbose test uses `WithWriter`.

All your line-93 asks are now in. Re-requesting review.

openapi3conv/openapi3_conv.go
reuvenharrison: Done in 7c84a48 — switched to slices.Contains.

openapi3conv/openapi3_conv.go
reuvenharrison: Done in 7c84a48 — simplified to in-place append.

openapi3conv/openapi3_conv.go
reuvenharrison: Done in 7c84a48 — split into Fprintf + Fprintln.

openapi3conv/openapi3_conv.go
reuvenharrison: Done in 7c84a48 — dropped the nil-before-Map() check.

openapi3conv/openapi3_conv.go
reuvenharrison: Done in 32160b0 — Upgrade no longer returns an error. Doc must be Validate()'d first.

openapi3conv/openapi3_conv.go
reuvenharrison: All four addressed in 32160b0: Target dropped, SkipVersionBump dropped, options privatized behind WithWriter (Option-style), DefaultTargetVersion replaced with internal latestTargetVersion = 3.2.0. UpgradeTo31 also gone.

fenollp (approved):
LGTM.

Small point: in your tests you may want to ensure each document verifies `.Validate(..)` before and after `.Upgrade(..)`. Also, you could add that property to the TestV3ApisGuruOpenapiDirectory test if you'd like.

reuvenharrison:
Thanks for the approval! Pushed 56035f5 addressing your small point — most Upgrade tests now use a `upgradeAndAssertValid` helper that calls `.Validate()` on both sides. Two tests opt out and document why (they intentionally feed a 3.0-stamped doc carrying 3.1-only constructs to exercise the rewrite, so the input doesn't validate as 3.0).

Skipped wiring this into `TestV3ApisGuruOpenapiDirectory` for now — that's a much bigger surface and a separate kind of test. Easy to add as a follow-up if you'd like.

Ready when you are.

fenollp:
Cool! Ready now:)
>Skipped wiring this into TestV3ApisGuruOpenapiDirectory for now — that's a much bigger surface and a separate kind of test. Easy to add as a follow-up if you'd like.

Yes I'd like that. I'll try it myself.

PR #1166, opened 2026-05-04
feat(openapi3): add per-type validation errors with cluster wrappers

reuvenharrison:
## Motivation

The document validation walker (`T.Validate`, `Info.Validate`, `Paths.Validate`, `Operation.Validate`, etc.) returns plain `errors.New(...)` values for ~50 distinct failure modes. Downstream code that wants to react to a specific failure has to compare on the human-readable message string:

```go
if err.Error() == \"value of version must be a non-empty string\" { ... }
```

Brittle: any upstream rewording silently breaks the consumer. There's no current way to map from an error to a stable identifier.

This shows up concretely in tools that aggregate findings across many specs — diff/lint tools, governance dashboards, CI integrations — where users need a stable rule ID to suppress (`--ignore-rules`) or filter on. Today they're forced into regex-bucket workarounds against the message text.

## What this PR adds

A typed `*ValidationError` carrying a stable `Code` (kebab-case subject-action, e.g. `info-version-required`) alongside the existing `Message`:

```go
// openapi3/validation_error.go
type ValidationError struct {
    Code    string
    Message string
}

func (e *ValidationError) Error() string { return e.Message }
```

Then convert each `errors.New(msg)` site in the validate paths to wrap the message:

```go
// Before
return errors.New(\"value of version must be a non-empty string\")

// After
return &ValidationError{
    Code:    \"info-version-required\",
    Message: \"value of version must be a non-empty string\",
}
```

This commit converts only the two sites in `info.go` to demonstrate the migration pattern. Bulk conversion of the remaining ~50 sites is deferred until the design settles.

## Backward compatibility

Fully additive — no existing behavior changes. Three pinned in tests:

| Caller pattern | Behavior |
|---|---|
| `err.Error() == \"value of version must be a non-empty string\"` | ✅ Same — `(*ValidationError).Error()` returns the original byte-identical message |
| `if err != nil { ... }` | ✅ Same |
| `errors.As(err, &target)` | ✅ Better — gains a new working target (`*ValidationError`); still works for everything that was working before |
| `MultiError` traversal | ✅ Same — `MultiError.As()` already recurses into elements (`openapi3/errors.go:41-49`); typed errors flow through naturally |

Three new tests pin all three:

- `TestValidationError_BackwardCompatibleErrorString` — `err.Error()` returns the same string as before for both info-Validate failure modes
- `TestValidationError_StructuredCodeViaErrorsAs` — `errors.As(err, &ve)` succeeds and yields the expected Code
- `TestValidationError_FlowsThroughMultiError` — `errors.As` reaches into a MultiError to find a wrapped `*ValidationError`

Full `go test ./...` is green; no changes outside the new file plus the two-site demonstration in `info.go`.

## Precedent

`openapi3` already exposes typed errors (`*SchemaError`, used internally with `errors.As` at three sites in `schema.go`) and a public sentinel (`ErrURINotSupported` in `loader_uri_reader.go`). `*ValidationError` is in the same direction — additive, not a paradigm shift.

## Open questions

1. **Code field type.** Currently bare `string`. Alternative: a typed `type ValidationCode string` plus a const block (`const InfoVersionRequired ValidationCode = \"info-version-required\"`). Typed-constant gives compile-time autocomplete and grep-ability but adds maintenance surface. Bare string is more flexible. I defaulted to bare; happy to switch if you prefer.
2. **Code naming convention.** I'm using `<subject>-<action>` kebab-case (modeled after oasdiff's rule IDs and GitHub Actions `::error::` titles). Open to alternatives — `OAS-XXXX` numeric, dotted (`oas.info.version.required`), etc.
3. **Bulk conversion scope.** Should this PR end up landing the full ~50-site conversion, or split into per-package follow-up PRs (one for paths/operation/response, one for schema, etc.)? Either works.
4. **Worth doing at all?** If you'd rather see this solved a different way, or not at all, I'd rather hear that now than iterate further.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -256,6 +256,10 @@ type AdditionalProperties = BoolSchema
     AdditionalProperties is a type alias for BoolSchema, kept for backward
     compatibility.
 
+type AnchorFieldFor31Plus struct{ ValidationError }
+
+func (e *AnchorFieldFor31Plus) As(target any) bool
+
 type BoolSchema struct {
 	Has    *bool
 	Schema *SchemaRef
@@ -373,6 +377,10 @@ func (m Callbacks) JSONLookup(token string) (any, error)
 func (callbacks *Callbacks) UnmarshalJSON(data []byte) (err error)
     UnmarshalJSON sets Callbacks to a copy of data.
 
+type CommentFieldFor31Plus struct{ ValidationError }
+
+func (e *CommentFieldFor31Plus) As(target any) bool
+
 type ComponentRef interface {
 	RefString() string
 	RefPath() *url.URL
@@ -411,6 +419,10 @@ func (components *Components) Validate(ctx context.Context, opts ...ValidationOp
     Validate returns an error if Components does not comply with the OpenAPI
     spec.
 
+type ConstFieldFor31Plus struct{ ValidationError }
+
+func (e *ConstFieldFor31Plus) As(target any) bool
+
 type Contact struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
 	Origin     *Origin        `json:"-" yaml:"-"`
@@ -433,6 +445,10 @@ func (contact *Contact) UnmarshalJSON(data []byte) error
 
 func (contact *Contact) Validate(ctx context.Context, opts ...ValidationOption) error
     Validate returns an error if Contact does not comply with the OpenAPI spec.
+
+type ContainsFieldFor31Plus struct{ ValidationError }
+
+func (e *ContainsFieldFor31Plus) As(target any) bool
 
 type Content map[string]*MediaType
     Content is specified by OpenAPI/Swagger 3.0 standard.
@@ -459,6 +475,30 @@ func (content *Content) UnmarshalJSON(data []byte) (err error)
 func (content Content) Validate(ctx context.Context, opts ...ValidationOption) error
     Validate returns an error if Content does not comply with the OpenAPI spec.
 
+type ContentEncodingFieldFor31Plus struct{ ValidationError }
+
+func (e *ContentEncodingFieldFor31Plus) As(target any) bool
+
+type ContentMediaTypeFieldFor31Plus struct{ ValidationError }
+
+func (e *ContentMediaTypeFieldFor31Plus) As(target any) bool
+
+type ContentSchemaFieldFor31Plus struct{ ValidationError }
+
+func (e *ContentSchemaFieldFor31Plus) As(target any) bool
+
+type DefsFieldFor31Plus struct{ ValidationError }
+
+func (e *DefsFieldFor31Plus) As(target any) bool
+
+type DependentRequiredFieldFor31Plus struct{ ValidationError }
+
+func (e *DependentRequiredFieldFor31Plus) As(target any) bool
+
+type DependentSchemasFieldFor31Plus struct{ ValidationError }
+
+func (e *DependentSchemasFieldFor31Plus) As(target any) bool
+
 type Discriminator struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
 	Origin     *Origin        `json:"-" yaml:"-"`
@@ -481,6 +521,18 @@ func (discriminator *Discriminator) UnmarshalJSON(data []byte) error
 func (discriminator *Discriminator) Validate(ctx context.Context, opts ...ValidationOption) error
     Validate returns an error if Discriminator does not comply with the OpenAPI
     spec.
+
+type DynamicAnchorFieldFor31Plus struct{ ValidationError }
+
+func (e *DynamicAnchorFieldFor31Plus) As(target any) bool
+
+type DynamicRefFieldFor31Plus struct{ ValidationError }
+
+func (e *DynamicRefFieldFor31Plus) As(target any) bool
+
+type ElseFieldFor31Plus struct{ ValidationError }
+
+func (e *ElseFieldFor31Plus) As(target any) bool
 
 type Encoding struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
@@ -600,6 +652,10 @@ func (m Examples) JSONLookup(token string) (any, error)
 func (examples *Examples) UnmarshalJSON(data []byte) (err error)
     UnmarshalJSON sets Examples to a copy of data.
 
+type ExamplesFieldFor31Plus struct{ ValidationError }
+
+func (e *ExamplesFieldFor31Plus) As(target any) bool
+
 type ExclusiveBound struct {
 	Bool  *bool    // For OpenAPI 3.0 style (modifier for min/max)
 	Value *float64 // For OpenAPI 3.1 style (actual bound value)
@@ -649,7 +705,8 @@ func (e *ExternalDocs) Validate(ctx context.Context, opts ...ValidationOption) e
     spec.
 
 type FieldVersionMismatchError struct {
-	// Field is the field name flagged (e.g. "summary", "identifier").
+	// Field is the field name flagged (e.g. "summary", "identifier",
+	// "$defs", "prefixItems", "contains", ...).
 	Field string
 	// MinVersion is the minimum OpenAPI version that allows the field
 	// (e.g. "3.1").
@@ -757,6 +814,14 @@ func (m Headers) JSONLookup(token string) (any, error)
 func (headers *Headers) UnmarshalJSON(data []byte) (err error)
     UnmarshalJSON sets Headers to a copy of data.
 
+type IDFieldFor31Plus struct{ ValidationError }
+
+func (e *IDFieldFor31Plus) As(target any) bool
+
+type IfFieldFor31Plus struct{ ValidationError }
+
+func (e *IfFieldFor31Plus) As(target any) bool
+
 type Info struct {
 	Extensions map[string]any `json:"-" yaml:"-"`
 	Origin     *Origin        `json:"-" yaml:"-"`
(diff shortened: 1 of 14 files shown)


fenollp:
Hey there! Agreed: errors should be easier to differentiate in a more Go way + strings are not reliable.

I welcome this change however I think it'd be best to introduce types instead of variants of a code string. This way errors can be differentiated not on their string but on their type + they can be wrapped in multiple types + most error strings could be turned to consts.

reuvenharrison:
Thanks for the quick look. Agreed on the principle.
Two clarifying questions before I rework:

**1.  Granularity**

Strict one-type-per-error means ~50 new exported types in openapi3:

- InfoVersionRequiredError
- PathLeadingSlashMissingError
- ...

Want that, or cluster related cases into shared types with payload fields?

- *RequiredFieldError{Path string} covers all "X must be non-empty" cases (~10 sites)
- *FieldVersionMismatchError{Field, MinVersion string} covers "X is for >=3.1" (~10 sites)
- ...

Clustering keeps the public API in the 10-15-type range while preserving type-based discrimination.

**2. Common base**

How about embedding a small base type in all types so callers who don't care about specifics can errors.As the base?

Happy to go either way.

fenollp:
Yes for the common base, the idea is to be able to differentiate as needed using any of errors.Is .As .AsType.

I think clustering where it makes sense is a great idea and is compatible with error wrapping: FieldVersionMismatchError would wrap the underlying LicenseIdentifierFieldFor31Plus error.

I'm not sure I care about the names / lengths as long as the wrapping makes sense.

reuvenharrison:
Pushed the design rework per your feedback (commit d1fb533). Three layers as discussed:

**Base** — `*ValidationError` (just carries `Message`). Catchall via `errors.As`.

**Clusters** — `*RequiredFieldError{Field, Cause}` and `*FieldVersionMismatchError{Field, MinVersion, Cause}`. Wrap leaves via `Unwrap`.

**Leaves** — `*InfoVersionRequired`, `*InfoTitleRequired`, `*LicenseIdentifierFieldFor31Plus`, `*InfoSummaryFieldFor31Plus`. One per converted site. Each implements `As` to expose the embedded base.

Demonstration set converts:

- `Info.Validate`'s two `errors.New` sites (info.title, info.version) → `*RequiredFieldError` wrapping their leaves.
- The shared `errFieldFor31Plus` helper → `*FieldVersionMismatchError` wrapping a typed leaf when the field is in the switch (`identifier`, `summary`), or a bare `*ValidationError` otherwise. The fallback path means **every existing `errFieldFor31Plus` call site immediately gains the cluster + base layers without touching its source** — only the leaf type remains to be added per site, as a 1-2 line addition.

Six tests pin the design end-to-end: backward-compat strings (every converted site's `Error()` returns the original byte-identical message), three-layer reachability for both clusters, leaf-vs-leaf discrimination, untyped-fallback behavior, MultiError flow-through.

Bulk conversion of the remaining ~50 sites can land in follow-up PRs once this shape is approved.

reuvenharrison:
Bulked the conversion in (commit 8cec548) — no follow-up PRs needed.

Coverage:

- **All `errFieldFor31Plus` dispatch** — 29 distinct fields. The 4 direct call sites (info.summary, license.identifier, doc.webhooks, doc.jsonschemadialect) plus the 25 schema fields rejected by `schema.go`'s `reject()` helper (const, examples, prefixItems, contains, minContains, maxContains, patternProperties, dependentSchemas, propertyNames, unevaluatedItems, unevaluatedProperties, if, then, else, dependentRequired, contentEncoding, contentMediaType, contentSchema, $defs, $schema, $comment, $id, $anchor, $dynamicAnchor, $dynamicRef). Each gets a typed leaf via the dispatch table in `validation_error.go`. Fields not in the table fall back to a bare `*ValidationError` so callers always retain the cluster + base layers.
- **All `RequiredFieldError` sites** — 5 fields: info.title, info.version, license.name, doc.openapi, server.url. Each gets a typed leaf wrapped in `*RequiredFieldError`.

That's ~33 new exported leaf types, each ~5 lines (type + As method).

Test coverage:

- `TestValidationError_BackwardCompatibleErrorString` — every converted site's `Error()` returns the original byte-identical message
- `TestValidationError_ThreeLayers_RequiredField` and `_FieldVersionMismatch` — three-layer reachability for both clusters
- `TestValidationError_LeafDifferentiation` — distinct leaves don't false-match each other
- `TestValidationError_AllRequiredFieldLeaves` — table-driven across the 3 newly-typed required fields
- `TestValidationError_SchemaFieldFor31PlusLeaves` — spot-checks 2 of the 25 schema-field leaves
- `TestValidationError_FieldVersionMismatch_UntypedFallback` — bare `*ValidationError` fallback path
- `TestValidationError_FlowsThroughMultiError` — all three layers reachable through `MultiError`

One pre-existing test needed an update: `TestServerValidation` was using `require.Equal` to compare against an `errors.New(...)` sentinel (which compared concrete types). Switched to `require.EqualError` on the message string, which is the contract being preserved across the migration.

Full `go test ./...` green.

fenollp:
Looks OK so far. Please have your tests use https://pkg.go.dev/errors#Unwrap and demonstrate the clustering.

reuvenharrison:
Done in 8fb8e07. New `TestValidationError_UnwrapWalksClusterToLeaf` exercises both clusters using `errors.Unwrap` directly: walks `*RequiredFieldError` → `*InfoVersionRequired` and `*FieldVersionMismatchError` → `*LicenseIdentifierFieldFor31Plus`, plus the terminal-leaf check (`Unwrap(leaf)` returns nil).

reuvenharrison:
Two more small commits since the test additions:

- **`204589f`** — dropped the unused `message` parameter from `newRequiredField` (lint hit). The cluster's `Error()` returns `Cause.Error()` so the message comes from the leaf; passing it twice was redundant.
- **`dc6e2f7`** — converted the 4 literal-string `errFieldFor31Plus("...")` call sites (`info.summary`, `license.identifier`, `webhooks`, `jsonschemadialect`) to direct per-site constructors (`newInfoSummaryFieldFor31Plus()`, etc.). Same shape as the existing `RequiredFieldError` constructors. The dispatch table now only carries the 25 schema-keyword fields reached from `schema.go`'s `reject` closure with a runtime field name — that's the only site that genuinely needs the table.

fenollp:
By the way to test usage of errors.AsType and since this repo uses go 1.25, you can write a test that mentions `//go:build 1.26` and use that function in the file. It will run only when the available toolchain is at least 1.26, which introduced said function.

reuvenharrison:
Done in b72a26f. New `validation_error_astype_test.go` is build-tagged `go1.26` and uses `errors.AsType` to reach the cluster, leaf, and base — same three layers as the existing `errors.As` test. Skipped on Go 1.25, picked up by CI's 1.26 jobs.

reuvenharrison:
One more in 0879509: switched `paths.go:109` and `path_item.go:212` from `%v` to `%w`. Both wrap inner errors during `Paths.Validate` / `PathItem` operation walking; with `%v` the chain stops there, so any typed leaf nested under a path/operation context (schema-keyword leaves like `*ConstFieldFor31Plus`, etc.) couldn't be reached via `errors.As` from the top-level `MultiError`. With `%w` the dispatch walks cleanly all the way down.

No behaviour change for anyone matching on `err.Error()` — the human-readable message is identical, only the unwrap chain changes.

fenollp (approved):
LGTM. Do you want me to merge this now or are you planning on adding commits?

reuvenharrison:
One more in 3b83447: `RequiredFieldError` and `FieldVersionMismatchError` now expose an `Origin *Origin` field carrying the source location of the offending element. Populated by the constructors from the relevant element's `Origin` (info, license, server, schema). Nil for doc-root fields (openapi version, webhooks, jsonSchemaDialect) since the loader doesn't track `*T`.

Three tests pin the contract: populated with `Loader.IncludeOrigin = true`, nil when off, nil for doc-root fields regardless. Backward compatible — `Origin` is `nil` on consumers that don't enable origin tracking.

reuvenharrison:
One more in 3c0b649: added `*SchemaValueError` cluster for the "this schema's example/default value doesn't satisfy the schema's own constraints" case. Wraps the underlying `*SchemaError` (or `MultiError` of them) returned by `VisitJSON`, carries `ValueKind` (`"example"` / `"default"`) and `Origin`.

Converts four call sites: `parameter.go`, `media_type.go`, and two in `schema.go` (default + example). `Error()` preserves the historical `"invalid example: <inner>"` / `"invalid default: <inner>"` message format byte-for-byte. Test exercises the cluster on a parameter whose example exceeds the schema's `maxLength`.

reuvenharrison:
Calling this PR feature-complete and ready for final review.

## What's in

- **Base** `*ValidationError` with `Message`
- **Clusters** (each with `Cause`, optional `Origin`, dedicated metadata fields):
  - `*RequiredFieldError` — "X must be non-empty" (5 sites: info.title/version, license.name, openapi, server.url)
  - `*FieldVersionMismatchError` — "X is for OpenAPI >=Y" (~29 sites: 4 direct + ~25 schema-keyword via the existing reject closure dispatch)
  - `*SchemaValueError` — "this schema's example/default doesn't satisfy its own constraints" (4 sites in parameter, media_type, schema)
  - `*PathParametersError` — operation declares fewer/more path params than the path template (1 site)
- **~33 leaf types**, one per converted call site, each implementing `As` to expose the embedded base
- **Origin** populated on cluster types when `Loader.IncludeOrigin = true`
- **`%v` → `%w`** fix on `paths.go:109` and `path_item.go:212` so the error chain walks cleanly

Tests cover all three layers (cluster / leaf / base) via `errors.As`, `errors.Unwrap`, and (under `//go:build go1.26`) `errors.AsType`. Backward compatibility: every converted site preserves its original `Error()` byte-for-byte.

## What's not in (the long tail)

Roughly 80–100 `errors.New(...)` / `fmt.Errorf(...)` sites across validators remain untyped — `oauth flow scopes`, `info must be an object`, response/header validations, security requirements, etc. Each follows the same shape as the conversions in this PR.

## Plan from here

Merge this PR; bulk-convert the long tail in follow-up PRs grouped by package (one for security_scheme, one for response/header, etc.). Two reasons:

1. The design is settled — those follow-ups are mechanical and shouldn't gate landing the framework.
2. Smaller per-PR diffs are easier for you to review than one giant conversion.

Each follow-up will reuse the existing clusters where they fit (most "X is missing" cases land under `RequiredFieldError`) and introduce new clusters only when a real shape calls for one (path-parameter mismatch was the latest example).

Happy to address any final review nits here. Otherwise, ready when you are.

fenollp (approved):

PR #1177, opened 2026-05-08
feat(openapi3): SchemaBothFormsExclusive cluster for union-typed schema fields

reuvenharrison:
Adds a `SchemaBothFormsExclusive` cluster covering the three schema fields stored as `BoolSchema` (a union of `Has *bool` and `Schema *SchemaRef`) where both forms are set simultaneously.

| Site | Leaf | Field |
|---|---|---|
| `schema.go:1732` | `*SchemaAdditionalPropertiesBothForms` | `additionalProperties` |
| `schema.go:1838` | `*SchemaUnevaluatedItemsBothForms` | `unevaluatedItems` |
| `schema.go:1855` | `*SchemaUnevaluatedPropertiesBothForms` | `unevaluatedProperties` |

This is a different shape from `MutuallyExclusiveFieldsError` (which is for two separate fields): here a single union-typed field has both forms populated. The cluster carries `Field string` so callers know which property fired without parsing the message.

## Backward compat

Every converted site preserves its original `Error()` string byte-for-byte.

## Tests

`TestValidationError_SchemaBothFormsLeaves` covers all three sites end-to-end.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -2087,6 +2087,28 @@ func (schema *Schema) WithUniqueItems(unique bool) *Schema
 
 func (schema *Schema) WithoutAdditionalProperties() *Schema
 
+type SchemaAdditionalPropertiesBothForms struct{ ValidationError }
+
+func (e *SchemaAdditionalPropertiesBothForms) As(target any) bool
+
+type SchemaBothFormsExclusive struct {
+	// Field is the name of the union-typed schema property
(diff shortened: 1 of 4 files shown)


fenollp (approved):
Hey there, could you turn all these PRs into a single one? Otherwise there will be a lot of rebasing to be done...

reuvenharrison:
Superseded by #1180 — combined per @fenollp's request on #1177 to avoid rebase churn. The original commit on this branch is preserved unchanged in #1180.

PR #1185, opened 2026-05-15
openapi3: aggregate independent validation errors via EnableMultiError

reuvenharrison:
## Why

Today `Validate` returns the first error it encounters and stops. A spec with several independent defects only surfaces one at a time: fix it, re-run, see the next, repeat. That serializes work that is fundamentally parallel and hides the true shape of "is my spec valid?" until the last fix lands.

## What

A new `ValidationOption`, **`EnableMultiError()`**, makes container validators aggregate independent problems and return them together as a flat `MultiError`. With the option off, behavior is unchanged.

Aggregation happens at the container fan-out points where continuing past an error is safe:

- `T.Validate` (across the doc sections: info, paths, components, security, servers, tags, externalDocs, webhooks, jsonSchemaDialect)
- `Paths.Validate` (across paths)
- `PathItem.Validate` (across HTTP methods)
- `Operation.Validate` (across parameters, requestBody, responses, externalDocs)
- `Components.Validate` (across schemas, parameters, request bodies, etc.)
- `Responses.Validate` (across status codes)

Leaf validators (`Schema`, `Info`, `Server`, individual `Parameter`, ...) still fail fast. Continuing past a leaf error can hit a nil deref because leaves often guard then dereference, so the change is deliberately scoped to the nil-guarded "for each independent sibling" sites where collect-and-continue is provably safe.

## Result shape: flat

Each top-level `MultiError` entry is **one fully-wrapped chain**, e.g.

```
SectionValidationError{Section: "paths",
  Cause: PathValidationError{Path: "/a",
    Cause: OperationValidationError{Method: GET,
      Cause: <typed leaf, e.g. RequiredFieldError{Field: "responses"}>}}}
```

Consumers iterating the `MultiError` get one entry per problem; the section/path/operation context travels with each leaf. Consumers using `errors.As` / `errors.Is` work unchanged in both modes, thanks to `MultiError.As` / `MultiError.Is` walking into contained errors.

This is achieved by an `emitWrapped(wrap, err)` helper that distributes the wrap over each leaf when the input is itself a `MultiError`, so per-section context is attached per-problem rather than wrapping a whole nested `MultiError`.

## Backward compatibility

- **Default behavior is byte-identical.** With no option passed, the validators take the same `return err`-on-first-error code path they always did. The full existing test suite passes through that path unchanged.
- **No public surface removed.** The `.github/docs/openapi3.txt` diff is purely additive (`EnableMultiError` added, nothing removed); the `docs.sh` breaking-change check returns 0.
- **`errors.As` / `errors.Is` keep working** for any consumer that uses them. Only direct type assertions on the *whole* returned error (e.g. `err.(*RequiredFieldError)`) would break in multi-error mode on a multi-defect spec — and that idiom is fragile regardless; `errors.As` is the documented alternative.

## Tests

New `validate_multi_error_test.go` covers:

- Option off: result is not a `MultiError`, first-error semantics preserved
- Option on, two bad paths: aggregated, walking the tree yields one leaf per path
- Option on, problems in two doc sections: both sections represented in the chain
- Option on, single-defect spec: still a `MultiError` (of one), `errors.As` walks through
- Option on, well-formed spec: returns nil

Full suite green.

## Open questions

- **`cmd/validate` flag?** This PR does not expose the new option through `cmd/validate`. Happy to add `--multi` as a follow-up commit on this PR (it would mirror the existing `--defaults` / `--examples` / `--patterns` flags), or leave it for a separate change. Preference?


openapi3/validate_multi_error_test.go
```diff
@@ -0,0 +1,169 @@
...
+	err := doc.Validate(context.Background(), openapi3.EnableMultiError())
+	require.Error(t, err)
+
+	var me openapi3.MultiError
+	require.True(t, errors.As(err, &me), "expected MultiError")
+	require.Equal(t, 2, countLeaves(err), "expected one leaf per bad path")
+
+	combined := err.Error()
+	require.Contains(t, combined, "/a")
+	require.Contains(t, combined, "/b")
```
fenollp: Why not use https://pkg.go.dev/github.com/stretchr/testify/require#ErrorContains ?

openapi3/validate_multi_error_test.go
```diff
@@ -0,0 +1,169 @@
...
+func TestValidate_MultiError_On_AggregatesAcrossSections(t *testing.T) {
+	// Problems in different document sections are also aggregated.
+	doc := loadDoc(t, twoBadSectionsSpec)
+	err := doc.Validate(context.Background(), openapi3.EnableMultiError())
+	require.Error(t, err)
+
+	var me openapi3.MultiError
+	require.True(t, errors.As(err, &me), "expected MultiError")
+	require.GreaterOrEqual(t, len(me), 2,
+		"expected at least one error per affected section")
```
fenollp: Why not check "exactly equal 2" ?

openapi3/response.go
```diff
@@ -75,19 +75,20 @@ func (responses *Responses) Status(status int) *ResponseRef {
 // Validate returns an error if Responses does not comply with the OpenAPI spec.
 func (responses *Responses) Validate(ctx context.Context, opts ...ValidationOption) error {
 	ctx = WithValidationOptions(ctx, opts...)
+	me := newErrCollector(ctx)
 
 	if responses.Len() == 0 {
 		return newResponsesNonEmptyRequired(responses.Origin)
```
fenollp: Why is this not wrapped in emit() ?

openapi3/components.go
```diff
@@ -108,100 +108,81 @@ func (components *Components) UnmarshalJSON(data []byte) error {
...
-			return fmt.Errorf("schema %q: %w", k, err)
+	me := newErrCollector(ctx)
+
+	validateMap := func(label string, names []string, validate func(k string) error) error {
+		for _, k := range names {
+			if idErr := ValidateIdentifier(k); idErr != nil {
+				if err := me.emit(fmt.Errorf("%s %q: %w", label, k, idErr)); err != nil {
+					return err
+				}
+				continue
```
fenollp, line 126: Let's add a comment mentioning why there's a `continue` here and in the other places it is used.

openapi3/validation_options.go
```diff
@@ -150,3 +168,82 @@ func getValidationOptions(ctx context.Context) *ValidationOptions {
...
+//
+// When multi-error mode is off, emit returns the error unchanged so the caller
+// fails fast — preserving the historical behavior byte-for-byte.
+//
+// emitWrapped applies wrap to err, distributing wrap over each leaf when err
+// is a MultiError. This is how container validators attach per-section /
+// per-path / per-operation context to each aggregated leaf.
+//
+// result returns the accumulated MultiError, or nil if none were recorded.
+type errCollector struct {
```
fenollp: Let's move this to error_collector.go

openapi3/paths.go
```diff
@@ -41,13 +41,17 @@ func WithPath(path string, pathItem *PathItem) NewPathsOption {
...
 	normalizedPaths := make(map[string]string, paths.Len())
 
 	for _, path := range paths.Keys() {
 		pathItem := paths.Value(path)
 		if path == "" || path[0] != '/' {
-			return fmt.Errorf("path %q does not start with a forward slash (/)", path)
+			if err := me.emit(fmt.Errorf("path %q does not start with a forward slash (/)", path)); err != nil {
+				return err
+			}
+			continue
```
fenollp, line 57: Ditto wrt `continue`

openapi3/openapi3.go
```diff
@@ -266,96 +266,108 @@ func (doc *T) Validate(ctx context.Context, opts ...ValidationOption) error {
...
-		if err := v.Validate(ctx); err != nil {
-			return wrap(err)
+		if err := me.emitWrapped(wrapSection("info"), v.Validate(ctx)); err != nil {
+			return err
 		}
-	} else {
-		return wrap(newInfoRequired())
+	} else if err := me.emit(wrapSection("info")(newInfoRequired())); err != nil {
+		return err
 	}
```
fenollp, line 309: Please revert that `wrap` change. I mean do keep `wrapSection` but revert `wrap` changes such that we keep setting `wrap` and thus not repeat eg. `wrapSection("info")`. See:

```go
	wrap = wrapSection("info")
	if v := doc.Info; v != nil {
		if err := me.emitWrapped(wrap, v.Validate(ctx)); err != nil {
			return err
		}
	} else if err := me.emit(wrap(newInfoRequired())); err != nil {
		return err
	}
```

fenollp (changes requested):
> Leaf validators (Schema, Info, Server, individual Parameter, ...) still fail fast. Continuing past a leaf error can hit a nil deref because leaves often guard then dereference, so the change is deliberately scoped to the nil-guarded "for each independent sibling" sites where collect-and-continue is provably safe.

I find this very dubious. I don't see any proof? It is true some validation code expects previous code to be valid and this PR does not make any note of this principle except in this short passage, and that requires hard work. But eg. license.go is definitely safe to add to this PR. Do you plan on converting more files in the future or does this stop here?

---

> cmd/validate flag? This PR does not expose the new option through cmd/validate. Happy to add --multi as a follow-up commit on this PR (it would mirror the existing --defaults / --examples / --patterns flags), or leave it for a separate change. Preference?

Yes please add a flag there.

---

I've added a few comments. Notably on some of the `continue`s introduced (all are implied).
I understand these are intended so that all the MultiErrors returned actually are errors that make sense (and are not just the result of diverging logic from previous errors).

In fact I am wondering if this should be part of the expressed documentation, or if this is just a temporary goal:

    Aggregation happens at container fan-out points (the document root, Paths,
    PathItem, Operation, Components, Responses, Webhooks). Validation of a
    single leaf element (for example, a single Schema) still stops at its first
    error.

Because to me, and I guess to others, this sentence is hard to make sense of. Plus, I don't see why we'd not return multiple errors for a single schema?

reuvenharrison:
Thanks for the thorough review.

**A. Leaf validators staying fail-fast.** Fair pushback. "Leaves are unsafe" overstates it; what's actually unsafe is the `if x.Foo == nil { return ... }` followed by `x.Foo.Bar()` pattern, where collecting past the guard would now panic. Schema, Parameter and MediaType have that shape (cross-field interdependencies, mutually-exclusive guards, internal state). The simple leaves with sequential independent checks have no such pattern.

Converted in this PR:
- `License` (your example), `Info`, `Server` (singular + `Servers` plural), `Tag` (singular + `Tags` plural), `ExternalDocs`.

Skipped `Contact` because its `Validate` has nothing to aggregate (just `validateExtensions`). `Schema`, `Parameter`, `MediaType` need per-method analysis and will follow in a separate PR.

Concrete result on a spec with empty `info.title`, empty `info.version`, and a missing `responses` field: validation now surfaces all three problems where it used to surface two.

**B. `--multi` flag on `cmd/validate`.** Added; mirrors the existing `--defaults` / `--examples` / `--patterns` flags.

**C. Docstring scope.** Rewrote: dropped the explicit fan-out list (would have gone stale), named the transition explicitly, explained why some leaves are not yet converted, and replaced the generic `errors.As` note with a concrete how-to.

Inline replies on each thread below.

openapi3/paths.go
reuvenharrison: Added comments at both Paths `continue`s (malformed path key, duplicate normalized path), plus the symmetric webhooks `continue` in `openapi3.go`.

openapi3/openapi3.go
reuvenharrison: Reverted to the `wrap = wrapSection("info")` reassignment pattern. `wrapSection` stays as the factory so each section only spells out its name once.

openapi3/validation_options.go
reuvenharrison: Moved. The `errCollector` type and its methods now live in a new `openapi3/error_collector.go`; `validation_options.go` is back to just options.

openapi3/response.go
reuvenharrison: Good catch. Routed through `me.emit` so the empty-responses error follows the same path as every other validation finding.

openapi3/validate_multi_error_test.go
reuvenharrison: Tightened from `require.GreaterOrEqual` to `require.Equal`.

openapi3/components.go
reuvenharrison: Added. The continue is there because descending into `validate(k)` for a component with a bad name only produces findings like `<bad-name>: <leaf-error>` with no resolution path until the name is fixed; skipping keeps the noise per component bounded to the one actionable identifier finding.

openapi3/validate_multi_error_test.go
reuvenharrison: Switched to `require.ErrorContains`.

reuvenharrison:
Two follow-up commits since the review, both surfaced by end-to-end testing in oasdiff:

**`e14b38a` — extend SchemaValueError wrapping to plural Examples.**
`Parameter.Validate` and `MediaType.Validate` already used `newSchemaValueError` for the singular `example:` field, but the plural `examples:` map loop returned a bare `fmt.Errorf` wrapping the underlying `*SchemaError`. That meant `errors.As(err, &SchemaValueError)` didn't match for plural examples, so downstream consumers fell back to a generic catchall instead of recognising the example-violates-schema cluster.

The plural loop now also wraps in `newSchemaValueError("example", ...)`, preserving the example key inside the wrap so the rendered message still names the offending example. Existing `example_validation_test.go` assertions and the `apis_guru` golden fixtures gained the `invalid example: ` prefix from the new wrap.

**`9b85457` — point SchemaValueError's Origin at the example's `value:`.**
The plural Examples Validate failure used the parent struct's Origin, so the resolved location deep-linked to the parameter's or media-type's start (e.g. the `in: query` line) rather than the offending example value. Added an `exampleValueOrigin` helper that pulls `Origin.Fields["value"]` from the Example struct when available, falling back to the example's struct origin, then to the parent.

No message format change in this one. The Cause string still matches what existed; only the line/column on the resolved Source tightens to the example's `value:` field.

reuvenharrison:
Heads-up on the next typing step. We tested the coverage and ~78% of findings now resolve to typed clusters with line/column; the remaining ~22% fall through to a catchall and trace back to five untyped kin sites:

- `path parameter %q must be required` (`parameter.go:327`)
- `operations %q and %q have the same operation id %q` (`paths.go:221`, in `validateUniqueOperationIDs`)
- `extra sibling fields: %+v` (`extension.go:26` plus the per-Ref `UnmarshalJSON` sites in `refs.go`)
- `schema %q: unsupported 'type' value %q` and the no-schema-name variant

Plan: ship those as a separate "type the remaining bare-error sites" PR after this one merges, to keep this PR focused on aggregation. If anything in the list jumps out as needing a different cluster shape or name, easier to discuss now than re-litigate post-merge.

openapi3/response.go
```diff
@@ -75,19 +75,23 @@ func (responses *Responses) Status(status int) *ResponseRef {
...
 func (responses *Responses) Validate(ctx context.Context, opts ...ValidationOption) error {
 	ctx = WithValidationOptions(ctx, opts...)
+	me := newErrCollector(ctx)
 
 	if responses.Len() == 0 {
-		return newResponsesNonEmptyRequired(responses.Origin)
+		if err := me.emit(newResponsesNonEmptyRequired(responses.Origin)); err != nil {
+			return err
+		}
+		return me.result()
```
fenollp: I don't think this line is needed.

fenollp: I actually think keeping this line in would be a bug: validation would never aggreggate with extension errors.

Other than this LGTM

reuvenharrison: Good catch, that early return blocked validateExtensions. Fixed in `83543b6` (drop the `return me.result()` and fall through to the trailing `me.finalize(validateExtensions(...))`; the for loop is a no-op on an empty Responses). Regression test in `61003a9` (`TestResponses_Validate_EmptyAndExtensionAggregate`): constructs an empty Responses with a bogus non-x- sibling key on Extensions and asserts both findings surface; reverting the fall-through edit makes it fail.

reuvenharrison: Audited the rest of the converted validators for the same pattern. Only two other `return me.result()` calls exist (`Tags.Validate` in tag.go, `Servers.Validate` in server.go) and both are in the trailing position with nothing after them; those plural slice types do not have `Extensions`, so the singular `Tag` and `Server` validators are where the extension check runs. Every other guard in the converted set already uses `if err := me.emit(...); err != nil { return err }`, which only returns in fail-fast mode and falls through to the trailing `me.finalize(validateExtensions(...))` in multi mode. So nothing else to fix here.

PR #1186, opened 2026-05-16
openapi3: cache compiled JSON Schema 2020-12 validator per *Schema

efritz:
`useJSONSchema2020` calls `newJSONSchemaValidator(schema)` on every invocation. Compilation is expensive (marshal → unmarshal → keyword transform → `jsonschema.Compile`), and `openapi3filter` automatically takes this path for every OpenAPI 3.1 request, making it dominant per-request overhead for long-lived servers.

This PR caches the compiled `*jsonSchemaValidator` in a package-level `sync.Map` keyed by `*Schema`: first call compiles, subsequent calls do a single `Load`, races converge via `LoadOrStore`. Compilation failures are not cached (deterministic given the schema; sentinel value not worth the complexity).

## Benchmark

Scratch benchmark validating a small object schema repeatedly (`darwin/arm64`, M4 Pro, `-benchtime=2s`):

| | ns/op | B/op | allocs/op |
|---|---:|---:|---:|
| Before | 39,030 | 59,628 | 714 |
| After |    832 |  1,434 |  34 |
| | **~47×** | **~42×** | **~21×** |

<details><summary>benchmark source</summary>

```go
package openapi3_test

import (
    "testing"

)

func BenchmarkJSONSchema2020Validator(b *testing.B) {
    maxLen := uint64(64)
    age0, age150 := float64(0), float64(150)
    schema := &openapi3.Schema{
        Type: &openapi3.Types{"object"},
        Properties: openapi3.Schemas{
            "name":  &openapi3.SchemaRef{Value: &openapi3.Schema{Type: &openapi3.Types{"string"}, MinLength: 1, MaxLength: &maxLen}},
            "age":   &openapi3.SchemaRef{Value: &openapi3.Schema{Type: &openapi3.Types{"integer"}, Min: &age0, Max: &age150}},
            "email": &openapi3.SchemaRef{Value: &openapi3.Schema{Type: &openapi3.Types{"string"}, Format: "email"}},
        },
        Required: []string{"name", "age"},
    }
    value := map[string]any{"name": "Jane Doe", "age": 30, "email": "jane@example.com"}

    if err := schema.VisitJSON(value, openapi3.EnableJSONSchema2020()); err != nil {
        b.Fatal(err)
    }
    b.ResetTimer()
    for i := 0; i < b.N; i++ {
        if err := schema.VisitJSON(value, openapi3.EnableJSONSchema2020()); err != nil {
            b.Fatal(err)
        }
    }
}
```

</details>

### DIFF
--- a/openapi3/schema_jsonschema_validator.go
+++ b/openapi3/schema_jsonschema_validator.go
@@ -200,13 +200,31 @@ func formatValidationError(verr *jsonschema.ValidationError, parentPath string) 
 	}
 }
 
-// useJSONSchema2020 validates using the JSON Schema 2020-12 validator
+// useJSONSchema2020 validates using the JSON Schema 2020-12 validator.
+//
+// The compiled validator is cached in [compiledJSONSchemaValidators] so that
+// the (expensive) compilation cost is paid once per *Schema rather than once
+// per call. NOTE: racey WRT mutations to *Schema fields after the first
+// call -- mirrors the assumption already made by [compiledPatterns].
 func (schema *Schema) useJSONSchema2020(settings *schemaValidationSettings, value any) error {
+	if cached, ok := compiledJSONSchemaValidators.Load(schema); ok {
+		return cached.(*jsonSchemaValidator).validate(value)
+	}
+
 	validator, err := newJSONSchemaValidator(schema)
 	if err != nil {
-		// Fall back to built-in validator if compilation fails
+		// Fall back to built-in validator if compilation fails. Compilation
+		// failures are not cached: the compiler is deterministic given the
+		// schema, so if it fails once it will fail again, but caching the
+		// failure would require a sentinel value and the fall-through path
+		// is rare enough that the extra complexity is not worth it.
 		return schema.visitJSON(settings, value)
 	}
 
-	return validator.validate(value)
+	// LoadOrStore so that two goroutines racing on the first compilation for
+	// a given schema converge on a single cached validator. Either compiled
+	// validator is functionally identical, so it is safe to discard ours if
+	// another goroutine stored first.
+	actual, _ := compiledJSONSchemaValidators.LoadOrStore(schema, validator)
+	return actual.(*jsonSchemaValidator).validate(value)
 }
--- a/openapi3/schema_jsonschema_validator.go
+++ b/openapi3/schema_jsonschema_validator.go
@@ -202,10 +202,10 @@ func formatValidationError(verr *jsonschema.ValidationError, parentPath string) 
 
 // useJSONSchema2020 validates using the JSON Schema 2020-12 validator.
 //
-// The compiled validator is cached in [compiledJSONSchemaValidators] so that
-// the (expensive) compilation cost is paid once per *Schema rather than once
-// per call. NOTE: racey WRT mutations to *Schema fields after the first
-// call -- mirrors the assumption already made by [compiledPatterns].
+// The compiled validator is cached in compiledJSONSchemaValidators by the address
+// of the Schema object. Note that this validator is cached on first construction,
+// and any changes ot the Schema object will not be reflected in the validator used
+// later in the program.
 func (schema *Schema) useJSONSchema2020(settings *schemaValidationSettings, value any) error {
 	if cached, ok := compiledJSONSchemaValidators.Load(schema); ok {
 		return cached.(*jsonSchemaValidator).validate(value)
--- a/openapi3/schema_jsonschema_validator.go
+++ b/openapi3/schema_jsonschema_validator.go
@@ -216,10 +216,7 @@ func (schema *Schema) useJSONSchema2020(settings *schemaValidationSettings, valu
 		return schema.visitJSON(settings, value)
 	}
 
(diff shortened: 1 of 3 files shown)


openapi3/schema_jsonschema_validator_cache_test.go
```diff
@@ -0,0 +1,24 @@
...
+	v, ok := compiledJSONSchemaValidators.Load(schema)
+	require.True(t, ok)
+	require.NotNil(t, v)
+
+	// A second call should reuse the same compiled validator
+	require.NoError(t, schema.useJSONSchema2020(&schemaValidationSettings{useJSONSchema2020: true}, "world"))
+	v2, ok := compiledJSONSchemaValidators.Load(schema)
+	require.True(t, ok)
+	require.Same(t, v, v2)
+}
```
fenollp, line 24: Please also demonstrate that updating the schema updates the cache.

openapi3/schema_jsonschema_validator.go
```diff
@@ -200,13 +200,31 @@ func formatValidationError(verr *jsonschema.ValidationError, parentPath string)
...
+	}
+
 	validator, err := newJSONSchemaValidator(schema)
 	if err != nil {
-		// Fall back to built-in validator if compilation fails
+		// Fall back to built-in validator if compilation fails. Compilation
+		// failures are not cached: the compiler is deterministic given the
+		// schema, so if it fails once it will fail again, but caching the
+		// failure would require a sentinel value and the fall-through path
+		// is rare enough that the extra complexity is not worth it.
```
fenollp: Let's not keep that comment.

openapi3/schema_jsonschema_validator.go
```diff
@@ -200,13 +200,31 @@ func formatValidationError(verr *jsonschema.ValidationError, parentPath string)
 	}
 }
 
-// useJSONSchema2020 validates using the JSON Schema 2020-12 validator
+// useJSONSchema2020 validates using the JSON Schema 2020-12 validator.
+//
+// The compiled validator is cached in [compiledJSONSchemaValidators] so that
+// the (expensive) compilation cost is paid once per *Schema rather than once
+// per call. NOTE: racey WRT mutations to *Schema fields after the first
+// call -- mirrors the assumption already made by [compiledPatterns].
```
fenollp: Let's rewrite this and be more explicit in what the cache key is.
For compiledPatterns the key is the string of the pattern. Here it looks like it's a pointer?

Also explain that comment about race. It makes no sense to me.

openapi3/schema_jsonschema_validator.go
```diff
@@ -200,13 +200,31 @@ func formatValidationError(verr *jsonschema.ValidationError, parentPath string)
...
+		// is rare enough that the extra complexity is not worth it.
 		return schema.visitJSON(settings, value)
 	}
 
-	return validator.validate(value)
+	// LoadOrStore so that two goroutines racing on the first compilation for
+	// a given schema converge on a single cached validator. Either compiled
+	// validator is functionally identical, so it is safe to discard ours if
+	// another goroutine stored first.
+	actual, _ := compiledJSONSchemaValidators.LoadOrStore(schema, validator)
```
fenollp: Are you trying to argue that Store & LoadOrStore are equivalent here?
Do you have a test that shows this? I don't see goroutines in your test.

I'd rather this just did a Store unless tests show some advantage in using LoadOrStore.

fenollp (changes requested):

openapi3/schema_jsonschema_validator.go
efritz: `Store` is certainly easier and also correct. The only difference would be that the Store-only version might return two distinct Schemas which have been constructed in an equivalent manner. As Schemas objects are read-only in the downstream code this shouldn't have any behavioral difference.

We can simplify.

openapi3/schema_jsonschema_validator_cache_test.go
efritz: Ah; maybe this approach doesn't work in general for an upstream patch.

In my usage, I assumed that once we're on this code path there would be no reasonable updates to the Schema object, so the cache would be useful for the lifetime of the schema.

Should the cache be busted on _any_ change to the Schema, if we can cache validators at all?

efritz:
@fenollp We could also go with this approach: https://github.com/getkin/kin-openapi/compare/master...efritz:kin-openapi:cache-jsonschema2020-validator-2?expand=1

However, we'd still need an answer to [the cache invalidation problem](https://github.com/getkin/kin-openapi/pull/1186#discussion_r3259588331) for general application.

openapi3/schema_jsonschema_validator_cache_test.go
fenollp: Yes. A way to update the cached data needs to exist.

If you can't come up with a satisfying cache key, one way to do this could be to piggy back on the Validate() call such that it does the compilation+caching at that time, not "on every invocation".

---

WRT [your other approach](https://github.com/getkin/kin-openapi/compare/master...efritz:kin-openapi:cache-jsonschema2020-validator-2) I am not getting it. Please flesh it out more.

PR #1183, opened 2026-05-13
openapi3: typed context errors for Validate() wrapper chain

reuvenharrison:
## Why

Today, `Validate()` propagates context (which section, path, operation a leaf error lives in) by wrapping with `fmt.Errorf`:

```go
return fmt.Errorf("invalid path %s: %w", path, err)
return fmt.Errorf("invalid operation %s: %w", method, err)
return fmt.Errorf("invalid components: %w", e)
// ... 14 sites total
```

Callers that want to render the context separately have to parse the rendered text with regex. That's the brittleness the typed-validation-error work in #1166 and #1180 eliminated for leaf errors. The same reasoning applies to the wrapper layers.

## What

Three typed wrappers cover all 14 wrap sites in `Validate()`:

```go
type SectionValidationError struct {
    Section string
    Cause   error
}

type PathValidationError struct {
    Path  string
    Cause error
}

type OperationValidationError struct {
    Method string
    Cause  error
}
```

Each implements `Error()` and `Unwrap()`. The `Error()` strings are byte-identical to the `fmt.Errorf` output they replace, so existing consumers parsing rendered text continue to work unchanged. Typed extraction is purely additive.

The naming places these positional wrappers in the existing `ValidationError` family (the base `ValidationError`, the cluster types, the per-site leaves), and they live in `openapi3/validation_error_context.go` alongside the rest of that family.

## Wrap sites converted

`SectionValidationError` (12 sites: 9 in openapi3.go plus 3 `external docs` wrappers in operation.go, tag.go, schema.go):

| File | Line | Section |
|---|---|---|
| openapi3.go | 283 | components |
| openapi3.go | 290 | info |
| openapi3.go | 299 | paths |
| openapi3.go | 308 | security |
| openapi3.go | 315 | servers |
| openapi3.go | 322 | tags |
| openapi3.go | 329 | external docs |
| openapi3.go | 336 | webhooks |
| openapi3.go | 347 | jsonSchemaDialect |
| operation.go | 215 | external docs |
| tag.go | 93 | external docs |
| schema.go | 1888 | external docs |

`PathValidationError` (1 site): `paths.go:113`

`OperationValidationError` (1 site): `path_item.go:212`

## Downstream usage

```go
var sve *openapi3.SectionValidationError
if errors.As(err, &sve) { /* sve.Section */ }

var pve *openapi3.PathValidationError
if errors.As(err, &pve) { /* pve.Path */ }

var ove *openapi3.OperationValidationError
[...]

PR #1187, opened 2026-05-16
openapi3: type the remaining bare-error validation sites

reuvenharrison:
> **Depends on #1189.** The pre-existing conflicting-paths bug fix that surfaced 14 new testdata findings was extracted to that PR per @fenollp review. The `ConflictingPathsError` cluster's unit test here requires the dedup-key fix to detect the conflict it asserts on, so #1189 needs to land first and this branch will then rebase onto master.

Follow-up to #1185, as promised in [this heads-up comment](https://github.com/getkin/kin-openapi/pull/1185#issuecomment-4460324365). Testing across the ~4,000 real-world specs in `testdata/` after #1185 merged showed a substantial portion of findings falling through to a catchall and traced back to bare `fmt.Errorf` call sites in the validators that returned errors consumers could only inspect by substring matching.

**Scope: every `fmt.Errorf` call inside every `Validate` function in the openapi3 package, both bare leaves and context wrappers, is now wrapped in a typed cluster, typed leaf, or typed context wrapper.** Consumers can dispatch on every error path via `errors.As` and extract the relevant context (which field, which section, which parameter, which OAuth flow, etc.) without parsing rendered messages.

## New typed clusters

| Cluster | Where | Fields |
|---|---|---|
| `PathParameterRequiredError` | `parameter.go` | `Param string`, `Origin *Origin` |
| `DuplicateOperationIDError` | `paths.go` | `OperationID string`, `Endpoint1 string`, `Endpoint2 string`, `Origin *Origin` |
| `ExtraSiblingFieldsError` | `extension.go` + every `*Ref.validateExtras` (via `refs.tmpl`) | `Fields []string`, `Origin *Origin` |
| `SchemaTypeError` | `schema.go` | `Type string`, `Origin *Origin` |
| `InvalidParameterInError` | `parameter.go` | `Value string`, `Origin *Origin` |
| `SchemaPatternRegexError` | `schema_pattern.go` (RE2 rejecting Perl features) | `Pattern string`, `Cause error`, `Origin *Origin` |
| `InvalidSecuritySchemeTypeError` | `security_scheme.go` | `Type string`, `Origin *Origin` |
| `InvalidHTTPSchemeError` | `security_scheme.go` | `Scheme string`, `Origin *Origin` |
| `UnresolvedRefError` | `refs.tmpl`-generated `refs.go` (9 sites) + `schema.go` (25 sites) | `Ref string`, `Origin *Origin` |
| `APIKeyInInvalidError` | `security_scheme.go` | `Value string`, `Origin *Origin` |
| `PathMustStartWithSlashError` | `paths.go` | `Path string`, `Origin *Origin` |
| `ConflictingPathsError` | `paths.go` | `Path1 string`, `Path2 string`, `Origin *Origin` |
| `DuplicateParameterError` | `parameter.go` | `In string`, `Name string`, `Origin *Origin` |
| `InvalidSerializationMethodError` | `encoding.go` + `parameter.go` + `header.go` (one cluster, three sites) | `Subject string`, `Style string`, `Explode bool`, `Origin *Origin` |

## Leaf reuses of existing typed clusters

- `security_scheme.go`: `RequiredFieldError` for `OpenIDConnectURLRequired` and `SecuritySchemeFlowsRequired`; `ForbiddenFieldError` for `SecuritySchemeInForbidden`, `SecuritySchemeNameForbidden`, `SecuritySchemeBearerFormatForbidden`, `SecuritySchemeFlowsForbidden`.
- `parameter.go`: `MutuallyExclusiveFieldsError` for `ParameterExampleAndExamplesExclusive`.
- `server.go`: `RequiredFieldError` for `ServerVariableDefaultRequired`.

## New typed context wrappers

Mirrors the SectionValidationError / PathValidationError / OperationValidationError pattern from #1183. Each wraps a previously-bare `fmt.Errorf` context wrap so the surrounding context is dispatchable via `errors.As`, and Unwrap chains to the inner typed leaf.

| Wrapper | Where | Fields |
|---|---|---|
| `ComponentValidationError` | `components.go` | `Section string`, `Name string`, `Cause error` |
| `ExternalDocsURLValidationError` | `external_docs.go` | `Cause error` |
| `HeaderFieldValidationError` | `header.go` (5 sites) | `Field string`, `Cause error` |
| `MediaTypeExampleValidationError` | `media_type.go` | `ExampleName string`, `Cause error` |
| `WebhookValidationError` | `openapi3.go` | `Name string`, `Cause error` |
| `ParameterFieldValidationError` | `parameter.go` (5 sites) | `ParameterName string`, `Field string`, `Cause error` |
| `ParameterExampleValidationError` | `parameter.go` (2 sites) | `ExampleName string`, `Cause error` |
| `SecuritySchemeFlowValidationError` | `security_scheme.go` | `Cause error` |
| `OAuthFlowValidationError` | `security_scheme.go` (4 sites) | `FlowKind string`, `Cause error` |
| `OAuthFlowFieldValidationError` | `security_scheme.go` | `Field string`, `Cause error` |

## Origin tracking

Every leaf cluster carries an `Origin *Origin` field populated when the document was loaded with `Loader.IncludeOrigin = true`. To make this possible:

- `validateExtensions` gained an `origin *Origin` parameter; all 28 callers pass their receiver's `Origin`.
- `newUnresolvedRef` is the typed constructor for `UnresolvedRefError`; all 34 call sites (9 in `refs.tmpl`-generated `refs.go` + 25 in `schema.go`) call it directly with their ref-receiver's `Origin`. `refs.tmpl` updated; `refs.go` regenerated.

`SchemaPatternRegexError` chains through to the original `*SchemaError` via `Unwrap` so callers using `errors.As` against the legacy `*SchemaError` still match.

## Compatibility

`Error()` strings are byte-identical for every cluster and every context wrapper. Existing golden fixtures pass unchanged.

## Tests

`validation_error_test.go` covers, for each cluster and wrapper: typed-dispatch via `errors.As`, populated identifying fields, Origin populated when the loader tracks it (and nil otherwise), and for context wrappers, that the chain reaches the inner typed leaf via `Unwrap`.


openapi3/loader.go
```diff
@@ -20,8 +20,8 @@ import (
 // for backward compatibility but is not safe for concurrent use.
 var IncludeOrigin = false
 
-func foundUnresolvedRef(ref string) error {
-	return fmt.Errorf("found unresolved ref: %q", ref)
+func foundUnresolvedRef(ref string, origin *Origin) error {
+	return newUnresolvedRef(ref, origin)
```
fenollp: Let's drop foundUnresolvedRef in favor of newUnresolvedRef. Rename everywhere.

openapi3/paths.go
```diff
@@ -73,7 +72,7 @@ func (paths *Paths) Validate(ctx context.Context, opts ...ValidationOption) erro
 			// identical findings without new information.
 			continue
 		}
-		normalizedPaths[path] = path
+		normalizedPaths[normalizedPath] = path
```
fenollp, line 75: Why this change?

Is this the reason for the updated goldens?

fenollp (changes requested):

openapi3/paths.go
reuvenharrison: Yes. It's a bug fix:
The conflicting-paths check was unreachable: the dedupe map was keyed by the literal path while the lookup used the normalized form. Changed the set to use the normalized key so the typed cluster actually fires. Surfaced 14 real-world conflicts in testdata/APIs-guru-openapi-directory that the validator was silently accepting (apigateway, backup, carbone, circuitsandbox, healthcare.gov, jellyfin, magento, reverb, signl4, trello, visma, three vtex variants); their *__validate golden fixtures are new in this PR and pin the corrected behavior.

openapi3/loader.go
reuvenharrison: Done in 4148d86: removed the wrapper and renamed all 40+ call sites across schema.go, refs.go, and refs.tmpl to call newUnresolvedRef directly. Regenerated refs.go from the template. Full openapi3 test suite green.

openapi3/paths.go
fenollp: Please extract this bugfix to a dedicated PR.

reuvenharrison: Extracted to #1189. One subtlety: the ConflictingPathsError cluster's unit test in this PR feeds two paths that conflict by template, and without the dedup-key fix the validator silently accepts them, so the test would fail. I left the paths.go line and the 14 golden fixtures here for now so the typed-cluster tests stay green. Once #1189 lands I'll rebase this branch onto master and they drop out of the diff automatically. Updated the PR description with a Depends-on banner.

fenollp: Merged.

reuvenharrison: Rebased onto master now that #1189 has landed.
Ready to merge AFAIK.

PR #1191, opened 2026-05-30
openapi3: document that a custom ReadFromURIFunc bypasses IsExternalRefsAllowed

reuvenharrison:
## What

Clarifies, at both ref-resolution sites in `loader.go` (`loadSingleElementFromURI` and `resolveRefPath`), that installing a custom `ReadFromURIFunc` disables the built-in `IsExternalRefsAllowed` enforcement, and that the custom reader is then fully responsible for its own access policy. Comment-only — no behavior change.

## Why

The loader only calls `allowsExternalRefs` when `ReadFromURIFunc == nil`:

```go
if loader.ReadFromURIFunc == nil {
    if err := loader.allowsExternalRefs(ref); err != nil {
        return nil, err
    }
}
```

So a custom `ReadFromURIFunc` owns the external-reference decision entirely. This is reasonable — a custom reader often *needs* to resolve cross-document references that `allowsExternalRefs` would reject (it treats every cross-document `$ref` as "external"), so the policy can't be enforced ahead of the reader without breaking those readers.

The risk is that it's silent. A natural way to write a custom reader is to handle the special cases and delegate the rest to `DefaultReadFromURI`:

```go
loader.ReadFromURIFunc = func(l *openapi3.Loader, u *url.URL) ([]byte, error) {
    if isSpecialCase(u) { return readSpecialCase(u) }
    return openapi3.DefaultReadFromURI(l, u) // reads http(s) AND local files, unconditionally
}
```

That reader resolves external `$refs` **even when the caller set `IsExternalRefsAllowed = false`**, because the built-in gate was skipped. On untrusted documents this is a local-file-read (`$ref: "/etc/passwd"`) and SSRF (`$ref: "http://169.254.169.254/..."`) vector. The existing comment ("defer the external-ref decision to it") describes the behavior but not this consequence, so the footgun is easy to miss.

This PR spells out the consequence and the responsibility at both sites, so anyone writing a custom reader knows they must apply their own host/scheme allowlist or re-check `IsExternalRefsAllowed`.

## Scope

Documentation only. I deliberately did **not** change the behavior: enforcing `allowsExternalRefs` ahead of a custom reader would break readers that legitimately resolve cross-document refs the built-in gate rejects. If a safe-by-default lever is desired (e.g. an opt-in `Loader` field to re-apply the gate even with a custom reader), happy to discuss separately.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -1276,7 +1276,9 @@ func (links *Links) UnmarshalJSON(data []byte) (err error)
     UnmarshalJSON sets Links to a copy of data.
 
 type Loader struct {
-	// IsExternalRefsAllowed enables visiting other files
+	// IsExternalRefsAllowed enables visiting other files. Enforced only when
+	// ReadFromURIFunc is nil; a custom ReadFromURIFunc bypasses this flag and
+	// owns the access policy itself — see ReadFromURIFunc.
 	IsExternalRefsAllowed bool
 
 	// IncludeOrigin enables recording the file/line/column of each OpenAPI element.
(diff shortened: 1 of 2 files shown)


fenollp:
Let's move these comments to somewhere that actually appears on godoc. Maybe the struct's field?

reuvenharrison:
Good call, done.

fenollp:
Now to update the docs goldens and we should be good

PR #1194, opened 2026-06-01
openapi3: remove decimal128 dependency

alexandear:
The PR replaces `decimal128` with `math/big.Rat` for `multipleOf` validation. `decimal128` was added in #1068 by fixing #817.

`math/big.Rat` was actually the first fix suggested in #817. It works correctly here because both values are formatted via `fmt.Sprintf("%.10f", ...)` before any arithmetic, converting `float64` to an exact decimal string that `big.Rat` parses without binary floating-point error.

The regression test `TestIssue817` (added in #1068) passes with the new implementation.


fenollp:
Sounds good to me!

PR #1196, opened 2026-06-02
openapi2,openapi3: replace marshmallow with encoding/json

alexandear:
The PR removes the external `marshmallow` dependency by replacing its single-use pattern with two `json.Unmarshal` calls. Also adds `openapi2` SchemaRef unmarshal tests which had no prior coverage.

This is needed for #1192.


fenollp (commented):
Thanks!

fenollp (approved):

PR #1199, opened 2026-06-03
openapi3: typed context wrappers for combinator and tag validation

### LINKED ISSUE #1188: Additional validation cases that could be typed (follow-up to #1187)
Follow-up to #1187, which typed every `fmt.Errorf` inside every `Validate` function in the `openapi3` package. During that sweep I noted adjacent untyped sites; this records the ones that are genuinely spec `Validate()` errors and could benefit from typed clusters in a future PR. Not a prioritization signal.

**Scope (trimmed).** This now covers only errors raised from `Validate()` on the spec. An earlier version of this issue also listed loader `$ref` errors (`loader.go`), the JSON-Schema-compilation and value-validation paths (`schema_jsonschema_validator.go`, the `schema.go` value checks), and the `NewRegexpFormatValidator` constructor panic. Those are dropped: they fire during loading, during runtime value validation, or (for the constructor) at setup time, not during spec validation, so they never surface as `Validate()` findings and a typed cluster there wouldn't be reachable via the `Validate()` `errors.As` walk.

## Candidates (spec validation only)

Each entry: current bare error → proposed typed cluster.

- Duplicate value in `enum: [...]` → `EnumDuplicateValueError{Value}`
- Duplicate entry in `required: [...]` → `RequiredFieldDuplicatedError{Field}`
- `oneOf` / `anyOf` / `allOf` element validation failure → `SchemaCombinatorElementValidationError{Combinator, Cause}` (context wrapper)
- `tags` element validation failure → `TagInvalidError{Cause}` (context wrapper)

## Notes

- All `Error()` strings can be preserved byte-for-byte, following the same backward-compat pattern as #1187.
- The four-category model (Base / Cluster / Leaf / Context wrapper) documented in `validation_error.go` and `validation_error_context.go` applies.
- No ordering implied; these are candidates, ranked however the maintainers prefer.

reuvenharrison:
Follow-up to #1188, in the same vein as the context wrappers from #1183.

Two `Validate()` sites previously propagated their inner error with no scope. This adds typed context wrappers so consumers can tell *where* a failure occurred via `errors.As`, without parsing the message:

- **`SchemaCombinatorElementValidationError{Combinator, Cause}`** wraps a failure in a `oneOf` / `anyOf` / `allOf` sub-schema, naming which combinator the element belongs to.
- **`TagValidationError{Name, Cause}`** wraps a per-tag failure in the document-root `tags` list, naming the tag. (Issue #1188 sketched this as `TagInvalidError{Cause}`; renamed to match the `*ValidationError` convention in `validation_error_context.go` and given a `Name` field so the wrapper actually says which tag.)

Both are category-4 context wrappers: they add scope via `Unwrap` and do not themselves report a failure, so `errors.As` still walks through to the inner cluster/leaf.

**Backward compatibility:** unlike #1187, these add a scope prefix to `Error()` (e.g. `invalid allOf element: ...`, `tag "pet": ...`), so they are *not* byte-for-byte. The apis-guru golden fixtures are updated to match (116 files, each just gaining the prefix).

**Scope:** implements items 3 and 4 from #1188. Per the discussion there, the enum-duplicate item is dropped (uniqueness is only a SHOULD), and the required-duplicate item plus tag-name uniqueness are tracked separately as genuine MUST checks.

Closes #1188.


openapi3/testdata/apis_guru_openapi_directory/amazonaws_com_kinesisanalyticsv2_2018_05_23_openapi_yaml__validate
```diff
@@ -1 +1 @@
-invalid components: schema "AddApplicationInputRequest": error parsing regexp: invalid or unsupported Perl syntax: `(?=`
+invalid components: schema "AddApplicationInputRequest": invalid allOf element: invalid allOf element: invalid allOf element: invalid allOf element: invalid allOf element: invalid allOf element: error parsing regexp: invalid or unsupported Perl syntax: `(?=`
```
fenollp: It'd be nice to avoid the stuttering. Do you think that's doable?

reuvenharrison: Good catch, done. `SchemaCombinatorElementValidationError.Error()` now collapses a run of the same combinator, so deeply nested `allOf` renders `invalid allOf element: <cause>` once instead of repeating the prefix per level. The typed chain is untouched (`Unwrap` / `errors.As` still reach every level); only the rendered string drops the repeats, and a different combinator nested inside (e.g. `allOf` within a `oneOf`) still shows both scopes. Added a wrapper test for that and regenerated the ~30 affected apis-guru goldens.

PR #1203, opened 2026-06-04
openapi3: enforce unique required entries and unique tag names in Validate()

### LINKED ISSUE #1200: openapi3: Validate() does not enforce two spec MUSTs (unique required entries, unique tag names)
Spun out of #1188 (typed-error follow-up is in #1199). Two rules the OpenAPI / JSON Schema specs state as MUST aren't enforced by `Validate()` today:

**1. Duplicate entries in a schema's `required` array.** JSON Schema 2020-12 §6.5.3 (OpenAPI 3.1) and draft-04 (OpenAPI 3.0) both say the elements of `required` MUST be unique. `Schema.validate` doesn't check it, so this passes:

```yaml
type: object
required: [id, id]
properties: { id: { type: string } }
```

**2. Duplicate tag names in the document-root `tags` list.** The OpenAPI Object spec says "Each tag name in the list MUST be unique." `Tags.Validate` validates each tag but never checks name uniqueness across the list, so this passes:

```yaml
tags:
  - name: pet
  - name: pet
```

Both are real spec MUSTs, but enforcing them is a behavior change: docs that pass `Validate()` today would start failing. That's why I kept them out of #1188/#1199 (which only add scope and don't change what validates) and split them here.

@fenollp, are you up for making these hard `Validate()` errors, or would you rather not break existing docs over them? If you're good with it, I'll send a PR using the existing cluster pattern, something like `DuplicateRequiredFieldError{Field}` and `DuplicateTagError{Name}` (cf. `DuplicateParameterError` / `DuplicateOperationIDError`).

reuvenharrison:
Closes #1200.

Two spec MUSTs that `Validate()` did not enforce:

1. **Unique `required` entries.** JSON Schema 2020-12 §6.5.3 (OAS 3.1) and draft-04 (OAS 3.0) both require the elements of a schema's `required` array to be unique. `schema.validate` now returns a `DuplicateRequiredFieldError` on a repeat.
2. **Unique tag names.** The OpenAPI Object says "Each tag name in the list MUST be unique." `Tags.Validate` now emits a `DuplicateTagError` on a repeat (empty names are left to per-tag validation, not flagged here).

Both new error types follow the existing `Duplicate*` cluster pattern (cf. `DuplicateParameterError` / `DuplicateOperationIDError`) and carry `Origin` when the document was loaded with `IncludeOrigin = true`.

```yaml
# now rejected
type: object
required: [id, id]
properties: { id: { type: string } }
```
```yaml
# now rejected
tags:
  - name: pet
  - name: pet
```

### Behavior change
As called out in #1200, this is a behavior change: documents carrying these duplicates that passed `Validate()` before will now fail. It is wired as a hard `Validate()` error to match the other `Duplicate*` checks. If you'd prefer it gated behind a `ValidationOption` instead, happy to switch.

### Tests / fixtures
- Unit tests for both errors (`errors.As` to the typed cluster + message).
- The full `openapi3` package passes, and **no apis-guru golden fixture changed** (the corpus trips neither check), so there is no fixture churn to review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/openapi3/tag.go
+++ b/openapi3/tag.go
@@ -33,9 +33,8 @@ func (tags Tags) Validate(ctx context.Context, opts ...ValidationOption) error {
 				if err := me.emit(&DuplicateTagError{Name: v.Name, Origin: v.Origin}); err != nil {
 					return err
 				}
-			} else {
-				seen[v.Name] = struct{}{}
 			}
+			seen[v.Name] = struct{}{}
 		}
 		wrap := func(e error) error { return &TagValidationError{Name: v.Name, Cause: e} }
 		if err := me.emitWrapped(wrap, v.Validate(ctx)); err != nil {
--- a/openapi3/tag.go
+++ b/openapi3/tag.go
@@ -23,7 +23,20 @@ func (tags Tags) Validate(ctx context.Context, opts ...ValidationOption) error {
 	ctx = WithValidationOptions(ctx, opts...)
 	me := newErrCollector(ctx)
 
+	// Each tag name in the list MUST be unique (OpenAPI Object). Empty names
+	// are left to per-tag validation, not flagged as duplicates here.
+	seen := make(map[string]struct{}, len(tags))
+
 	for _, v := range tags {
+		if v.Name != "" {
+			if _, dup := seen[v.Name]; dup {
+				if err := me.emit(&DuplicateTagError{Name: v.Name, Origin: v.Origin}); err != nil {
+					return err
+				}
+			} else {
+				seen[v.Name] = struct{}{}
+			}
+		}
 		wrap := func(e error) error { return &TagValidationError{Name: v.Name, Cause: e} }
 		if err := me.emitWrapped(wrap, v.Validate(ctx)); err != nil {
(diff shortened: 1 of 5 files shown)


openapi3/tag.go
```diff
@@ -23,7 +23,20 @@ func (tags Tags) Validate(ctx context.Context, opts ...ValidationOption) error {
...
+	// are left to per-tag validation, not flagged as duplicates here.
+	seen := make(map[string]struct{}, len(tags))
+
 	for _, v := range tags {
+		if v.Name != "" {
+			if _, dup := seen[v.Name]; dup {
+				if err := me.emit(&DuplicateTagError{Name: v.Name, Origin: v.Origin}); err != nil {
+					return err
+				}
+			} else {
```
fenollp: No need for this else here no?

fenollp (changes requested):

openapi3/tag.go
reuvenharrison: fixed

PR #1204, opened 2026-06-07
openapi3gen: Add an option to customize generated field names (properties) for structs

d1vbyz3r0:
## Description
This PR adds customizable schema property naming.

Currently, openapi3gen derives property names from JSON/YAML tags only, which makes it difficult to integrate with frameworks that use different binding tags (`form`, `query`, etc.) while still relying on `openapi3gen` for schema generation.

The new option allows callers to override the generated property name before it is added to `schema.Properties` and any customizers are called.

This is particularly useful for generators and framework integrations where request binding semantics differ from JSON serialization semantics, such as Echo's `form` and `xml` binding tags.

## Usage example
```go
package main

import (
	"encoding/json"
	"fmt"
	"reflect"

	"github.com/getkin/kin-openapi/openapi3gen"
)

type AuthForm struct {
	Login    string `form:"login"`
	Password string `form:"password"`
	Ignore   string `json:"-" form:"-"`
}

func main() {
	schemas := make(openapi3.Schemas)
	g := openapi3gen.NewGenerator(
		openapi3gen.UseAllExportedFields(),
		openapi3gen.CreateFieldNameGenerator(func(field reflect.StructField, defaultName string) string {
			if tag := field.Tag.Get("form"); tag != "" && tag != "-" {
				return tag
			}
			return defaultName
		}),
	)

	ref, _ := g.NewSchemaRefForValue(&AuthForm{}, schemas)
	schema, _ := json.MarshalIndent(ref, "", "  ")
	fmt.Println(string(schema))
	// output:
	// {
	//   "properties": {
	//     "login": {
	//       "type": "string"
	//     },
	//     "password": {
	//       "type": "string"
	//     }
	//   },
	//   "type": "object"
	// }
}
```

## Changes
Existing behaviour remains unchanged when `CreateFieldNameGenerator` is not provided. 

Callback receives `defaultName`, allowing custom naming rules to preserve names resolved from existing JSON and YAML tags.

### Embedded Field Handling
While testing the new feature, I noticed that promoted fields from anonymous embedded structs were generated twice.

Anonymous structs without explicit `json` tag are already recursively exanded by `appendFields`:
https://github.com/getkin/kin-openapi/blob/6c01290ed1ab01f6c34d8d8f81e1761e44f3c108/openapi3gen/field_info.go#L39-L48

Each promoted field was then additionally processed by the dedicated embedded field branch inside `generateWithoutSaving`:
https://github.com/getkin/kin-openapi/blob/6c01290ed1ab01f6c34d8d8f81e1761e44f3c108/openapi3gen/openapi3gen.go#L347-L361

Execution subsequently continued through the common field generation path:
https://github.com/getkin/kin-openapi/blob/6c01290ed1ab01f6c34d8d8f81e1761e44f3c108/openapi3gen/openapi3gen.go#L377-L388

This caused the same field schema and `SchemaCustomizer` callbacks to be processed twice.

Previously, the duplication was hidden because both writes used the same map key. Field name customization exposed this behavior, so I removed the redundant processing branch from `generateWithoutSaving` and added a regression test.

### DIFF
--- a/.github/docs/openapi3gen.txt
+++ b/.github/docs/openapi3gen.txt
@@ -33,6 +33,13 @@ type ExportComponentSchemasOptions struct {
 	ExportGenerics         bool
 }
 
+type FieldNameGenerator func(field reflect.StructField, defaultName string) string
+    FieldNameGenerator allows client to set custom name for struct fields in
+    the generated schema. defaultName is the name, determined by generator's
+    standard JSON, YAML and Go field name resolution rules. Useful for
+    processing custom binding tags, such as `form` or `xml`. Function should
+    always return non-empty string.
+
 type Generator struct {
 	Types map[reflect.Type]*openapi3.SchemaRef
 
@@ -59,6 +66,8 @@ type Option func(*generatorOpt)
 func CreateComponentSchemas(exso ExportComponentSchemasOptions) Option
     CreateComponents changes the default behavior to add all schemas as
     components Reduces duplicate schemas in routes
+
+func CreateFieldNameGenerator(fngnrt FieldNameGenerator) Option
 
 func CreateTypeNameGenerator(tngnrt TypeNameGenerator) Option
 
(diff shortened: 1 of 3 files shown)


fenollp (approved):
Great! Thanks

PR #1206, opened 2026-06-25
openapi3: add T.WalkSchemas to visit every schema in a document

reuvenharrison:
## What

Adds `(*T).WalkSchemas(fn)`, a general-purpose traversal that visits **every schema reachable from a document exactly once**, in document order.

```go
var SkipSubtree = errors.New("skip schema subtree")
type WalkSchemasFunc func(location string, schema *SchemaRef) error
func (doc *T) WalkSchemas(fn WalkSchemasFunc) error
```

- **Complete coverage.** Components (schemas, parameters, headers, request bodies, responses, callbacks), paths and their operations (parameters, request bodies, responses, headers, callbacks), webhooks, and every sub-schema keyword: `properties`, `items`, `allOf`/`anyOf`/`oneOf`, `not`, `additionalProperties`, `prefixItems`, `contains`, `patternProperties`, `dependentSchemas`, `propertyNames`, `if`/`then`/`else`, `$defs`.
- **Cycle-safe.** Each distinct `*Schema` is visited once, so shared `$ref` targets are deduped and self-referential schemas don't loop.
- **Locatable.** The callback gets each schema's JSON Pointer (RFC 6901), for example `/paths/~1pets/get/responses/200/content/application~1json/schema`.
- **Mutation-friendly.** It passes `*SchemaRef`, so the walk serves transformers, not only read-only inspection.
- **Familiar shape.** `SkipSubtree` mirrors `filepath.SkipDir`, and any other returned error aborts the walk.

## Why

Traversing every schema in a document is something many consumers need (validation, code generation, schema transformation, `$ref`/dependency analysis, documentation), and each one currently re-implements the same recursion, which is easy to get wrong (it is simple to miss `prefixItems`, `patternProperties`, `$defs`, or a component/callback entry point). Centralizing it here makes it correct once, tested once, and consistent across consumers.

Purely additive: one new file plus tests, no changes to existing types or behavior.

## Tests

`openapi3/walk_test.go` covers full traversal across the entry points, deterministic JSON Pointer locations for inline schemas, dedup of shared `$ref`s, cycle safety (a self-referential schema), `SkipSubtree` pruning, and error-abort.

## Status

Opening as a **draft** to get feedback on the API shape (the `WalkSchemasFunc(location, *SchemaRef)` signature and the `SkipSubtree` sentinel) before finalizing. Happy to adjust naming or semantics.

### DIFF
--- a/openapi3/walk.go
+++ b/openapi3/walk.go
@@ -1,0 +1,304 @@
+package openapi3
+
+import (
+	"errors"
+	"strconv"
+	"strings"
+)
+
+// SkipSubtree, when returned by a WalkSchemasFunc, tells WalkSchemas not to
+// descend into the current schema's sub-schemas. The schema itself has already
+// been visited. Any other non-nil error stops the walk and is returned by
+// WalkSchemas. It mirrors filepath.SkipDir.
+var SkipSubtree = errors.New("skip schema subtree")
+
+// WalkSchemasFunc is called once for each schema visited by WalkSchemas.
+//
+// location is the JSON Pointer (RFC 6901) of the schema within the document,
+// e.g. "/components/schemas/Pet/properties/tag" or
+// "/paths/~1pets/get/responses/200/content/application~1json/schema". sr is
+// never nil and sr.Value is never nil; the callback may modify sr.Value in
+// place, so WalkSchemas serves transformers and not only read-only inspection.
+//
+// Returning SkipSubtree skips this schema's sub-schemas; returning any other
+// error aborts the walk and is returned by WalkSchemas.
+type WalkSchemasFunc func(location string, schema *SchemaRef) error
+
+// WalkSchemas visits every schema reachable from the document exactly once, in
+// document order, invoking fn for each. It follows resolved $ref targets
+// (schema.Value) and guards against reference cycles, so each distinct *Schema
+// is visited a single time regardless of how many references point at it.
+//
+// It covers schemas under components (schemas, parameters, headers, request
+// bodies, responses, callbacks), the paths and their operations (parameters,
+// request bodies, responses, headers, callbacks), and webhooks, then recurses
+// through every sub-schema keyword: properties, items, allOf/anyOf/oneOf, not,
+// additionalProperties, prefixItems, contains, patternProperties,
+// dependentSchemas, propertyNames, if/then/else, and $defs.
+//
+// It is useful for validation, code generation, schema transformation,
+// $ref/dependency analysis, and documentation: any consumer that needs to act
+// on every schema in a document without re-deriving the (easy to get wrong)
+// traversal itself.
+func (doc *T) WalkSchemas(fn WalkSchemasFunc) error {
+	if doc == nil {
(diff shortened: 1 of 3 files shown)


openapi3/walk.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+type schemaWalker struct {
+	fn   WalkSchemasFunc
+	seen map[*Schema]struct{}
+}
+
+// escapeJSONPointerToken escapes a single JSON Pointer reference token per
+// RFC 6901: '~' becomes '~0' and '/' becomes '~1'.
+func escapeJSONPointerToken(s string) string {
```
fenollp: Rename this in accordance with github.com/getkin/kin-openapi/blob/e56b2a193dfdeea333e71b87b2318b3ab578e58c/openapi3/loader.go#L1367

openapi3/walk.go
```diff
@@ -0,0 +1,304 @@
...
+	if !strings.ContainsAny(s, "~/") {
+		return s
+	}
+	return strings.NewReplacer("~", "~0", "/", "~1").Replace(s)
+}
+
+func (w *schemaWalker) document(doc *T) error {
+	if c := doc.Components; c != nil {
+		for name, sr := range c.Schemas {
```
fenollp: I'm not sure where "document order" is defined, but anyway shouldn't  visiting be deterministic? We're ranging over unordered maps here. Is that OK?

openapi3/walk.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return nil
+}
+
+func (w *schemaWalker) pathItem(loc string, item *PathItem) error {
+	if item == nil {
+		return nil
+	}
+	for i, pr := range item.Parameters {
+		if err := w.parameter(loc+"/parameters/"+strconv.Itoa(i), pr); err != nil {
```
fenollp: What is `location` like for specs expressed in >1 files? Maybe I'm confusing location with JSON pointer path here?

fenollp (changes requested):
I believe this is a needed addition. I'd love if this PR included and actual example with an actual use case though.

openapi3/walk.go
reuvenharrison: Done, renamed to `escapeRefString` to pair with the existing `unescapeRefString`.

openapi3/walk.go
reuvenharrison: Good catch. Dropped the vague "document order" wording and made the traversal deterministic: maps are now visited in sorted key order via `slices.Sorted(maps.Keys(...))`, so repeated walks of the same document produce the same order.

openapi3/walk.go
reuvenharrison: You were right, it is a JSON Pointer, not a file location. Renamed the parameter to `jsonPointer` and documented that it is an RFC 6901 pointer within the resolved document. It is file-agnostic: once the loader resolves external `$ref`s the document is a single tree, so the pointer addresses a position in that tree and never encodes a source file.

PR #1211, opened 2026-07-03
openapi3: support OAS 3.2 itemSchema

meoyawn:
Summary:
- add Media Type Object itemSchema support for OpenAPI 3.2
- resolve, validate, marshal, unmarshal, and walk itemSchema refs
- add SSE itemSchema regression coverage

Tests:
- task check
- verified regression tests fail on upstream/master without implementation


fenollp:
Merged in https://github.com/getkin/kin-openapi/pull/1213

PR #1214, opened 2026-07-07
openapi3: make operation responses optional for 3.1+

AlexanderWangY:
As per the OpenAPI 3.1 specification, the Operation's responses can be optional. It is valid to leave it empty.

However, `null` and `{}` are *NOT* valid states, it must either be explicitly empty and OpenAPI 3.1 or later to be valid, any `null` or `{}` values we automatically result in an error no matter the version.

You can see this spec [here](https://spec.openapis.org/oas/v3.1.0#operation-object), note how compared to the [3.0 spec](https://spec.openapis.org/oas/v3.0.3#operation-object), the responses field is no longer designated as *REQUIRED*.


fenollp:
Amazing! Thanks

PR #1215, opened 2026-07-09
openapi31: support 3.1+ reference metadata overrides

AlexanderWangY:
Adds OpenAPI 3.1 Reference Object `summary` and `description` override support. This is described in the [reference object spec](https://spec.openapis.org/oas/v3.1.2.html#reference-object).

Changed ref.go and refs.tmpl to support new summary and description fields excluding Schema type.

These overrides do **NOT** apply in OpenAPI 3.0, and therefore are ignored. Furthermore, only the following Reference Object types support overrides:

- Example (summary and description)
- Header (description only)
- Parameter (description only)
- RequestBody (description only)
- Response (description only)
- Link (description only)
- SecurityScheme (description only)

These do not apply to CallbackRef or SchemaRef, as these fields are not allowed.

Some minor details to pay attention to:
- If you don't capture IsOpenAPI31OrLater() at the start, it may be evaluated "after" resolution context changes.
- Metadata resolution/overriding is done in both mutually exclusive paths
   - Normal reference resolution
   - shouldVisitRef is false resulting in a backtracking callback
   
**Note**: Updated the Issue #513 regression test because it previously treated description beside $ref as invalid in every version. `summary` and `description` are now recognized Reference Object fields and they override supported metadata in OAS 3.1 and are **STILL ERRORS**  in OAS 3.0.

#### Testing
Added regression testing in `loader_test.go`.

You can try it out!
```bash
go test ./openapi3 -run '^TestReferenceObjectMetadataIsVersionAware$' -count=1
```

#### Unrelated
CI was not having a fun time with the formatting after template generation. Instead, I changed `refsgenerator.go` to feed the generated content into a formatter and then write to a file to preserve formatting. This is a purely cosmetic problem.

### DIFF
--- a/openapi3/issue513_test.go
+++ b/openapi3/issue513_test.go
@@ -185,7 +185,7 @@ components:
 	}
 }
 
-func TestIssue513KOMixesRefAlongWithOtherFieldsDisallowed(t *testing.T) {
+func TestIssue513ReferenceDescriptionIsVersionAware(t *testing.T) {
 	spec := `
 openapi: "3.0.3"
 info:
@@ -199,7 +199,7 @@ paths:
       summary: Delete something
       responses:
         200:
-          description: A sibling field that the spec says is ignored
+          description: Reference description
           $ref: '#/components/responses/SomeResponseBody'
 components:
   responses:
@@ -226,7 +226,16 @@ components:
 			doc, err := sl.LoadFromData([]byte(strings.ReplaceAll(spec, "3.0.3", majmin)))
 			require.NoError(t, err)
 			err = doc.Validate(sl.Context)
-			require.ErrorContains(t, err, `extra sibling fields: [description]`)
+			if majmin == "3.0" {
+				require.ErrorContains(t, err, `extra sibling fields: [description]`)
+				return
+			}
+			require.NoError(t, err)
+
+			response := doc.Paths.Value("/v1/operation").Delete.Responses.Status(200)
+			require.NotNil(t, response.Value)
+			require.NotNil(t, response.Value.Description)
+			require.Equal(t, "Reference description", *response.Value.Description)
 		})
 	}
 }
--- a/openapi3/loader.go
+++ b/openapi3/loader.go
@@ -647,7 +647,19 @@ var (
 	errMUSTSecurityScheme = errors.New("invalid securityScheme: value MUST be an object")
 )
 
+func applyHeaderRefMetadata(value *Header, component *HeaderRef, isOpenAPI31OrLater bool) *Header {
+	if !isOpenAPI31OrLater || value == nil || component.Description == nil {
+		return value
+	}
+
+	result := *value
+	result.Description = *component.Description
+	return &result
+}
+
 func (loader *Loader) resolveHeaderRef(doc *T, component *HeaderRef, documentPath *url.URL) (err error) {
+	isOpenAPI31OrLater := doc.IsOpenAPI31OrLater()
+
 	if component.isEmpty() {
 		return errMUSTHeader
 	}
@@ -657,7 +669,7 @@ func (loader *Loader) resolveHeaderRef(doc *T, component *HeaderRef, documentPat
 			return nil
 		}
 		if !loader.shouldVisitRef(ref, func(value any) {
-			component.Value = value.(*Header)
+			component.Value = applyHeaderRefMetadata(value.(*Header), component, isOpenAPI31OrLater)
 			refPath, _ := loader.resolveRefPath(ref, documentPath)
 			component.setRefPath(refPath)
 		}) {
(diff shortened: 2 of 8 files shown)


AlexanderWangY:
~~This git differ is killing me.~~ Ok so I just forgot to run doc generation.

AlexanderWangY:
Rate limited in CI 🤣

openapi3/refs.tmpl
```diff
@@ -18,6 +18,11 @@ type {{ $type.Name }}Ref struct {
 	Extensions map[string]any
 	Origin     *Origin `json:"-" yaml:"-"`
 
+	{{- if ne $type.Name "Schema" }}
+	Summary     *string `json:"summary,omitempty" yaml:"summary,omitempty"`
+	Description *string `json:"description,omitempty" yaml:"description,omitempty"`
```
fenollp, line 24: Please add the `// OpenAPI >= 3.1` comments

openapi3/ref.go
```diff
@@ -8,12 +8,15 @@ import (
...
+// Ref represents the common fields of an OpenAPI Reference Object.
+// Summary and Description are supported by OpenAPI 3.1 and later.
+// See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.2.md#reference-object
 type Ref struct {
-	Ref        string         `json:"$ref" yaml:"$ref"`
-	Extensions map[string]any `json:"-" yaml:"-"`
-	Origin     *Origin        `json:"-" yaml:"-"`
+	Ref         string         `json:"$ref" yaml:"$ref"`
+	Summary     *string        `json:"summary,omitempty" yaml:"summary,omitempty"`
+	Description *string        `json:"description,omitempty" yaml:"description,omitempty"`
```
fenollp, line 17: Please add the `// OpenAPI >= 3.1` comments

openapi3/issue513_test.go
```diff
@@ -209,14 +209,24 @@ components:
...
+	} {
+		t.Run(test.version, func(t *testing.T) {
 			t.Parallel()
 			sl := NewLoader()
-			doc, err := sl.LoadFromData([]byte(strings.ReplaceAll(spec, "3.0.3", majmin)))
+			doc, err := sl.LoadFromData([]byte(strings.ReplaceAll(spec, "3.0.3", test.version)))
 			require.NoError(t, err)
-			err = doc.Validate(sl.Context)
-			require.ErrorContains(t, err, `extra sibling fields: [description]`)
+			require.NoError(t, doc.Validate(sl.Context))
```
fenollp: First, revert cosmetic changes. `require.NoError(t, doc.Validate(sl.Context))` should stay 2 lines. That's how I like it + it adds nothing to the diff you bring. :)

Main point: 
> Note: Updated the Issue https://github.com/getkin/kin-openapi/issues/513 regression test because it previously treated description beside $ref as invalid in every version. summary and description are now recognized Reference Object fields and they override supported metadata in OAS 3.1 and are ignored in OAS 3.0, while unrelated sibling fields remain disallowed.

> are ignored in OAS 3.0

This is wrong. This test ensures they are an error in 3.0.

fenollp (changes requested):

openapi3/issue513_test.go
AlexanderWangY: I see, so any sibling field (whether its summary or descruiption) is considered erroneous for OAS 3.0. Thanks for the clarification!

AlexanderWangY: I ended up changing the `validateExtras()` function in refs.tmpl to ignore summary and description for OAS 3.1+, it still errors in 3.0.

openapi3/loader_test.go
```diff
@@ -174,6 +174,120 @@ paths:
...
+			wantDescription: "component description",
+		},
+		{
+			name:            "OpenAPI 3.1 applies reference metadata",
+			version:         "3.1.0",
+			wantSummary:     "reference summary",
+			wantDescription: "reference description",
+		},
+	} {
+		t.Run(test.name, func(t *testing.T) {
```
fenollp, line 257: Please add a t.Parallel()

openapi3/loader.go
```diff
@@ -647,7 +647,20 @@ var (
 	errMUSTSecurityScheme = errors.New("invalid securityScheme: value MUST be an object")
 )
 
+// applyHeaderRefMetadata applies OpenAPI 3.1 overrides without mutating the referenced Header.
```
fenollp: Unless I am mistaken, all these could mutate `value` directly. Also drop the comment line please.

fenollp (changes requested):
LGTM. Some notes. Please rebase.

openapi3/loader.go
```diff
@@ -647,7 +647,19 @@ var (
...
+func applyHeaderRefMetadata(value *Header, component *HeaderRef, isOpenAPI31OrLater bool) *Header {
+	if !isOpenAPI31OrLater || value == nil || component.Description == nil {
+		return value
+	}
+
+	result := *value
+	result.Description = *component.Description
+	return &result
+}
+
```
fenollp: Again, modify the value. Don't return a copy. 
You removed the comment, that's great. 
Now do this for all the other implementations...

If you can, implement that helper function using a generic type. Otherwise the repetition is fine.

fenollp (changes requested):

openapi3/loader.go
AlexanderWangY: Oops, apologies I misunderstood the last review. On it!

PR #1217, opened 2026-07-09
qol: Makefile for shorthand commands + preflight scripts

### LINKED ISSUE #1216: QoL: precommit or pre-PR script
I've run into this a little, but it would be nice to have a centralized Makefile or script to run docs.sh maps.sh, formatting, testing, etc. I have forgotten one too many times to run a generation only to have the CI (rightfully) crush my dreams.

Thoughts? Is this something that would be interesting and has anyone else faced this problem?

AlexanderWangY:
Closes #1216 

Adds a root Makefile with common commands you will probably need for generation, formatting, testing, and preparing for PRs.

This would be good to always run `make prepare` before committing for a PR,  the CI really doesn't like it if you forget to generate or format.

Help message and some minor details were generated, never trying to format printed messages manually ever again. 😢

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.DEFAULT_GOAL := help
+.DEFAULT_GOAL := prepare
 
 .PHONY: help generate format test prepare
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,31 @@
+.DEFAULT_GOAL := help
+
+.PHONY: help generate format test prepare
+
+help:
+	@printf '%s\n' \
+		'Usage: make <target>' \
+		'' \
+		'  generate  Regenerate Go code, map helpers, and API documentation.' \
+		'  format    Apply the repository formatters and import cleanup.' \
+		'  test      Run the full Go package test suite.' \
+		'  prepare   Run generation, formatting, and tests in sequence.'
+
+generate:
+	go generate ./...
+	./maps.sh
+	./docs.sh
+
+format:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,31 @@
...
+
+generate:
+	go generate ./...
+	./maps.sh
+	./docs.sh
+
+format:
+	go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -test -diff -fix -omitzero=false ./...
+	go fmt ./...
+	go install github.com/incu6us/goimports-reviser/v3@latest
```
AlexanderWangY: Hmm, not sure if we want to force people to install a package, this follows what the CI does when checking goimports, but this is rarely an issue.

Makefile
```diff
@@ -0,0 +1,31 @@
...
+		'  test      Run the full Go package test suite.' \
+		'  prepare   Run generation, formatting, and tests in sequence.'
+
+generate:
+	go generate ./...
+	./maps.sh
+	./docs.sh
+
+format:
+	go run golang.org/x/tools/go/analysis/passes/modernize/cmd/modernize@latest -test -diff -fix -omitzero=false ./...
```
AlexanderWangY: Again, not sure if we want to be asking people to pull a package to format.

fenollp:
Please put the prepare target first, so a naked make call executes it.

AlexanderWangY:
I can just set the `DEFAULT_GOAL` to be prepare, which effectively makes the behavior the same as moving prepare first.

PR #1219, opened 2026-07-11
openapi3: remove StringMap and its pre-OriginTree origin-stripping machinery

reuvenharrison:
The OriginTree rewrite (#1142) moved origin tracking to a separate pass that extracts `__origin__` before JSON conversion, so no unmarshaler sees it anymore. A family of `StringMap` / origin-stripping machinery was left behind; this removes it end to end.

### Commits
1. **Remove origin-stripping leftovers** — three dead `delete(map, originKey)` calls (`Encoding`, `Webhooks`, `Server.Variables`), the now-unused `originKey` const, and the stale `StringMap` doc comment. (`TestOrigin_*` already assert `__origin__` never appears as a map key.)
2. **Simplify the StringMap decode** — the two-step `json → map[string]any → deepCast → V` existed to strip `__origin__` and control key ordering; post-rewrite it equals a plain `json.Unmarshal`. Dropping `deepCast` also removes a `float64` round-trip precision risk on large integers.
3. **Remove the StringMap type** — its two fields become plain maps: `Discriminator.Mapping` → `map[string]MappingRef`, `OAuthFlow.Scopes` → `map[string]string` (reverting `Scopes` to its pre-#1018 type).
4. **Drop the helpers and their pass-through UnmarshalJSON** — `unmarshalStringMap`/`unmarshalStringMapP` are now trivial wrappers, and the 13 named map types that called them from a custom `UnmarshalJSON` were pure pass-throughs. Remove those methods (default map unmarshalling is identical) and delete the helpers. None of the types have a `MarshalJSON`, so marshalling was already default.
5. **Regenerate API docs** (`docs.sh`) and add the changelog entry.

### Breaking change
Removes the exported `openapi3.StringMap[V]` type (noted in the README changelog). Field-level impact is minimal — `StringMap[V]`'s underlying type is `map[string]V`, so value-level use (range/index/assign) is unaffected; only code that named the type explicitly needs a one-line update.

No behavior change; full `openapi3` suite green. (The one red test in CI, `TestExtraSiblingsInRemoteRef`, is a pre-existing flake that fetches an unreachable external host — it fails on `master` too.)


fenollp (commented):
LGTM. I like net-negative diffs!
Please rebase.

PR #1222, opened 2026-07-13
docs: move the StringMap changelog entry to v0.143.0

reuvenharrison:
#1219 added its breaking-change entry under `### v0.142.0`, but v0.142.0 was tagged shortly before that PR merged, so the StringMap removal is not in it — it ships in the next release. Move the entry to `### v0.143.0` so the changelog matches the tags.

(If the next release gets a different number, this heading should follow it.)


fenollp:
Thanks!

Issue #390, opened 2021-07-29
Any way to get more detailed errors


rtfpessoa:
Hi,

When using the `oapi.OapiRequestValidatorWithOptions` is there a way to get a more detailed error?

When I have a parameter which is an enum and the value passed is not in the enum would be great to know the field failing the validation.

Instead I get a huge message about the whole schema:
```plaintext
{"error": "code=400, message=Request body has an error: doesn't match the schema: Doesn't match schema \"oneOf\", internal=Request body has an error: doesn't match the schema: Doesn't match schema \"oneOf\"\nSchema:\n  {\n    \"description\": \"Details of an event.\",\n    \"oneOf\": [\n      {\n        \"$ref\": \"#/components/schemas/Deployment\"\n      },\n      {\n        \"$ref\": \"#/components/schemas/Change\"\n      },\n      {\n        \"$ref\": \"#/components/schemas/Incident\"\n      }\n    ],\n    \"type\": \"object\"\n  }\n\nValue:\n  {\n    \"$type\": \"incidenta\",\n    \"incident_id\": \"1627554151\",\n    \"source\": \"api\",\n    \"time_created\": \"2021-07-29T09:10:00Z\",\n    \"time_resolved\": \"2021-07-29T09:10:00Z\"\n  }\n"}
```

In this case only the field `$type` is wrong not the whole object.
Also would be even better if we could let people know what are the valid options.

Any idea if any of this is possible or if this could even be implemented? I would be up to help if some pointers are provided.

fenollp:
#230 should change the schema validation code and bring more detailed errors.

I'm not sure what you mean with `OapiRequestValidatorWithOptions`: git-grep'ing returns nothing.

> Also would be even better if we could let people know what are the valid options.

We can always write more documentation :)

rtfpessoa:
Now I see the problem. I am in the wrong project. :man_facepalming: 
I need to ask this in `github.com/deepmap/oapi-codegen`

I am sorry for the confusion

rtfpessoa:
After more digging seems like github.com/deepmap/oapi-codegen is using this project after all for the validation of the requests.

The code to do the validation is around here https://github.com/getkin/kin-openapi/blob/master/openapi3/schema.go

Do you think it would be possible to add more details as I was looking for above? Would it be worth looking deeper?

fenollp:
I mean, now I'm curious.

kevin-postman:
[...]

PR #1223, opened 2026-07-13
openapi3: stable codes for validation errors

reuvenharrison:
Validation errors have typed identities (#1018 through #1170) but no portable one: a consumer that wants to suppress a specific finding, set per-rule severity, emit machine-readable diagnostics, or link to per-rule documentation can only key off message text, which makes every wording improvement silently breaking (related asks: #390, #194).

This adds a stable, kebab-case `Code()` to every typed validation error, plus a `ValidationErrorCodes()` catalog: the rule-identifier pattern of ESLint, Spectral, and staticcheck. Codes become the contract; messages and Go type names become freely improvable.

### Design (after review)
- **Every code is a literal declared on its error type; nothing is derived.** 96 typed errors, one line each. This also completes per-site types for the three sites that lacked them: `itemSchema` (via a `fieldFor32PlusLeaves` map mirroring the 3.1 one) and the two `SchemaValueError` kinds (`ExampleViolatesSchema`, `DefaultViolatesSchema`, embedding `SchemaValueError` with an `As` bridge so `errors.As` with a `*SchemaValueError` target keeps matching).
- **Why declared codes and not `reflect.TypeOf(err).Name()`**: renaming a Go type is a normal refactor and Go users get a compile error; but these identifiers also live in config files (ignore lists, severity settings) and logs, where there is no compiler, so a type rename would break them silently. Declared codes keep the two contracts independent: types can be renamed freely, codes stay stable, and the catalog test guards them.
- `CodedError` interface for uniform access via `errors.As` (an assertion target for consumers, like `net.Error`).
- `ValidationErrorCodes()` returns the sorted catalog (92 codes; a few types share a code where the same rule fires at two sites, e.g. `in-forbidden`). `TestValidationErrorCodes` pins it bidirectionally against an inventory of every typed error, using zero values since codes are constants. An end-to-end test exercises the surface through real document validation, one case per wrapper shape.

No behavior change; full `openapi3` suite green.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -432,6 +432,8 @@ type CodedError interface {
     Renaming or removing a code is a breaking change and is recorded in the
     README changelog. The full set is available from ValidationErrorCodes.
 
+    Like net.Error, this interface exists as an assertion target for consumers;
+    the package itself constructs concrete error types and does not consume it.
     Reach the code of a wrapped error with errors.As:
 
         var coded openapi3.CodedError
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -258,6 +258,11 @@ func ValidateIdentifier(value string) error
     ValidateIdentifier returns an error if the given component name does not
     match IdentifierRegExp.
 
+func ValidationErrorCodes() []string
+    ValidationErrorCodes returns every code a validation error can carry,
+    sorted. TestValidationErrorCodes keeps it in sync with the errors the
+    validators emit.
+
 func WithValidationOptions(ctx context.Context, opts ...ValidationOption) context.Context
     WithValidationOptions allows adding validation options to a context object
     that can be used when validating any OpenAPI type.
@@ -278,6 +283,8 @@ type APIKeyInInvalidError struct {
     either omits `in:` or sets it to a value outside {query, header, cookie}.
     Carries the rejected value so callers can render or filter; empty string
     means the field was missing entirely.
+
+func (e *APIKeyInInvalidError) Code() string
 
 func (e *APIKeyInInvalidError) Error() string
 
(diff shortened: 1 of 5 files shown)


fenollp:
Aren't typed errors enough? Are you missing granularity?

reuvenharrison:
Codes address two shortcomings of typed errors:

1. The generic type families lack granularity: `FieldVersionMismatchError` alone covers ~30 validation rules, so identity there is type plus field string; the code captures that composite as one stable value.
2. A type cannot leave the process: config files (ignore lists, per-rule severities), JSON output, LSP `Diagnostic.code`, and docs anchors all need a stable string.

Types stay the Go-native identity with the structured fields; the code is its serializable name. Matching also collapses to one `errors.As` on `CodedError` plus a switch, instead of ~90 arms.

fenollp:
Matching 90 arms? I thought that's why introducing subtypes of errors was intended to avoid this. Having many (many) very specific sub-error types and type-switching on them or on their super-types.

For (2): the error type is the string. `reflect.TypeOf(err).Name()`

I am not against generating one error type per field. Would that be a problem for you?
To me, introducing strings is a step backwards when we have the machinery to define typed errors, nest them and use the type system instead of strings. Am I missing something?

reuvenharrison:
One type per field sounds good. Most of them already exist: `fieldFor31PlusLeaves` already maps each 3.1 field to its own type. The missing ones are `itemSchema` and the two `SchemaValueError` kinds. I will add them here.

With a type per field, nothing is derived anymore: each leaf type declares its code as a literal, next to the type. I will rework the PR that way and remove the string building.

I recommend to declare the string codes explicitly rather than using `reflect.TypeOf(err).Name()`: renaming a Go type is a normal refactor, and Go users get a compile error. But users also put these identifiers in config files (ignore lists, severity settings) and parse them from logs. There is no compiler there, so the rename breaks them silently. With declared codes the two are independent: types can be renamed freely, codes stay stable, and the catalog test guards them.

The catalog is the other reason: `ValidationErrorCodes()` returns the full list, so a tool can show all rules, generate docs, and test that its mapping covers everything. Reflection cannot enumerate the types that exist.

reuvenharrison:
Done: every code is now a literal declared on its error type, the string building is deleted, and the three missing per-site types are added (`ItemSchemaFieldFor32Plus`, `ExampleViolatesSchema`, `DefaultViolatesSchema`, the latter two with an `As` bridge so `errors.As` with a `*SchemaValueError` target keeps matching). The inventory test simplifies to zero-value instances.

PR #1227, opened 2026-07-15
openapi3: preserve origin for a $ref to a schema under an arbitrary top-level key

reuvenharrison:
A `$ref` to a schema stored under an arbitrary top-level key in another file (e.g. `./schemas.yaml#/User`, the Swagger-2-era "definitions bag") resolves through `T.Extensions`, a generic map that carries no origin, and the json round-trip that decodes it into the typed component drops origin entirely. With `IncludeOrigin` enabled the resolved schema then has a nil `Origin`, unlike the `/components/schemas` path.

Fix: retain each document.s origin tree when it is unmarshaled (keyed by the document, only under `IncludeOrigin`), and on that generic-map path (and only there) walk the retained tree down the ref fragment and apply the subtree, so the resolved object carries the same origins a typed resolution would, with the original file.s line numbers. No file is read or parsed twice. Best-effort: a missing tree or fragment leaves the object without origins, exactly as before. The common typed path is untouched.

### Why it matters
A consumer that uses origins for source locations (error anchors, editor links, source slicing) gets the wrong location for any element behind this ref shape: the origin chain falls back to the referencing document, pointing at the `$ref` site in the root file instead of the schema's own file and line. Whole-file refs (`./user.yaml`) and components-structured refs already carry correct origins; this closes the remaining shape.

The same applies to a same-document ref to an arbitrary top-level key (`$ref: .#/User.`): it drills through the identical Extensions path, so it is covered too, including documents loaded from memory.

Repro test + fixtures included (`TestOrigin_ExternalRefToArbitraryTopLevelKey`); full `openapi3` suite green.

### DIFF
--- a/openapi3/origin_external_ref_test.go
+++ b/openapi3/origin_external_ref_test.go
@@ -28,7 +28,21 @@ func TestOrigin_ExternalRefToArbitraryTopLevelKey(t *testing.T) {
 	require.NotNil(t, user.Origin.Key, "the origin has a key location")
 	require.Equal(t, "arbitrary_key_schemas.yaml", filepath.Base(user.Origin.Key.File),
 		"origin points at the file the schema lives in, not the referencing document")
-	require.NotZero(t, user.Origin.Key.Line, "the origin has a source line")
+	require.Equal(t, Location{
+		File:    user.Origin.Key.File,
+		Line:    1,
+		Column:  1,
+		Name:    "User",
+		EndLine: 7, EndColumn: 19,
+	}, *user.Origin.Key, "the key origin spans the whole User block in arbitrary_key_schemas.yaml")
+	require.Equal(t, 2, user.Origin.Fields["type"].Line, "field origins are attached too")
+
+	// the subtree gets origins as well, with the same file
+	id := user.Properties["id"].Value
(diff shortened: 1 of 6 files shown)


openapi3/origin_external_ref_test.go
```diff
@@ -0,0 +1,87 @@
...
+    id: { type: string }
+`
+	loader := NewLoader()
+	loader.IncludeOrigin = true
+	doc, err := loader.LoadFromData([]byte(spec))
+	require.NoError(t, err)
+
+	user := doc.Paths.Value("/users").Get.Responses.Value("200").Value.Content["application/json"].Schema.Value
+	require.NotNil(t, user.Origin, "a same-document arbitrary-key $ref carries an origin")
+	require.NotZero(t, user.Origin.Key.Line)
```
fenollp: Don't you want to assert more than just "not empty" on these origin / line values?

reuvenharrison: Good point, strengthened: the key origin is now asserted exactly (file, line, column, name, and the block end span), plus a field origin and a nested property origin, in both the external-file and from-data tests.

PR #1231, opened 2026-07-23
openapi3filter: prefer non-empty value for repeated scalar query params

### LINKED ISSUE #1230: openapi3filter: repeated query param with empty first value bypasses schema validation
**What's wrong**
When a query parameter with `allowEmptyValue: true` is supplied more than once and the first occurrence is empty, request validation accepts the request even when a later occurrence carries a value that violates the parameter's schema. The scalar decoder only reads the first value, and the empty-string short-circuit added in #1228 (fix for #1134) then returns before schema validation runs.

**Where**
- Scalar query decoding uses only `values[0]`: [openapi3filter/req_resp_decoder.go#L511-L521](https://github.com/getkin/kin-openapi/blob/30e292371a89ca6d1f91a9553e0aa518b49f838d/openapi3filter/req_resp_decoder.go#L511-L521)
- The `allowEmptyValue` empty-string bypass returns before `schema.VisitJSON`: [openapi3filter/validate_request.go#L230-L234](https://github.com/getkin/kin-openapi/blob/30e292371a89ca6d1f91a9553e0aa518b49f838d/openapi3filter/validate_request.go#L230-L234)

The decoder passes `values[0]` (the empty first occurrence) to `parsePrimitive`, so the effective value is `""`; the guard at `validate_request.go:232` sees an allowed empty string and returns `nil` without validating. The invalid later value is never checked.

**How it manifests**
Order-dependent validation. For an optional date param with `allowEmptyValue: true`:
- `?dateOfBirth=&dateOfBirth=not-a-date` → accepted (should be rejected)
- `?dateOfBirth=not-a-date&dateOfBirth=` → correctly rejected with a date-format schema error

**Repro**
1. Define an optional query param with `allowEmptyValue: true` and a `format: date` string schema.
2. Route through gorillamux; validate a request with the param supplied twice, empty first: `dateOfBirth=&dateOfBirth=not-a-date`.
3. Validation passes.
4. Reverse the order and validation returns a schema error, showing the result depends on value ordering.

**Suggested fix**
When a scalar parameter receives repeated values, don't treat an empty first value as authoritative if other supplied values exist — validate the non-empty value (or reject the ambiguous repetition) while preserving the single-empty `allowEmptyValue` behavior from #1228.

Found while running [Ito](https://ito.ai) (AI code review, free for open source) against recently merged PRs — full analysis: https://app.ito.ai/share/1ee5649c-fc14-4bb6-98fe-4fa83765a5dc.

sonnemusk:
## What

When a scalar query parameter is repeated and the first value is empty, decode the first non-empty occurrence instead of always taking `values[0]`.

## Why

With `allowEmptyValue: true`, an empty first value short-circuits schema validation. That made order-dependent results:

- `?dateOfBirth=&dateOfBirth=not-a-date` was accepted
- `?dateOfBirth=not-a-date&dateOfBirth=` was rejected

A single empty value still works as before (#1228 / #1134).

Fixes #1230.

## Test

- `TestIssue1230`
- existing `TestIssue1134`

### DIFF
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -514,10 +514,24 @@ func (d *urlValuesDecoder) DecodePrimitive(param string, sm *openapi3.Serializat
 		return nil, ok, nil
 	}
 
-	if schema.Value.Type == nil && schema.Value.Pattern != "" {
-		return values[0], ok, nil
+	// For scalar params, Go's url.Values may hold repeated keys. Prefer the first
+	// non-empty occurrence so an empty leading value cannot hide a later one
+	// from schema validation (see #1230). A single empty value is still returned
+	// so allowEmptyValue behavior is preserved.
+	raw := values[0]
+	if raw == "" && len(values) > 1 {
+		for _, v := range values[1:] {
+			if v != "" {
+				raw = v
+				break
+			}
+		}
 	}
-	val, err := parsePrimitive(values[0], schema)
+
+	if schema.Value.Type == nil && schema.Value.Pattern != "" {
+		return raw, ok, nil
+	}
+	val, err := parsePrimitive(raw, schema)
 	return val, ok, err
 }
 
--- a/openapi3filter/req_resp_decoder.go
+++ b/openapi3filter/req_resp_decoder.go
@@ -514,10 +514,9 @@ func (d *urlValuesDecoder) DecodePrimitive(param string, sm *openapi3.Serializat
 		return nil, ok, nil
 	}
 
-	// For scalar params, Go's url.Values may hold repeated keys. Prefer the first
-	// non-empty occurrence so an empty leading value cannot hide a later one
-	// from schema validation (see #1230). A single empty value is still returned
-	// so allowEmptyValue behavior is preserved.
+	// Repeated query keys: prefer the first non-empty value so an empty
+	// leading occurrence cannot hide a later value from schema validation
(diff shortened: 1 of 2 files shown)


openapi3filter/req_resp_decoder.go
```diff
@@ -514,10 +514,24 @@ func (d *urlValuesDecoder) DecodePrimitive(param string, sm *openapi3.Serializat
...
+	raw := values[0]
+	if raw == "" && len(values) > 1 {
+		for _, v := range values[1:] {
+			if v != "" {
+				raw = v
+				break
+			}
+		}
+	}
+
```
fenollp: Please remove the ai comment. 

Also, this loop should be under the condition that allowemptyvalue is true, shouldn't it?

fenollp (changes requested):

sonnemusk:
Thanks for the review.

- Dropped the verbose comment.
- On `allowEmptyValue`: `DecodePrimitive` only receives the schema, not the parameter, so it cannot gate on `parameter.AllowEmptyValue` without a wider API change. Preferring the first non-empty repeated value is the right fix for #1230 either way: with `allowEmptyValue: true`, a leading `""` would otherwise skip schema checks entirely (`validate_request.go`) and hide a later invalid value; with `allowEmptyValue: false`, an empty scalar is still validated by the schema when no non-empty sibling exists. Happy to follow up if you want `allowEmptyValue` threaded into the decoder interface.

PR #1235, opened 2026-08-02
openapi3: store field locations in a slice, not a map

reuvenharrison:
## Why

`Origin.Fields` is `map[string]Location`. A collection carries a handful of scalar fields, but a Go map allocates a whole bucket per collection whatever it holds, so these maps dominate the **retained** size of a parsed document.

Profiling a 16.8 MB spec loaded with `IncludeOrigin`, `originFromSeq` (where the inserts happen) allocated **339 MB, 13.9% of the total**. With the slice it drops out of the top six entirely.

## What

`FieldLocations`, a slice in document order. Each `Location` already carries its `Name`, so the lookup key costs nothing extra, and a linear scan beats a map at these sizes.

- `Get(name)` returns a location or the zero value; `Lookup(name)` also reports presence.
- `MarshalJSON` / `UnmarshalJSON` keep the serialized shape a name-keyed object, so **output is unchanged**.
- `Lookup` is a hand-written loop rather than `slices.IndexFunc` on purpose: the closure does not inline, so `IndexFunc` pays a call per element. Measured on 3/6/12 fields it is 5-100% slower on a hit and 2-3x slower on a miss, and misses dominate (most fields carry no recorded location). A comment says so, to save the next reader from "modernising" it.
- `sort.Slice` becomes `slices.SortFunc` at the two name-ordering sites, since `sort.Slice` swaps through reflection.
- A small copy is removed: `recordMapKeyLocations` built a whole `*Origin` purely to read its `Fields`, then copied that slice into a second one and discarded the rest. Nothing else holds it, so it is sorted in place and handed over.
- `Sequences` deliberately stays a map; the reasoning is on the `Origin` type.

## Results

16.8 MB spec, `IncludeOrigin`, median of 3 (peak from 7 runs):

| | peak RSS | total allocated | retained |
|---|---|---|---|
| before | 933 MB | 2401 MB | 478 MB |
| after | 918 MB | 2176 MB | **253 MB** |
| | -1.6% | -9.4% | **-47%** |

Retained is the target. A differ holds two documents resident for the duration of a comparison, so this halves the floor of the whole operation rather than of one load.

`originFromSeq` is the second-largest allocator, not the largest. The largest is the JSON decode, a separate concern tracked in #1236.

## Breaking change

`Origin.Fields` changes type, so `origin.Fields[name]` no longer compiles. The mechanical replacement is `origin.Fields.Lookup(name)`, same `(Location, bool)` shape. `for range` over `Fields` and `len(Fields)` are unaffected.

For scale: adapting oasdiff, a heavy consumer of this API, took 5 call sites and 13 test fixtures.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -1010,6 +1010,11 @@ func (f FieldLocations) Lookup(name string) (Location, bool)
     The scan is linear: collections have few fields, and a linear scan over a
     contiguous slice beats a map lookup at these sizes.
 
+    Deliberately a hand-written loop rather than slices.IndexFunc: the closure
+    does not inline, so IndexFunc pays a call per element. Measured on 3/6/12
+    fields it is 5-100% slower on a hit and 2-3x slower on a miss, and misses
+    are the common case here (most fields carry no recorded location).
+
 func (f FieldLocations) MarshalJSON() ([]byte, error)
     MarshalJSON keeps the serialized shape a name-keyed object, as it was when
     this was a map, so the change is invisible to anything reading the output.
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -1912,6 +1912,15 @@ type Origin struct {
     in the collection. Sequences is a map of the location of each item in
     sequence-valued fields.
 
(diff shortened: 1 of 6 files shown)


fenollp (approved):
LGTM

PR #1238, opened 2026-08-03
openapi3: decode YAML natively instead of round-tripping through JSON

### LINKED ISSUE #1236: openapi3: decoding allocates ~65x the document size, mostly re-parsing JSON it just generated
## Summary

Loading a document allocates roughly **65x its size on disk**, in the default configuration. The largest single contributor is a JSON text round trip performed in the middle of decoding: the document is decoded to `map[string]any`, re-serialized to JSON bytes, and parsed again. **About 62% of all allocation goes on parsing JSON this process generated moments earlier.**

Filing this for discussion rather than proposing a patch, since the fix has real trade-offs and the shape is worth agreeing on first.

## Measurement

Real public specs, default `Loader`, `IncludeOrigin` off:

| spec | size | total allocated | peak RSS | retained |
|---|---|---|---|---|
| [github/rest-api-description](https://github.com/github/rest-api-description) `api.github.com.yaml` | 9.3 MB | 601 MB (65x) | 208 MB | 54 MB |
| APIs.guru `stripe.com/2022-11-15` | 3.6 MB | 236 MB (66x) | 95 MB | 19 MB |

`alloc_space` for the GitHub spec, flat:

```
153 MB 25.1%  encoding/json.(*decodeState).objectInterface
 78 MB 12.8%  yaml3.(*parser).node
 56 MB  9.2%  reflect.unsafe_New
 46 MB  7.6%  reflect.mapassign_faststr0
 34 MB  5.6%  encoding/json.Unmarshal
 32 MB  5.2%  encoding/json.unquote
```

Cumulatively:

```
582 MB 95.8%  yaml.Unmarshal      (the whole decode)
377 MB 62.0%  yaml.jsonUnmarshal  (the re-parse)
361 MB 59.3%  encoding/json unmarshal
```

The 62% share is stable across specs; the 65x multiplier is not, and rises on documents that are structurally dense (short field names, few long descriptions).

## Mechanism

The yaml decode path does:

```
[]byte
  -> yaml3 Decode            -> map[string]any
  -> convertToJSONableObject
  -> json.Marshal            -> JSON bytes          <-- re-serialize
  -> json.Unmarshal          -> the target struct   <-- re-parse
```

The round trip exists for a good reason: the openapi3 types implement `UnmarshalJSON`, not `UnmarshalYAML`, so reaching them requires JSON. But a 9.3 MB document becomes an interface tree, then JSON text, then a second full parse.

JSON input avoids all of this: `unmarshal` tries `json.Unmarshal` first and returns on success, so a JSON document is parsed exactly once. The round trip is specific to YAML input.

That fast path is also the existence proof for a fix. It shows the target types can be fully populated by a single `json.Unmarshal` with no intermediate representation; the open question is only how to feed them from YAML without going through text.

## Why it is like this

Worth saying plainly, since none of it is an oversight.

The yaml layer is a fork of [ghodss/yaml](https://github.com/ghodss/yaml) (via invopop/yaml), and the round trip is its stated purpose, not a side effect. From its own package doc:

> this package first converts YAML to JSON using go-yaml and then uses `json.Marshal` and `json.Unmarshal` to convert to or from the struct. This means that it effectively reuses the JSON struct tags as well as the custom JSON methods `MarshalJSON` and `UnmarshalJSON` unlike go-yaml.

The rationale is [ghodss's 2014 post](https://web.archive.org/web/20150812020634/http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/). For openapi3 it is load-bearing: every type carries `json` tags and many implement `UnmarshalJSON`, so decoding with go-yaml directly would mean maintaining parallel `yaml` tags and `UnmarshalYAML` methods across the package.

The `map[string]any` intermediate is not a bare stepping stone either. `convertToJSONableObject` reconciles YAML and JSON semantics, and does it **target-aware**: it walks the destination `reflect.Value` alongside the value to decide coercions. Non-string YAML map keys become strings; whether a scalar stays a string depends on the destination field's type; levels that implement `UnmarshalJSON` or `TextUnmarshaler` are left alone.

So this is an 11-year-old design that solves a real problem, and the intermediate carries semantics. That constrains the fix rather than ruling it out, but it means "just skip the round trip" is not available.

## Possible directions

1. **Transcode the node tree straight to JSON bytes**, skipping `map[string]any`. Keeps `json.Unmarshal` and every `UnmarshalJSON` hook exactly as they are, and removes the interface tree, most of `reflect.mapassign_faststr0`, and the decode into `any`. The catch is the target-aware coercion above: a purely syntactic transcode would change behaviour for YAML scalars whose interpretation depends on the destination type, so those rules have to be preserved, which means threading the destination through the transcode.
2. **Decode `map[string]any` into the target directly**, without the text round trip. Keeps the coercion exactly as it is (it already runs on this representation) and only removes the serialize/re-parse, but it has to reimplement dispatch to `UnmarshalJSON`, which is where the correctness risk sits.
3. **Native `UnmarshalYAML` on the openapi3 types.** One parse, no intermediate, no JSON text.

### The one I would pursue: option 3

It measures fastest by a clear margin (numbers below), and unlike 1 and 2 it removes the intermediate rather than making it cheaper, so the 62% does not come back in a different shape later.

It is also the only one that fixes positions rather than working around them. A `*yaml.Node` already carries `Line`/`Column`, so each `UnmarshalYAML` can read the position off the node it is handed. That is a better answer than anything reachable from the round trip, and it applies to JSON documents too, since JSON parses as YAML.

Two things I did not expect when I tried it, both of which make the case better than I would have guessed:

**The methods are about three lines each.** The known field set can be read off the struct's `yaml` tags instead of restated. The JSON versions maintain that list by hand -- `Schema.UnmarshalJSON` is 91 lines, roughly 60 of them `delete` calls -- so a field added to a struct and forgotten in the list silently becomes an extension today. A reflection-derived set cannot drift. That also puts the ~30 methods closer to mechanical than to 30 individual design problems.

**The JSON hooks are the actual cost.** `Responses.UnmarshalJSON` re-marshals every child back to JSON and re-parses it, once per entry; `SchemaRef.UnmarshalJSON` parses the same bytes up to four times. Native decoding skips all of it, which is why it beats even the JSON fast path.

### What it costs, stated plainly

- **~30 methods.** Mechanical, but not free, and `Schema` is the one with real semantics in it (`$ref` siblings, the 3.1 conditionals). I have not ported it.
- **Two decoders, or a breaking change.** Keeping `UnmarshalJSON` means two implementations that must not drift. Dropping it is a breaking public API change, since callers `json.Unmarshal` these types directly.
- **One behaviour difference to decide.** A number inside an extension arrives as `float64` from `encoding/json` and `int` from the YAML decoder. Marshaling normalises it, so comparing serialized output does not catch it -- a consumer doing `Extensions["x-n"].(float64)` would panic. Either normalise on the YAML path or document it.

### Before anyone writes it

I have a spike covering four types (`Responses`, `ResponseRef`, `Response`, `MediaType`) with equivalence tests against the JSON path, which is where the numbers below come from. Extending it to the rest is a real chunk of work, so:

**Is this a direction you would take?** If option 3 is not something you want in openapi3 -- because of the two-decoders problem, the API break, or anything else -- that is a completely reasonable answer and worth saying now rather than after a large PR appears. Option 1 is the conservative alternative and I would be happy to pursue that instead.

## How much is actually recoverable

Worth stating plainly, because "62% is spent re-parsing JSON" invites the inference that removing it recovers 62%. It does not, and I would rather put the number here than have someone discover it after the work.

Measured on a 2000-endpoint synthetic document (3.7 MB YAML / 5.1 MB JSON of the same content) decoded into a struct with both tag sets and no custom unmarshalers, so every decoder does the same work:

| | ns/op | B/op |
|---|---|---|
| `encoding/json`, JSON input | 33.4M | 12.1 MB |
| native YAML decode into the struct | 120M | 84.4 MB |
| the round trip above | 184M | 143 MB |

So on a struct with no custom unmarshalers, option 3 is **1.55x** on time, not 2.6x: `json.Unmarshal` is fast, and what replaces it -- YAML struct decoding -- is not. **No YAML decoder here is within 3.5x of `encoding/json`, on fewer bytes.**

**But that understates it for openapi3, and I would rather correct myself than leave the smaller number standing.** I ported four real types -- `Responses`, `ResponseRef`, `Response`, `MediaType` -- to `UnmarshalYAML` and measured 400 responses with content and extensions:

| path | ns/op | B/op | allocs/op |
|---|---|---|---|
| round trip (what happens today for YAML) | 5.84M | 5.33 MB | 90,381 |
| `json.Unmarshal` + the `UnmarshalJSON` hooks | 3.02M | 2.76 MB | 48,401 |
| native `UnmarshalYAML` | **2.30M** | **2.32 MB** | **37,129** |

**2.54x faster than today, and 1.31x faster than the JSON fast path.** The hooks are the difference and a hook-free benchmark cannot see them: `Responses.UnmarshalJSON` re-marshals every child back to JSON and re-parses it, once per entry, and `SchemaRef.UnmarshalJSON` parses the same bytes up to four times (the ref, the extra keys, a sibling schema, the value). There are 16 nested `json.Marshal` calls across `refs.go`, `maplike.go` and `openapi3.go`. Handing the already-parsed child node to the child decoder skips all of it.

Options 1 and 2 keep `json.Unmarshal` and remove less, and I have not measured them.

That also answers the obvious suggestion of swapping the YAML library. I measured `goccy/go-yaml`, the performance-oriented alternative: 302M ns/op and 432 MB/op on the same input, slower than the round trip it would replace. `UseJSONUnmarshaler` did not change that. `go.yaml.in/yaml/v3` is within noise of the current parser.

Happy to share the harness.

## With IncludeOrigin

The same GitHub spec allocates 842 MB instead of 601 MB. The round trip is present either way; it is simply a smaller fraction of a larger number.

Origins make option 3 disproportionately attractive, though it is a general win and not an origin-only one. They currently ride *inside* the decoded data -- encoded into the node tree, decoded, then extracted again -- purely so they survive the JSON serialization. Under native decoding that cycle disappears, since each `UnmarshalYAML` is handed a node that already carries the position. Measured against the origin-enabled path, native decoding is 3.2x faster and 4.7x leaner.

## Relationship to #1235

Disjoint. #1235 targets the origin maps, which exist only when `IncludeOrigin` is on and which are **retained** for the document's lifetime; it halves retained size and moves peak by 1.6%. Everything above is **transient** allocation in the default configuration, garbage by the time the load returns, and #1235 does not touch it. Neither blocks the other.

## Reproducing

Load either spec above and compare `runtime.MemStats` plus peak RSS from `getrusage`; a heap profile with `-sample_index=alloc_space` gives the ranking. Happy to share the harness.

reuvenharrison:
Fixes #1236, which measured the cost and left the shape open. This is the patch for it.

## How to review this

The file count is misleading. Most of it is generated or is evidence, and the part that needs reading is four files.

| | files | changed lines | |
|---|---|---|---|
| hand-written logic | 14 | 1,435 | read |
| tests | 23 | 1,001 | skim |
| generated | 5 | 574 | skip, CI regenerates and diffs them |
| generator and template | 3 | 126 | read, they produce the generated files |
| testdata goldens | 44 | 87 | evidence, see below |

Four files are 80% of the logic, in the order I would read them:

1. `openapi3/native_yaml.go` (275) the shared decode helpers
2. `openapi3/native_yaml_special.go` (191) the hand-written types
3. `openapi3/origin.go` (520) positions, the largest single piece
4. `openapi3/end_positions.go` (163) block extents, described under "Two things worth reviewing closely"

The rest of the logic is small edits to `loader.go`, `marsh.go` and `schema.go`.

**The goldens are the argument, not the churn.** The suite runs 4,140 real-world specs from APIs-guru. 44 of their golden files move, by 87 lines in total, and almost every one is the same wording shift:

```
- error converting YAML to JSON: yaml: line 1647: found a tab character...
+ yaml: line 1647: found a tab character...
```

That is the JSON round trip disappearing from the error text. Three further specs now load rather than failing, because boolean schemas decode (#1257). Beyond that, 4,140 documents parse and validate to the same result as before, which is the evidence that replacing the decode path preserves behaviour.

**On the public API.** This adds about 40 exported `UnmarshalYAML` methods and one function, `UnmarshalFromData`. The methods are not new API to call: `yaml.Unmarshaler` is an interface, so a type opts in by having the method, and there is no way to implement it unexported. Nothing is removed and no existing signature changes.

## What this is

kin decodes a YAML document by converting it to `map[string]any`, re-serialising that to JSON text, and parsing it again, because these types implement `UnmarshalJSON` and not `UnmarshalYAML`. Positions cannot survive that trip, so they are smuggled through as synthetic `__origin__` nodes and reapplied afterwards by a reflection walk.

This replaces the round trip with `UnmarshalYAML` on every type, decoding straight from the node, against unmodified `go.yaml.in/yaml/v3`.

## Why it is worth doing

Loading a 1.7 MB spec (`netbox.dev/3.4`), this branch against master:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| master, origins off | 231.2M | 167.5 MB | 2,819,635 |
| **this branch, origins off** | **81.3M** | **64.9 MB** | **1,129,709** |
| master, origins on | 372.2M | 298.1 MB | 4,464,255 |
| **this branch, origins on** | **142.1M** | **115.9 MB** | **1,316,670** |

2.8x faster with origins off, 2.6x with them on, and roughly a third of the allocations in both cases. Reproduce with a benchmark that calls `LoadFromData` on any large spec. #1236 has the profile behind the allocation figures: about 62% of everything allocated during a load went on parsing JSON the decoder had generated moments earlier, which is exactly what this removes.

It also drops both YAML dependencies:

```
-	github.com/oasdiff/yaml v0.1.1
-	github.com/oasdiff/yaml3 v0.0.14
+	go.yaml.in/yaml/v3 v3.0.5
```

Those forks exist to carry the origin nodes and a `DisableTimestamps` option through the round trip. Neither is needed once the decode reads positions off the node directly, so kin goes back to a stock parser.

## Shape of the change

The types fall into four groups, which is what makes this tractable:

- **22 shadow-struct types**, generated from a template, following the pattern already used for the `$ref` wrappers. The JSON versions restate the known field set as a list of deletes (`Schema.UnmarshalJSON` is 91 lines, about 60 of them that list), which silently misfiles a field added to the struct and forgotten in the list. Here the field set comes off the struct tags and cannot drift.
- **9 `$ref` wrappers**, one helper plus thin generated methods.
- **3 maplike collections** (`Paths`, `Responses`, `Callback`), one generic helper.
- **Special cases**: `Header` defers to `Parameter`; `Types`, `BoolSchema` and `ExclusiveBound` are union-typed scalars; `T` and `Operation` need explicit handling for the first-key rule and for explicitly-null responses.

## Two things worth reviewing closely

**Block extents are derived, not parsed.** `Location.EndLine`/`EndColumn` cannot come from go-yaml, which reports where a node starts and nothing about where it stops.

`EndLine` comes from the tree, in two passes. The first takes the largest line among a node's descendants; deriving that from the tree rather than from indentation is what makes a sequence item work, since a parameter's key location is the item's first key, sitting at the same column as the keys that follow it, so no column comparison can separate the end of the item from the start of its own second field. The second pass grows that over lines belonging to the block that carry no node, which is the body of a block scalar: `description: |` is a single scalar whose `Line` is where it starts, so a block ending in one would otherwise stop there.

Checked against a parser that records ends natively ([yaml/go-yaml#391](https://github.com/yaml/go-yaml/pull/391)), the two passes agree on **99.98%** of 7.18M blocks across a 4,240-file corpus; the first pass alone agrees on 99.60%, the difference being exactly the trailing-block-scalar case. The residual is a block reported a line or two short where indentation is genuinely ambiguous.

`EndColumn` does not come from the tree at all. It is the length of that last line plus one, read from the source text. This is the weaker half of the pair: it marks the end of the **line**, not the end of the node, so a trailing comment or trailing whitespace is included. A block whose last line is `type: object   # note` reports `EndColumn` past the comment rather than past `object`. For slicing a block out of its source, which is what these fields are for, the line is what matters and the column rarely is. Worth knowing before relying on it for anything narrower.

**YAML 1.1 implicit resolution.** The old path normalised every scalar through JSON's type system on the way through. Decoding natively keeps YAML 1.1 semantics, so implicit resolution applies where it did not before. Three cases are handled explicitly: date-shaped scalars are retagged before decode (reproducing what `DisableTimestamps` did), integer values reaching `any`-typed fields are normalised to `float64` so a consumer's type switch does not depend on whether the source wrote `42`, `0x2A` or `4_2`, and non-string mapping keys are retagged as strings. The last one is a behaviour improvement rather than a fix: the old path mangled a key like `18_24` into `1824`, and it now loads as written.

## Status

`go test ./...` is green.

Two behaviours changed deliberately:

- A type error now names the line instead of the struct field, since nothing renders the document as JSON first: `line 24: cannot unmarshal !!map into openapi3.SchemaRefs` rather than `json: cannot unmarshal object into field Schema.allOf of type openapi3.SchemaRefs`. The line number is more useful to a spec author, but the field path is genuinely lost. Worth a second opinion.
- Duplicate keys. JSON resolves a repeated key last-one-wins, YAML rejects it. The json path is kept as a fallback so such documents still load, without origins.

`openapi2` is migrated too, but minimally: it keeps its JSON round trip and every `UnmarshalJSON` untouched, and only the YAML front half changes, which is what lets the dependencies go. Migrating it to native decoding as well is separate work.

**Concurrent decodes now serialise, and this is the part I would most like a view on.** Three package-level variables carry per-decode state (the file, the origins-enabled flag, and the end index), because `UnmarshalYAML` receives a node and nothing else, with no way to pass context through the call. `TestIssue741` loads concurrently with a separate `Loader` per goroutine, and the race detector caught all three, so a mutex is held for the length of a decode. Single-decode throughput is unaffected, which is what the table above measures, but a caller parsing several documents at once no longer does so in parallel. The path being replaced passed the file as an argument, so this is a regression on that axis and the only one in the change.

Three ways out, none taken here:

- **A context on the decode**, which go-yaml does not have yet, but which is asked for in [yaml/go-yaml#187](https://github.com/yaml/go-yaml/issues/187). This is the one that would fix it properly: an `UnmarshalYAML` receiving the context its decode started with retires all three variables at once, and the lock with them. That work is now [yaml/go-yaml#392](https://github.com/yaml/go-yaml/pull/392) by @ccoVeille, which replaced the version I had open as #390. The receiving interface is the same shape either way, `UnmarshalYAML(ctx context.Context, node *Node) error`, so the conversion I have working locally carries over: all three variables and the mutex deleted, suite green, race clean. Only the entry points differ, `LoadContext` / `DecodeContext` / `UnmarshalContext` rather than the `WithContext` spelling. It is not proposed here because of timing, not doubt. #392 targets `main`, which is `v4` and still at `rc`, and the maintainers have said the released `v3` line takes no runtime changes, so kin cannot use it until v4 ships.
- **A file recorded on the node.** The callback already has the node, so it could read the file from it. That wants a field on go-yaml's `Node`, which does not exist, and it is the expensive shape: an `any` is 16 bytes on a 168-byte `Node`, and a 1.8 MB document parses to 145,671 nodes, so the field would cost more than the document does, paid by every user, to carry a value identical for the whole tree.
- **Filling the file in after the decode**, by walking the document. That removes one variable but not the other two, so the lock would still be needed, and it trades a lock for a reflection pass over every loaded document.

Happy to take direction on which is worth pursuing, or to leave the lock and revisit when the upstream question settles.

## Validation

- Every full document in `testdata` decodes identically to the JSON path.
- Origins were compared against the previous implementation on `Key.Line`/`Column`, `Fields` and `Sequences` across the same documents.
- The apis-guru corpus is what surfaced the timestamp problem, and is the reason the YAML 1.1 section above is stated rather than assumed absent.

Bugs the suite caught along the way, each of which would otherwise have shipped: origins were produced when the caller had not asked for them (the gate read the package global rather than the flag the `Loader` passes, and since `NewLoader` seeds itself from that global, test pollution made it look correct); a parameter inside a sequence had no `Key`, because stamping walked mappings only; an alias reported the line it sat on rather than the anchor it points at; and a `$ref` resolved through `T.Extensions` stamped the referencing document's filename instead of the file the schema lives in.


openapi3/native_yaml_refs.go
```diff
@@ -0,0 +1,45 @@
...
+			}
+		case strings.HasPrefix(k, "x-"):
+			var a any
+			if v.Decode(&a) != nil {
+				continue
+			}
+			if *extensions == nil {
+				*extensions = make(map[string]any)
+			}
+			(*extensions)[k] = a
```
snowyukitty: **Blocking:** every non-`x-` sibling not handled by this switch is silently dropped, so the wrapper's `extra` slice stays empty. A `ResponseRef` with `$ref` plus `bogus: true` fails validation on the base but validates on this head, for both JSON and YAML input. Could this helper record unknown sibling names the same way `UnmarshalJSON` does, with a focused test for both input formats?

openapi3/native_yaml.go
```diff
@@ -0,0 +1,174 @@
...
+// tag is a deliberate request for a time.Time and is left alone.
+func stripTimestamps(n *yaml.Node) {
+	if n == nil {
+		return
+	}
+	if n.Kind == yaml.ScalarNode && n.Tag == "!!timestamp" && n.Style != yaml.TaggedStyle {
+		n.Tag = "!!str"
+	}
+	for _, c := range n.Content {
+		stripTimestamps(c)
```
snowyukitty, line 273: A nested unquoted numeric key in an extension now survives as `map[any]any`, so a successfully loaded document later fails `json.Marshal` with `unsupported type: map[interface {}]interface {}`. The base remains marshalable. Could the OpenAPI 3 path apply the same non-string-key reconciliation used by `internal/yamlconv`, and add a load-then-marshal regression test?

openapi3/native_yaml.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, err
+	}
+	if node.Kind != yaml.MappingNode {
+		return nil, nil
+	}
+	known := knownYAMLFields(reflect.TypeOf(out).Elem())
+
+	var ext map[string]any
+	for i := 0; i+1 < len(node.Content); i += 2 {
+		key := node.Content[i].Value
```
snowyukitty: The decoder has already applied a YAML merge here, but this raw-key pass also stores `<<` as an extension. A previously accepted merge-anchor `Info` object then fails `Validate` with `extra sibling fields: [<<]`. Please skip merge-key syntax during extension collection and cover the public load + validate path.

openapi3/native_yaml_special.go
```diff
@@ -0,0 +1,166 @@
...
+	*ext = make(map[string]any)
+	*out = make(map[string]*V, len(node.Content)/2)
+	for i := 0; i+1 < len(node.Content); i += 2 {
+		k, v := node.Content[i].Value, node.Content[i+1]
+		if strings.HasPrefix(k, "x-") {
+			var a any
+			if err := v.Decode(&a); err != nil {
+				return err
+			}
+			(*ext)[k] = a
```
snowyukitty, line 29: These maplike and ref helper paths bypass `normalizeNumbers`: numeric extensions on a `Responses` collection and a `ResponseRef` change from `float64` to `int`, for JSON as well as YAML because both now use `UnmarshalYAML`. Applying the same normalization as ordinary struct extensions would keep the public value shape consistent.

snowyukitty (changes requested):
Thanks for the careful native-decoding work. I tested the exact base `5d526a9` and head `754b77a` through `NewLoader().LoadFromData`. The same four focused tests pass on the base and fail on the head. I think the first issue is blocking because it makes invalid Reference Objects validate successfully; the other three are compatibility gaps worth pinning before this leaves draft. Existing CI is green, but its native-scalar and extra-sibling tests do not cover these helper paths.

reuvenharrison:
@snowyukitty thank you, all four reproduce. I rebuilt each one myself against `5d526a9` and `754b77a` rather than blindly trusting them, and every one passes on the base and fails on the head, including the `$ref` sibling case for JSON input as well as YAML.

Following the thread further, these are not four separate bugs. They are four symptoms of one cause, and there are more.

## The root cause

Decoding through JSON applied four normalizations in one place, to every object: non-string map keys became strings, integers became `float64`, merge keys were applied, and unknown siblings were recorded.

Decoding natively splits that work across three functions, `decodeStructWithExtensions`, `unmarshalMaplikeYAML` and `unmarshalRefYAML`. Each has to apply all four itself, and only `decodeStructWithExtensions` handles most of them. Every issue you found is one normalization missing from one of those functions.

So I went looking for the rest, and found two more.

**A merge key in a keyed collection.** The merged entry is lost, and validation then complains about something else:

```yaml
x-tmpl: &d
  "200": {description: ok}
paths:
  /a:
    get:
      responses:
        <<: *d
```

```
base: loads and validates
head: invalid paths: invalid path /a: invalid operation GET: a short description of the response is required
```

**A nested numeric key in a `$ref` extension.** The same `map[interface {}]interface {}` marshal failure as your second issue, this time through `unmarshalRefYAML` instead of `decodeStructWithExtensions`.

## What I am changing

Fixing the four places you named would leave both of those, and would let the next decode function, or the next normalization, miss one again without anything failing. So instead:

- one helper that all three functions call to collect undeclared keys, which skips merge keys, normalizes numbers and converts non-string keys, so none of them can leave one out
- `unmarshalRefYAML` stops using a `switch` whose default branch silently drops siblings, which is the blocking bug
- a table test covering each decode function against each normalization, instead of six one-off regression tests

Your last point is the one I would underline: CI was green across a 1979-line rewrite of the decode path with six real regressions in it. The missing tests were the finding.

Fix and tests to follow on this branch.

reuvenharrison:
Pushed as `a27deb9`. All four are fixed, plus three more that turned up once I looked for the same mistake everywhere instead of patching the four places.

## The three nobody had reported

- **A merge key in a keyed collection** dropped the merged entry. `responses: {<<: *d}` lost the response, then failed validation with `a short description of the response is required`, which points at the wrong thing.
- **A numeric key in a `$ref` extension** left an unmarshalable `map[any]any`, the same failure as your second issue but through `unmarshalRefYAML`.
- **A `$ref` reached through a merge key** was lost completely. `mappingValue` looked the key up without expanding merges, found nothing, and the wrapper decoded itself as a value instead. The table test caught this one after the first fix was already in.

## What changed

Three helpers that all three decode functions now share:

- `normalizeDecoded` replaces `normalizeNumbers`. It widens integers and also converts a non-string map key to a string. YAML allows any scalar as a key and `json.Marshal` rejects `map[any]any`, so a document could load and then fail to marshal.
- `mappingPairs` **applies** merge keys instead of skipping them. Only `Decode` expands a merge, and these passes read `node.Content` directly, so `<<` arrived unexpanded. Ignoring it dropped the merged content; treating it as a normal key recorded `<<` as an extension or as a member. `mappingEntry` uses it too, which is what fixes the merged `$ref`.
- `collectExtensions` returns the undeclared keys, decoded through the two above, plus their names.

`unmarshalRefYAML` no longer uses a `switch`. Its default branch dropped every sibling it did not name, so `extra` stayed empty and `Validate` had nothing to report. It now records every undeclared sibling and keeps the `x-` ones as extensions, matching `UnmarshalJSON`. `SchemaRef` passes nil for summary and description, so those count as siblings there, exactly as the JSON path treats them.

## The test

One table covering each decode function against each normalization:

| | records siblings | widens integers | converts keys | applies merges |
|---|---|---|---|---|
| `decodeStructWithExtensions` | yes | yes | yes | yes |
| `unmarshalMaplikeYAML` | n/a | yes | yes | yes |
| `unmarshalRefYAML` | yes | yes | yes | yes |

`unmarshalMaplikeYAML` has no sibling case because every non-`x-` key of a keyed collection is a member of it. That exemption is recorded with its reason, and the test fails on any combination that is neither tested nor exempted, so adding a decode function or a normalization has to come with its tests.

**The whole table also passes on `5d526a9`.** It pins what decoding through JSON already guaranteed, rather than inventing new requirements.

Full suite green. Thanks again, the four you found were the thread worth pulling.

reuvenharrison:
Pushed `c0f57a2`: boolean schemas are decoded on the native path.

`Schema` was in the generator's list, so its `UnmarshalYAML` went straight to `decodeMapping`, which needs a mapping node and could not take the `true` or `false` a JSON Schema 2020-12 document may write wherever a schema is expected. It moves to `native_yaml_special.go` alongside the other union-typed values and takes a `!!bool` scalar before falling through to the same `decodeMapping` body the generator emitted. Since this branch routes JSON through the same decoder, both forms are covered.

The commit before it merges #1258, which adds the `Boolean` field this needs. That merge collapses once #1258 lands on master.

openapi3/native_yaml_refs.go
reuvenharrison: Fixed in `a27deb9`. The switch no longer decides what gets recorded: `collectExtensions(node, known)` returns every undeclared sibling and `*extra = names` keeps all of them, with only the `x-` ones retained as extensions. That matches what `UnmarshalJSON` does, so `Validate` can report an unknown key alongside a `$ref` again. `SchemaRef` passes nil for summary and description, so those count as siblings there, exactly as on the JSON path.

openapi3/native_yaml.go
reuvenharrison: Fixed in `a27deb9`. `normalizeDecoded` (which replaced `normalizeNumbers`) has a `map[any]any` case that converts each key with `fmt.Sprintf("%v", k)` and recurses, so a nested unquoted numeric key becomes a string key and the document marshals. It is reached from all three decode paths through `decodeAny`, so the fix is not specific to the site you flagged.

openapi3/native_yaml.go
reuvenharrison: Fixed in `a27deb9`. `mappingPairs` applies the merge rather than letting the raw-key pass see it: a `<<` key is expanded into its source pairs and never emitted as a pair of its own, so it cannot land in extensions. Explicit keys win over merged ones and earlier merged sources win, per the merge key spec. Skipping `<<` instead would have dropped the merged content, which was a second bug in the same place.

openapi3/native_yaml_special.go
reuvenharrison: Fixed in `a27deb9`, though not by adding a call at this site. `normalizeNumbers` became `normalizeDecoded`, and every path that puts a decoded value behind an `any` now goes through `decodeAny`, which applies it: `unmarshalMaplikeYAML` calls `decodeAny` for its `x-` keys, while `unmarshalRefYAML` and `decodeStructWithExtensions` both reach it via `collectExtensions`. So numeric extensions on a `Responses` collection and on a `ResponseRef` keep the same `float64` shape as ordinary struct extensions. The table test pins each decode function against each normalization and fails on any untested combination.

PR #1239, opened 2026-08-04
openapi3: parse openapi versions numerically, reject unsupported ones

mromaszewicz:
The version accessors were exact-string whitelists: any version off the list (3.0.5, 3.1.3, 3.2.1, 3.3.0, ...) made OpenAPIMajorMinor return "" and every Is* predicate return false, so Validate silently applied stricter-than-3.0 rules with no diagnostic.

- Parse MAJOR[.MINOR[.PATCH[-pre][+build]]] numerically; the OrLater predicates are now real ordering comparisons, so future patch and minor releases (3.2.1, 3.3.0, 3.10.0) classify correctly.
- Validate reports a new openapi-version-unsupported error (OpenAPIVersionUnsupportedError) for unparseable versions and for major versions other than 3, instead of misvalidating silently. Parseable 3.x versions are accepted forward-compatibly.
- OpenAPIMajorMinor now reports any well-formed version, including non-3.x ones (e.g. "4.0.0" yields "4.0" instead of "").


fenollp (changes requested):
Thanks for improving on this. However I do not wish to merge the part that is parsing versions as semvers. This is much more complex than it needs to be in my opinion. Also you implement more than the spec asks for: https://spec.openapis.org/oas/v3.2.0.html#versions-and-deprecation (it's not fully semver). Please revert that part. I'm fine with adding new version strings as they come: not every day. If you feel differently please explain. 

Also this needs rebasing.

mromaszewicz:
I was messing around with it, and found the string based comparison to be fragile. I always thought this version was like semver! I'll simplify.

mromaszewicz:
Ready for another round. Your call. I think parsing the version numerically allows for better version comparisons, but this isn't a big deal. You are welcome to close this PR if you don't like it, it's your code :) 

I did this because people might start using newer specs, whose version isn't in the whitelist, and things would break. It's less work in the future since you don't need to update the whitelist, but also, OpenAPI version bumps don't happen that often.

fenollp (changes requested):
Let's remove the parsing. Per the spec only the major and minor parts are to be used by library authors. A regexp matching `^3[.][012]([.][0-9]+)?$` should be enough to output the error you introduced. The only comparison API we need is already here with the OpenAPI3orlater methods.

PR #1240, opened 2026-08-04
openapi3: support OAS 3.2 query and additionalOperations

mromaszewicz:
Add the OpenAPI 3.2 Path Item Object fields for custom HTTP methods: the fixed query field (HTTP QUERY) and additionalOperations, a map of custom method names to operations.

- PathItem gains Query and AdditionalOperations; Operations(), GetOperation() and SetOperation() cover them. SetOperation no longer panics on unknown methods; they are stored in AdditionalOperations.
- Both fields are version-gated: validating a pre-3.2 document that uses them fails with query-field-for-3-2-plus / additional-operations-field-for-3-2-plus. AdditionalOperations keys must be RFC 9110 tokens and must not duplicate a fixed-field method.
- Schema/parameter walkers address custom methods with /additionalOperations/<METHOD> JSON pointers, preserving case.
- openapi2conv.FromV3 and FromV3PathItem return an error instead of panicking for methods Swagger 2.0 cannot express.
- routers/legacy: guard a nil-node dereference in the FindRoute fallback when the document declares an operation absent from the route trie.
- Routers and openapi3filter need no changes: both derive method sets from Operations(), proven by new routing and request-validation tests for QUERY and custom methods.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -18,6 +18,8 @@ OpenAPI 3.1 Features:
 
 OpenAPI 3.2 Features:
   - Media Type Object itemSchema for streaming sequential media types
+  - Path Item Object query field for the HTTP QUERY method
+  - Path Item Object additionalOperations for custom HTTP methods
 
 The implementation maintains 100% backward compatibility with OpenAPI 3.0.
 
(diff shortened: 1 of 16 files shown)


fenollp (approved):

fenollp:
Thanks!

PR #1241, opened 2026-08-05
openapi3: export PathItemMethods

reuvenharrison:
Follow-up to #1240.

`fixedPathItemMethods` records which HTTP methods have a dedicated Path Item Object field. `Operations()`, `GetOperation()` and `SetOperation()` all consult it, but it is unexported, so a caller outside the package that needs the same distinction has to write the list out again.

A list written out again is one that silently misses the next method the spec adds, and #1240 has just made that concrete. OpenAPI 3.2 promoted QUERY to a fixed field, so a consumer carrying its own nine-method list now classifies QUERY as a custom method: it reads `additionalOperations` for it, finds nothing, and treats an operation the document does declare as absent. The same will happen to whatever 3.3 adds.

`PathItemMethods()` returns the set, sorted, so a caller can ask instead of guess:

```go
for method, operation := range pathItem.Operations() {
    if slices.Contains(openapi3.PathItemMethods(), method) {
        // fixed field
    } else {
        // custom method, from additionalOperations
    }
}
```

The test pins the export against the behaviour it describes rather than against a copy of the list: every method it returns must round-trip through `SetOperation`/`GetOperation` leaving `AdditionalOperations` empty, and a method it omits must land in that map. So the two cannot drift.

Sorted rather than in spec order, matching `Operations()` and `operationEntries()`, which already order methods with `componentNames`.

### DIFF
--- a/.github/docs/openapi3.txt
+++ b/.github/docs/openapi3.txt
@@ -203,6 +203,12 @@ func Int64Ptr(value int64) *int64
 
     Deprecated: Use Ptr instead.
 
+func PathItemMethods() []string
+    PathItemMethods returns the HTTP methods that have a dedicated Path Item
+    Object field, sorted. Any other method an operation is registered under
+    lives in AdditionalOperations, so a caller that treats the two groups
+    differently can tell them apart without listing the fixed methods itself.
+
 func Ptr[T any](value T) *T
     Ptr is a helper for defining OpenAPI schemas.
 
--- a/openapi3/path_item.go
+++ b/openapi3/path_item.go
@@ -56,6 +56,14 @@ var fixedPathItemMethods = map[string]struct{}{
 	MethodQuery:        {},
 }
 
+// PathItemMethods returns the HTTP methods that have a dedicated Path Item
+// Object field, sorted. Any other method an operation is registered under
+// lives in AdditionalOperations, so a caller that treats the two groups
+// differently can tell them apart without listing the fixed methods itself.
+func PathItemMethods() []string {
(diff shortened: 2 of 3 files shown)


reuvenharrison:
The Windows job failed on the first run of this branch and is green on the re-run, along with macOS, Ubuntu, goimports and shellcheck. Recording what it was, in case it turns up on someone else's PR.

The failure was in `TestIssue883/Roundtrip_using_yaml_pkg`, which this PR does not touch:

```
--- FAIL: TestIssue883/Roundtrip_using_yaml_pkg
panic: runtime error: index out of range [2951872096] with length 28
	regexp.(*Regexp).doOnePass       regexp/exec.go:445
	gopkg.in/yaml.v3.resolve         yaml.v3@v3.0.1/resolve.go:204
	gopkg.in/yaml.v3.(*parser).scalar  yaml.v3@v3.0.1/decode.go:247
```

An index of ~2.9e9 inside the regexp one-pass engine reads as memory corruption rather than a logic error. The same shape has been reported intermittently elsewhere (gorilla/mux#742, where a long-working route panicked occasionally), so it looks like a flake rather than something in this branch.

What I checked before concluding that:

- the diff is one exported function and its test, and `componentNames` copies the map's keys into a fresh slice, so nothing here mutates shared state or aliases the package variable;
- `TestIssue883` passes 20 consecutive runs locally under `-race`;
- the branch is one commit ahead of master with nothing behind it, so it is not a stale base.

One sighting is not enough to open an issue on. If it shows up again on Windows, it is probably worth reporting against the toolchain rather than here.

PR #1249, opened 2026-08-17
openapi3filter: handle deserializing deepObject param with additionalProperties into maps

### LINKED ISSUE #1112: query params deepObject with additionalProperties: true
I have this schema:
```
- name: properties
  in: query
  description: Filter animals by properties
  required: false
  style: deepObject
  schema:
    type: object
    additionalProperties: true
```

it generates this type:
```
Properties *map[string]interface{} `json:"properties,omitempty"`
```

Which is good and basically what i want, but the request validation fails for this url: 
```animals?properties[vaccinated]=true&properties[color]=black&properties[coat_length]=large```
i get error: Invalid format for parameter properties: error assigning value to destination: error binding map: unhandled type: interface {}

If i say: 
```
additionalProperties:
  type: string 
```
Then it works but the type is `*map[string]string` and i would manually have to go through jsonschema and loop over properties and convert types from string to bool or whatever. Is there a way to make it work with basically any object, or somehow skip validation for only that specific param?

Berserk-hub150:
## Summary

Fixes `deepObject` query decoding for objects with explicit free-form `additionalProperties: true`.

- recognizes dynamic-only deepObject parameters as present
- retains values allowed by explicit `additionalProperties: true`
- keeps schema-valued additional properties typed
- preserves the existing behavior for `additionalProperties: false` and an unspecified keyword
- preserves the existing nested schema-valued decoding behavior
- leaves form/path/header/cookie object decoding unchanged
- adds regression coverage for the reported case and relevant edge cases

## Validation

- `go test ./openapi3filter -run '^TestIssue1112' -count=1`
- `go test ./openapi3filter -count=1`
- `go vet ./openapi3filter`
- `git diff --check`
- exactly 17 Signed-off-by commits

Fixes #1112

### DIFF
--- a/openapi3filter/issue1112_test.go
+++ b/openapi3filter/issue1112_test.go
@@ -113,3 +113,13 @@ func TestIssue1112KnownPropertyKeepsTypedValue(t *testing.T) {
 		"color": "black",
 	}, value)
 }
+
+func TestIssue1112AdditionalPropertiesFalseDoesNotBecomeFreeForm(t *testing.T) {
+	schema := issue1112ObjectSchema(openapi3.AdditionalProperties{Has: openapi3.Ptr(false)})
+
+	value, found, err := issue1112Decode(t, "properties", "properties[color]=black", schema)
+
+	require.NoError(t, err)
+	require.False(t, found)
+	require.Equal(t, map[string]any{}, value)
+}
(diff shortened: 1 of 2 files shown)


fenollp (changes requested):
LGTM. Please address the CI failure and we're go.

PR #1252, opened 2026-08-19
openapi3: compare multipleOf on exact rationals, not fixed-precision decimals

### LINKED ISSUE #1109: MulipleOf Validation is failing for numbers 999999999.99, 999999999.90
Referring to PR https://github.com/getkin/kin-openapi/pull/1068/changes

I encountered and issue in Test-Case `TestIssue817` of `schema_test.go`. If you add the value `999999999.99` to the `validData` array, the test fails.
We have encountered this as we were testing the borders or our API. In Version `0.1.31.0` we did not have the issue.

Proposal: Instead of using precision `"%.10f"` for every use case, we could determine the precision dynamically based on the `multipleOf` value

snowyukitty:
Fixes #1109.

### What breaks

`multipleOf` formats both operands with `"%.10f"` before parsing them into
`big.Rat`. That fixed precision is deliberate — #1068 introduced it to absorb
float64 representation error, and #1194 carried it across when `decimal128` was
removed — but the tolerance is *absolute*, so it is only correct inside a narrow
band of magnitudes.

**Above roughly `1e6`**, ten fractional digits behind a large integer part demand
more significant digits than a float64 carries, so the format prints the binary
error rather than the value. `999999999.99` becomes `"999999999.9900000095"` and
is rejected, even though the reporter's schema declares that same number as its
`maximum`:

```go
schema := &Schema{Type: &Types{"number"}, MultipleOf: Ptr(0.01),
    Min: Ptr(-999999999.99), Max: Ptr(999999999.99)}
schema.VisitJSON(999999999.99)  // error: number must be a multiple of 0.01
```

**Below roughly `5e-11`**, every operand formats to `"0.0000000000"`. A value
then becomes a multiple of anything, and a small `multipleOf` collapses the
denominator to zero, so `big.Rat.Quo` panics. Validating a document that merely
declares a small `multipleOf` takes the process down:

```go
var s Schema
json.Unmarshal([]byte(`{"type":"number","multipleOf":1e-12}`), &s)
s.Validate(context.Background())  // no error: nothing rejects this schema
s.VisitJSON(5.0)                  // panic: division by zero
```

`multipleOf: 0` panics the same way, and a negative `multipleOf` accepted every
number. `Schema.Validate` has no check that would stop such a document from
loading, so this is reachable from an ordinary specification.

### The change

Compare exact rationals instead of formatted decimals. `big.Rat.SetFloat64` is
exact, so there is no formatting step, no parse that can fail silently, and no
denominator that can collapse to zero.

The tolerance becomes relative to the value: accept when the quotient is an
integer, or when the nearest candidate multiple sits within `|value| * 2^-51` of
the value — a couple of units in the last place.

Measuring against the value rather than against the quotient matters. A
quotient-relative window grows without bound and would pass one half around
`2^49`, after which every number validates; `2^50` against `multipleOf: 3` is a
concrete case, and there is a regression test for it. The candidate multiples are
derived through `big.Int`, so no quotient can overflow.

### Behaviour changes

| input | before | after |
| --- | --- | --- |
| `999999999.99` with `multipleOf: 0.01` | rejected | accepted |
| `multipleOf` below the old decimal grid, e.g. `1e-12`, incl. subnormals | panic | validated |
| `multipleOf: 0` | panic | `multipleOf must be greater than zero` |
| negative `multipleOf` | accepted everything | `multipleOf must be greater than zero` |
| `1e-300` with `multipleOf: 0.01` | accepted (treated as zero) | rejected |
| `1.00000000004` with `multipleOf: 1` | accepted | rejected |

The last row is the one worth flagging: the window is now tied to float64
precision rather than to ten decimal places, so a value that differs from a
multiple by more than a few ULP is rejected where the old grid rounded it in.
Computed floats such as `0.1+0.2` against `multipleOf: 0.1` are unaffected.

One limit is inherent rather than chosen. When `multipleOf` is smaller than the
value's own ULP — `1e16` against `multipleOf: 3`, where consecutive float64
values are 2 apart — the question is not decidable from the stored number, and
the check accepts. No relative tolerance can both absorb one ULP at `0.3` and
reject a one-unit gap at `1e16`.

Two cases are deliberately left alone, with comments saying why: a non-finite
`multipleOf` is treated as no constraint, and a non-finite value keeps its
current outcome through the exported `VisitJSONNumber`. `SchemaError.Error()`
marshals both the value and the schema as JSON, and NaN has no JSON encoding, so
reporting either would panic while *formatting the error* — a worse outcome than
what is there now. Making `SchemaError` renderable for non-finite input is a
separate change. Note that `VisitJSON` already rejects non-finite input before
any keyword runs; only the exported number entry point reaches this code.

### Testing

`TestIssue817` gains the values from the report. A new
`TestSchemaMultipleOfAcrossMagnitudes` covers the invariant by scenario: large
magnitudes, preserved tolerance for computed floats, a `multipleOf` smaller than
the value's precision including a subnormal, a value far below the `multipleOf`,
zero, non-finite input, `multipleOf` bounds, a no-panic case for every previously
crashing input, and an assertion that the failure still names `multipleOf` in
`SchemaError.SchemaField`.

The computed-float cases go through a `//go:noinline` helper on purpose. Written
as a literal, `0.1+0.2` is folded exactly at compile time and the case would
assert nothing.

`go test ./openapi3/`, `go vet ./...`, and `gofmt -l .` are clean.
`go test ./...` is clean except `openapi2conv/TestIssue440`, which fails
identically on an unpatched checkout of the same base
(`invalid character '.' looking for beginning of value`) and is unrelated.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -2361,6 +2361,80 @@ func (schema *Schema) visitJSONBoolean(settings *schemaValidationSettings, value
 	return
 }
 
+// multipleOfSlack bounds, as a negative power of two, how far a value may sit
+// from the nearest multiple and still count as one, measured relative to the
+// value itself. A float64 carries about 2^-53 of relative error, so this is a
+// couple of units in the last place: enough to absorb the error a caller's own
+// arithmetic introduced, small enough that a genuine remainder never fits.
+//
+// The window must be measured against the value and not against the quotient.
+// A quotient-relative window grows without bound, and once value/multipleOf
+// reaches 2^49 it exceeds one half, at which point every number validates.
+const multipleOfSlack = 51
+
+// multipleOfFailure reports whether value satisfies multipleOf, and the reason
+// when it does not.
+//
+// The comparison runs on exact rationals rather than on decimals formatted at a
+// fixed precision. A fixed precision is only correct inside a narrow magnitude
+// band: above it the formatting emits binary representation error and rejects
+// valid values, and below it every operand formats to zero, which both accepts
+// any value and divides by zero.
+func multipleOfFailure(value, multipleOf float64) (string, bool) {
+	// A non-finite multipleOf cannot be reported: SchemaError renders the whole
+	// schema as JSON and NaN has no JSON encoding, so building that error would
+	// panic instead of describing the problem. Treat it as no constraint, which
+	// at least replaces the division-by-zero panic the fixed-precision
+	// formatting produced here.
+	if math.IsNaN(multipleOf) || math.IsInf(multipleOf, 0) {
+		return "", true
+	}
+	// JSON Schema requires multipleOf to be strictly greater than zero. Without
+	// this guard a zero or negative value either divided by zero or accepted
+	// every number.
+	if multipleOf <= 0 {
+		return fmt.Sprintf("multipleOf must be greater than zero, got %g", multipleOf), false
+	}
+	// VisitJSON rejects a non-finite number before any keyword runs, but the
+	// exported VisitJSONNumber does not, so one can still reach this point.
+	// Reporting it here is out of scope: SchemaError marshals Value, and NaN has
+	// no JSON encoding, so the error would panic when rendered. Preserve the
+	// existing outcome and leave that entry point to a separate change.
+	if math.IsNaN(value) || math.IsInf(value, 0) {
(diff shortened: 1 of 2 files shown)


fenollp:
Thank you very much!

PR #1258, opened 2026-08-25
openapi3: accept boolean schemas

### LINKED ISSUE #1257: Boolean schemas fail to unmarshal for items, prefixItems and other schema-valued keywords
In JSON Schema 2020-12, which OpenAPI 3.1 adopts, a boolean is a valid schema anywhere a schema is expected. `true` accepts everything and `false` accepts nothing. `BoolSchema` already models this, but only `unevaluatedItems` and `unevaluatedProperties` use it (plus `additionalProperties` through the alias). The other schema-valued keywords are typed `*SchemaRef` or `SchemaRefs`, so a boolean there fails to load the whole document.

## Reproducer

```go
const spec = `
openapi: 3.1.0
info: {title: t, version: "1"}
components:
  schemas:
    Pair:
      type: array
      prefixItems:
        - type: string
        - type: integer
      items: false
paths: {}
`
_, err := openapi3.NewLoader().LoadFromData([]byte(spec))
// failed to unmarshal data: json error: invalid character 'o' looking for beginning of value,
// yaml error: error unmarshaling JSON: while decoding JSON:
// json: cannot unmarshal bool into field Schema.items of type openapi3.Schema
```

Same failure for each of these:

| document | result |
|---|---|
| `items: false` | load error |
| `items: true` | load error |
| `prefixItems: [true, {type: string}]` | load error |
| `not: true` | load error |

Tested on v0.146.0; `Schema.Items` is still `*SchemaRef` on v0.147.0.

## Why `items: false` in particular

`prefixItems` alone leaves an array open at the end: positions beyond the listed entries are unconstrained. Closing it to a fixed-length tuple is done with `items: false`, which is the idiom the JSON Schema documentation gives for tuples. So a document cannot currently describe a strict tuple at all, and the failure is a load error rather than a field being ignored, which takes the whole document with it.

## Suggestion

Widen the schema-valued keywords to `BoolSchema`, the type already present for exactly this. `items` is the one with a concrete idiom behind it; `prefixItems`, `not`, `contains`, `if`/`then`/`else` and the `properties` values have the same latitude in the specification.

That is a breaking change to the field types, so it may want to go with other 3.1 work rather than alone. A narrower first step would be `items` only, since that is where a boolean is idiomatic rather than merely legal.

reuvenharrison:
Fixes #1257.

JSON Schema 2020-12, which OpenAPI 3.1 adopts, allows a boolean wherever a schema is expected. `true` accepts every instance, `false` accepts none. Every schema-valued keyword is typed `*SchemaRef` or `SchemaRefs`, so a boolean in one of those positions fails to unmarshal and takes the whole document with it:

```
json: cannot unmarshal bool into field Schema.items of type openapi3.Schema
```

## Approach

`SchemaRef.UnmarshalJSON` ends in `json.Unmarshal(data, &x.Value)`, so every schema-valued position funnels through `Schema.UnmarshalJSON`. Handling a bare boolean there covers `items`, `prefixItems`, `not`, `contains`, `if` / `then` / `else`, the values of `properties`, `$defs` and the rest in one place, rather than widening each field individually.

The result is recorded in a new field:

```go
// Boolean is set when the schema was written as a JSON Schema boolean
// schema rather than an object: `true` accepts every instance, `false`
// accepts none. It is nil for an ordinary object schema, and Validate
// rejects one that carries any other keyword.
//
// Neither direction goes through a struct tag, hence the "-" on both:
// UnmarshalJSON takes the bare boolean before the field decode, and
// MarshalYAML emits it again ahead of the field map. OpenAPI >=3.1.
Boolean *bool `json:"-" yaml:"-"`
```

Around it:

- `MarshalYAML` emits the bare boolean, so a document round-trips to what it was written as rather than to the object it is equivalent to.
- `visitJSON` accepts `true` and rejects `false` explicitly, ahead of the `IsEmpty` short circuit that would otherwise let `false` accept everything.
- `IsEmpty` reports neither boolean as empty, and consults the field last so a schema carrying both a boolean and other keywords is judged by the keywords. Reporting `true` as empty would make `visitJSON` skip validation, which is how `not: true` would come to accept every value when it should accept none.
- `Validate` rejects a `Boolean` set alongside any other keyword or extension, under a new `boolean-schema-with-other-keywords` code in the `MutuallyExclusiveFieldsError` cluster. Such a schema would marshal back as the bare boolean and drop the rest, so it is refused rather than left to lose content on the next serialization.

This is additive. `Boolean` is nil for an ordinary schema and no existing field changes type, so callers reading `schema.Items.Value` are unaffected. Choosing this over widening the fields to `BoolSchema` keeps the change out of the many call sites that use `Items` and friends today.

## Corpus

Three documents in the APIs-guru corpus were failing to load for exactly this reason. Their goldens move from `__load` to `__validate`, where they now report unrelated pre-existing complaints. For example the previous `codat.io/sync-for-commerce` golden was:

```
failed to unmarshal data: json error: invalid character 'o' looking for beginning of value,
yaml error: ... json: cannot unmarshal bool into field Schema.properties of type openapi3.Schema
```

## Tests

`openapi3/schema_boolean_test.go` covers a boolean in each of `items`, `prefixItems`, `not` and a `properties` value; instance validation of both booleans against every JSON type including null; the marshal round-trip; `IsEmpty`; and that an ordinary object schema is unchanged.

It also covers the compositions, which are where a boolean stops being trivial:

- `not: true` matches nothing and `not: false` matches everything, asserted through the composition rather than only on the decoded value.
- `Validate` refusing a `Boolean` next to a type, a property or an extension.
- Both validation paths. `VisitJSON` documents `EnableJSONSchema2020` as the way to validate a 3.1 schema, and a boolean schema exists only in 3.1, so the option's path is asserted to agree with the built-in one. It is not purely external either: `useJSONSchema2020` falls back to `visitJSON` when the schema does not compile, which a bare boolean at the root does.
- `items: false` closing a tuple at the `prefixItems` length, on both paths.

## Closed tuples

`items: false` is how a tuple is closed to a fixed length, so it is the idiom this change exists to make expressible. It works end to end now that #1259 has landed on master and `visitJSONArray` applies `items` only to the positions past `prefixItems`: under `prefixItems: [string, integer]` with `items: false`, a prefix such as `["a"]` is valid and a third element is refused. The built-in and 2020-12 validators agree on it, and both are asserted.

## Checks

`go generate`, `maps.sh`, `docs.sh`, `go mod tidy`, `go vet`, `go fmt`, `nilness` and the style greps all pass. `modernize` reports two findings, both present on master and in files this does not touch. Full suite green, including `-count=10 -short`.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -151,6 +151,17 @@ type Schema struct {
 
 	DependentRequired map[string][]string `json:"dependentRequired,omitempty" yaml:"dependentRequired,omitempty"` // OpenAPI >=3.1
 
+	// Always is set when the schema was written as a JSON Schema boolean
+	// schema rather than an object, and says what that schema always does:
+	// `true` accepts every instance, `false` accepts none. It is nil for an
+	// ordinary object schema, and Validate rejects one that carries any other
+	// keyword.
+	//
+	// Neither direction goes through a struct tag, hence the "-" on both:
+	// UnmarshalJSON takes the bare boolean before the field decode, and
+	// MarshalYAML emits it again ahead of the field map. OpenAPI >=3.1.
+	Always *bool `json:"-" yaml:"-"`
+
 	Defs          Schemas `json:"$defs,omitempty" yaml:"$defs,omitempty"`       // OpenAPI >=3.1
 	SchemaDialect string  `json:"$schema,omitempty" yaml:"$schema,omitempty"`   // OpenAPI >=3.1
 	Comment       string  `json:"$comment,omitempty" yaml:"$comment,omitempty"` // OpenAPI >=3.1
@@ -480,6 +491,10 @@ func (schema Schema) MarshalJSON() ([]byte, error) {
 
 // MarshalYAML returns the YAML encoding of Schema.
 func (schema Schema) MarshalYAML() (any, error) {
+	if x := schema.Always; x != nil {
+		return *x, nil
+	}
+
 	m := make(map[string]any, 61+len(schema.Extensions))
 	maps.Copy(m, schema.Extensions)
 
@@ -684,8 +699,28 @@ func (schema Schema) MarshalYAML() (any, error) {
 	return m, nil
 }
 
+// unmarshalBooleanSchema reports whether data is a JSON Schema boolean schema,
+// and its value if so. Only a bare `true` or `false` qualifies; an object, a
+// string or anything else is an ordinary schema for the caller to decode.
+func unmarshalBooleanSchema(data []byte) (bool, bool) {
+	var boolean bool
+	if err := json.Unmarshal(data, &boolean); err != nil {
+		return false, false
+	}
+	return boolean, true
+}
+
 // UnmarshalJSON sets Schema to a copy of data.
 func (schema *Schema) UnmarshalJSON(data []byte) error {
+	// JSON Schema 2020-12 allows a boolean wherever a schema is expected, so
+	// every SchemaRef position can hold one. SchemaRef.UnmarshalJSON funnels
+	// into here, which is why this one case covers them all.
+	if boolean, ok := unmarshalBooleanSchema(data); ok {
+		origin := schema.Origin
+		*schema = Schema{Always: &boolean, Origin: origin}
+		return nil
+	}
+
 	type SchemaBis Schema
 	var x SchemaBis
 	if err := json.Unmarshal(data, &x); err != nil {
@@ -1379,6 +1414,14 @@ func (schema *Schema) IsEmpty() bool {
 	if cs := schema.ContentSchema; cs != nil && cs.Value != nil && !cs.Value.IsEmpty() {
 		return false
 	}
+	// Last, so a schema carrying both a boolean and other keywords is judged by
+	// the keywords. Neither boolean is empty: `false` rejects every instance,
+	// and `true` is accepted by visitJSON explicitly rather than by skipping
+	// validation, which is what leaves `not: true` free to reject.
+	if schema.Always != nil {
+		return false
+	}
+
 	return true
 }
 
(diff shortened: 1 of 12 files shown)


openapi3/schema.go
```diff
@@ -151,6 +151,15 @@ type Schema struct {
...
 	DependentRequired map[string][]string `json:"dependentRequired,omitempty" yaml:"dependentRequired,omitempty"` // OpenAPI >=3.1
 
+	// Boolean is set when the schema was written as a JSON Schema boolean
+	// schema rather than an object: `true` accepts every instance, `false`
+	// accepts none. It is nil for an ordinary object schema, and no other
+	// field is populated when it is set.
+	//
+	// Marshalling is handled by MarshalYAML, which emits the bare boolean, so
+	// this carries no struct tag of its own. OpenAPI >=3.1.
+	Boolean *bool `json:"-" yaml:"-"`
```
fenollp: How about naming the field `Empty`?

Also schema Validate should check this field and IsEmpty agree. Eg. This field may be set after deserialization and a new serialization may surprise users. 

This implies that IsEmpty should check this field at the very last step.

fenollp (changes requested):
Please share your thoughts on my comments

openapi3/schema.go
reuvenharrison: Both of the Validate/IsEmpty points were right, and chasing the second one turned up a bug. Pushed in `cae310b`.

**Validate and IsEmpty agreeing.** Validate now rejects a `Boolean` set alongside any other keyword, including extensions, rather than let the next serialization drop them:

```go
rest := *schema
rest.Boolean = nil
if !rest.IsEmpty() || len(schema.Extensions) != 0 {
    return stack, newSchemaBooleanFieldsExclusive(schema.Origin)
}
```

New code `boolean-schema-with-other-keywords` under the `MutuallyExclusiveFieldsError` cluster. `IsEmpty` consults the field last, as you suggested, so a schema carrying both is judged by its keywords.

**The bug that fell out of it.** `IsEmpty` returned `*Boolean`, so a `true` schema reported itself empty, and `visitJSON` skips validation for an empty schema. That made `not: true` accept every value when it should accept none:

```
not: true   VisitJSON(1) = <nil>     // wrong, matches nothing
```

`IsEmpty` now reports a boolean schema as not empty in either direction, and `visitJSON` accepts `true` explicitly rather than by skipping. `not: true` rejects and `not: false` accepts, both asserted now.

The same shape exists on master independently of this PR: `{not: {}}` also reports empty and accepts everything, since the empty schema matches everything and so `not: {}` matches nothing. Filed and fixed separately in #1260, which reproduces on master and does not depend on this one.

**On the name.** I would rather not use `Empty`, because it inverts for one of the two values: `false` is the schema that accepts nothing, so `Empty: &false` would read as "not empty" while meaning the most restrictive schema there is. The connection to `IsEmpty` is also weaker now that neither boolean counts as empty. `Boolean` follows the specification, which calls `true` and `false` boolean schemas.

If you would still rather it were something else, `Bool` or `Always` both work for me and I will rename without argument. I only want to avoid a name that says the opposite of what it means for `false`.

fenollp: I like Always :-)

reuvenharrison: Done in `72cac71`.

reuvenharrison:
Your point on #1260 about gating behind a version check applies here, so I have added a gate in `33dedd9`. (https://github.com/getkin/kin-openapi/pull/1260#issuecomment-5436938668)

A boolean schema is JSON Schema 2020-12, so it does not exist in 3.0, where a schema MUST be a Schema Object. Before this the PR accepted `items: false` in a 3.0 document without comment:

```
3.0.0  load  -> ok, items.Always=false
3.0.0  Validate -> <nil>
```

Validate now reports `boolean-schema-for-3-1-plus` through the same `FieldVersionMismatchError` path as `const`, `prefixItems` and the rest of the 2020-12 family, so a 3.0 document gets the same treatment for this as for every other 3.1 construct. Parsing is unchanged, since the version is a validation matter.

fenollp (approved):

PR #1260, opened 2026-08-26
openapi3: reject every value for `not: {}`

reuvenharrison:
The empty schema matches every instance, so `not: {}` matches none. Today it accepts every value instead.

```yaml
components:
  schemas:
    Nothing:
      not: {}
```

```go
nothing.IsEmpty()          // true
nothing.VisitJSON("x")     // nil, accepted
nothing.VisitJSON(1)       // nil, accepted
```

## Cause

`IsEmpty` judged the `not` by whether its own value was empty:

```go
if n := schema.Not; n != nil && n.Value != nil && !n.Value.IsEmpty() {
    return false
}
```

An empty `not` therefore did not make the schema non-empty, and `visitJSON` short-circuits on an empty schema before `visitNotOperation` runs. So the one keyword present was never applied, and the strictest schema expressible became the most permissive one.

A `not` is a constraint whatever it holds, so its presence alone makes the schema non-empty:

```go
if schema.Not != nil {
    return false
}
```

A non-empty `not` already made the schema non-empty and is unaffected.

## Tests

`openapi3/schema_not_empty_test.go` covers `not: {}` at the root and through a property, across every JSON type, and pins that a non-empty `not` still behaves as before.

Full suite green, no golden files move, and `docs.sh` reports no public API change.

## Note

Found while reviewing #1258, which lets a schema be written as a boolean. `not: true` is the same defect through the new syntax and is fixed there, since that PR introduces the syntax. This one is independent and reproduces on master today.

### DIFF
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1281,7 +1281,12 @@ func (schema *Schema) IsEmpty() bool {
 		schema.Const != nil {
 		return false
 	}
-	if n := schema.Not; n != nil && n.Value != nil && !n.Value.IsEmpty() {
+	// A `not` is a constraint whatever it holds, and an empty one is the
+	// strongest of all: `not: {}` matches nothing, since the empty schema
+	// matches everything. Judging it by whether its own value is empty had the
+	// effect of dropping it, and visitJSON skips an empty schema entirely, so
+	// `not: {}` accepted every value instead of rejecting every value.
+	if schema.Not != nil {
 		return false
 	}
 	if ap := schema.AdditionalProperties.Schema; ap != nil && ap.Value != nil && !ap.Value.IsEmpty() {
--- a/openapi3/schema.go
+++ b/openapi3/schema.go
@@ -1281,11 +1281,9 @@ func (schema *Schema) IsEmpty() bool {
 		schema.Const != nil {
 		return false
 	}
-	// A `not` is a constraint whatever it holds, and an empty one is the
-	// strongest of all: `not: {}` matches nothing, since the empty schema
-	// matches everything. Judging it by whether its own value is empty had the
-	// effect of dropping it, and visitJSON skips an empty schema entirely, so
-	// `not: {}` accepted every value instead of rejecting every value.
+	// A `not` constrains the schema whatever it holds, and an empty one is the
+	// strongest constraint of all: `not: {}` matches nothing, since the empty
+	// schema matches everything.
 	if schema.Not != nil {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


fenollp:
So this is correct for json schema, however before oasv3.1 schemas weren't exactly this as you know. Then, an empty schema still did not allow everything and this is why I introduced IsEmpty(). 

So this should be gated behind a version check.

reuvenharrison:
After analyzing this more closely, master is buggy on 3.0 too. Per the 3.0 spec `not: {}` should accept nothing at all; on master it accepts everything except null, because `IsEmpty` short circuits and the `not` is never applied, so the schema behaves exactly as if it had been written `{}`.

```
MASTER 3.0  not:{} vs null    -> rejected
MASTER 3.0  not:{} vs string  -> ACCEPTED
MASTER 3.0  not:{} vs number  -> ACCEPTED
MASTER 3.0  not:{} vs bool    -> ACCEPTED
MASTER 3.0  not:{} vs array   -> ACCEPTED
MASTER 3.0  not:{} vs object  -> ACCEPTED
```

The derivation is the same in both versions. `{}` is a schema with no constraints, so it matches every instance. `not: S` matches exactly what `S` does not. `{}` rejects nothing, so `not: {}` matches nothing.

In 3.0 that empty schema is an OAS Schema Object and in 3.1 it is a JSON Schema 2020-12 schema, but the dialect only limits which keywords may appear inside it. 3.0's adjustment for `not` says exactly that, "Inline or referenced schema MUST be of a Schema Object and not a standard JSON Schema", and it uses the same sentence for `allOf`, `oneOf`, `anyOf` and `properties`. An empty schema is unconstrained either way.

The one thing that could make 3.0 differ here is `nullable`, and it does not. It "adds `null` to the allowed type specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object". `{}` has no `type`, so nothing is added and no type constraint applies. `{}` still matches everything, null included.

So both versions want the same result and there is nothing for a gate to select between. The one line change gives it in both:

```
3.0.0  not:{} rejects: null string number bool array object
3.1.0  not:{} rejects: null string number bool array object
```

It also does not depend on how `{}` treats null today. Currently null is rejected by the nullability rule: the inner `{}` rejects null, so the `not` is satisfied and the outer schema rejects null on its own account. If `{}` is ever made to accept null, the inner accepts, the `not` fails, and the outer rejects null that way instead. Same result by a different route.

Separately, `{}` rejecting a bare null is a real difference from the spec, but it applies to 3.1 as well, so it is not what separates the versions. That one is worth its own issue.

Added null coverage to the tests in `d5a2602`, plus the 3.1 case, since those were what I had not asserted.

Sorry for the convoluted argument. I simplified it as far as I could and it may still take a second read to hold it all in your head, but I am confident it is correct.

